PR #18, opened 2017-05-18
Use Webpack to build scss (+ better management of node_modules in Docker)

### LINKED ISSUE #13: Use Webpack/Sass to compile CSS

Atala:
Fix #13

Can you check if it works with Vagrant? I have added some conf for Docker [here](https://github.com/coopcycle/coopcycle-web/blob/alois/webpack.config.js#L53-L57) that may break assets serving from localhost.

If it doesn't work, I think I will add an env var in docker-compose so we know when we are using Docker or Vagrant.


alexsegura:
Nice! ❤️

I'll check this tonight.

alexsegura:
I rebased a little bit. 

Also, I ended moving the Webpack dev server URL to `app/config/parameters.yml`, it is actually simpler.

PR #28, opened 2017-06-09
Docker tests

Atala:
Setup to tests with docker 

I don't know why the diff is so big, maybe because I merged master back into my branch :/ ?

Anyway it almost work by me, only issue is with Behat tests, I have this log
```php
      Type error: Too few arguments to function AppBundle\Action\Order\Pick::__construct(), 0 passed in vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php on line 202 and exactly 7 expected (Behat\Testwork\Call\Exception\FatalThrowableError)
```
I didn't find a fix yet, I guess it's symphony-related, any idea @alexsegura ? 

**Edit**: maybe the service registratrion from dunglas isn't working in docker, I have:
```sh
php bin/console debug:container AppBundle
11:27:53 ERROR     [console] Error thrown while running command "debug:container AppBundle". Message: "No services found that match "AppBundle".
```


Atala:
Also, I have added phpunit as a composer depedency, I guess it's better (?) - and so I have it properly installed in my container out-of-the box thanks to `composer install`

alexsegura:
For PHPUnit actually I also did it a[ 2 commits before](https://github.com/coopcycle/coopcycle-web/commit/5785753e4f4624817c2dfb7a237ed9e4def89871), sorry 😐

Since we are now building against versions of PHP which are not installed by default on Travis, the `phpunit` didn't work correctly. 
Actually the `phpunit` shell command looks for PHPUnit in the `vendor` folder in the current directory, then it fallbacks to the parent folder, etc...
I upgraded to PHPUnit 6.

I removed the wrong commits from the branch, the Travis build works. 
The `composer.lock` file was wrong. 

Now lets test & review 🙂

docker/postgres/create-test-db.sh
```diff
@@ -0,0 +1,21 @@
...
+CREATE DATABASE coopcycle_test;
+
+EOSQL
+
+# Load PostGIS into coopcycle_test
+echo "Loading PostGIS extensions into coopcycle_test"
+"${psql[@]}" --dbname="coopcycle_test" <<-'EOSQL'
+    CREATE EXTENSION IF NOT EXISTS postgis;
+    CREATE EXTENSION IF NOT EXISTS postgis_topology;
+    CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;
```
alexsegura: What are those extensions used for?

alexsegura:
Ok, I know why there was this error specifically on Docker:

```
Type error: Too few arguments to function AppBundle\Action\Me::__construct(), 
0 passed in vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php 
on line 202 and exactly 7 expected (Behat\Testwork\Call\Exception\FatalThrowableError)
```

It seems the `DunglasActionBundle` had a specific bug under Alpine Linux, which explains why the bug only occurred on Docker. 
 
capture d ecran 2017-06-10 a 11 59 18

[This is fixed ](4f705fadf4a2adb57a57744d89838650aee00617)

Atala:
Soooo nice. I should have noticed that `"dunglas/action-bundle": "^0.3"` wouldn't install the 0.4.1 version though.

Can you revert a3befa7 ?

alexsegura:
Still facing this:

```
Warning: file_get_contents(): php_network_getaddresses: getaddrinfo failed: 
Name does not resolve
```

This is because the `php` container tries to make an HTTP request to `http://osrm:5000`, but the `osrm` host is not resolved. 

I tried adding an entry in `depends_on` & `links` for the `php` container to be able to see the `osrm` container, but it doesn't work. 
It works for other containers, I don't understand...

alexsegura:
Ok, I should have read the logs: 

```
osrm_1      | [warn] Missing/Broken File: /data/data.osrm.datasource_indexes
osrm_1      | [warn] "/data/data.osrm.datasource_indexes" is not found
coopcycleweb_osrm_1 exited with code 1
```

The `osrm` container was not starting, so this is why it wasn't available. 
I rebuild the OSRM files, and now it works 😅

README.md
```diff
@@ -127,7 +143,7 @@ vagrant ssh -c 'sudo -u www-data php /var/www/coopcycle/bin/console doctrine:sch
 
 With Docker
 ```
-php bin/console doctrine:schema:create --env=test
+docker-compose run php bin/console doctrine:schema:create --env=test
```
Atala: 👍

PR #42, opened 2017-08-23
Add fields to the delivery address form

Atala:
Fields added : 
 - Name
 - First Name
 - Additional Information
 - Phone Number


Atala:
Need to fix the build, wait for it.

tests/ContainerAwareUnitTestCase.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Tests;
+
+use PHPUnit\Framework\TestCase;
+require_once __DIR__.'/../app/AppKernel.php';
+
+
+class ContainerAwareUnitTestCase extends TestCase {
```
alexsegura: Looks like this is now integrated in Symfony, use [KernelTestCase](http://symfony.com/doc/current/testing/doctrine.html) instead

src/AppBundle/Entity/Base/PostalAddress.php
```diff
@@ -78,11 +91,31 @@
...
+
+    /**
+     * @var string Phone Number, mandatory for delivery.
+     *
+     * @ORM\Column(nullable=true)
+     * @Assert\Type(type="string")
+     * @Assert\NotBlank(groups={"delivery_address"})
+     * @ApiProperty(iri="https://schema.org/streetAddress")
+     */
+    protected $phoneNumber;
```
alexsegura: This should be in the `ApiUser` class IMHO. 
One customer may have several addresses (home, work, a friend's house), but the phone number is the same.

src/AppBundle/Entity/Base/PostalAddress.php
```diff
@@ -50,15 +51,27 @@
...
 
     /**
+     * @var string Family name. In the U.S., the last name of an Person. This can be used along with givenName instead of the name property.
+     *
+     * @ORM\Column(nullable=true)
+     * @Assert\Type(type="string")
+     * @Assert\NotBlank(groups={"delivery_address"})
+     * @ApiProperty(iri="https://schema.org/familyName")
+     */
+    private $familyName;
```
alexsegura: Is the `familyName` really helpful? This is an address.
Some people may get delivered at "Acme Company" address.

Are we ok that `name` should the name of the item, i.e. "Home", "Work"?
I think everything can be managed through the `description` field for additional details

src/AppBundle/Entity/Base/PostalAddress.php
Atala: OK

src/AppBundle/Entity/Base/PostalAddress.php
Atala: Name is important IMHO. For ex. : you deliver to one person at Acme Company, you enter the building and need to find the person. (+ I think people are expecting us to take their name).

Arf I didn't get that `name` was for the address name.

Atala:
Proposal : 
 - add field 'firstName', 'lastName' and 'phoneNumber' to the ApiUser model
 - create a model 'DeliveryAddress' which is linked to an order and frozen at each order (extends 'Address' with firstName, lastName,  phoneNumber)
 - add field 'description' to Address

UI : 
 - address entering in two steps when unknown address : first address, then additional details
 - if an user as already an address, all fields are prefilled and editable

Atala:
Closing this one and re-starting on a fresh branch.

PR #52, opened 2017-08-31
Freeze address when creating a delivery

Atala:


tests/AppBundle/Utils/CartTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace AppBundle\Utils;
+namespace Tests\AppBundle\Utils;
```
alexsegura: I think this is useless.

tests/Fixtures/RestaurantFixture.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public function getDependencies()
+    {
+        return array(AddressFixture::class);
+    }
+}
+
+
+?>
```
alexsegura: Useless.

Atala:
I add difficulties to find the places where a change was necessary because I didn't know the codebase well-enough so the PR is convoluted. It breaks also the JS widget at the order/address creation (**-> need to fix it**).

However I still think it is the good way to go : 
 - an user has a set of `Addresses` that he can edit at will
 - a `Delivery` is linked to a `DeliveryAddress` which is frozen once and for all.

@alexsegura any thoughts? Also the behat tests are fine on my machine but NOK on travis, can you give it a try?

tests/AppBundle/Utils/CartTest.php
Atala: Most of the tutorials I saw where using a different namespace for tests. Not a PHP expert here, but isn't it possible that it will cause issues if some classes from the app and from the tests collide?

tests/Fixtures/RestaurantFixture.php
Atala: Yeah, was planning it to use it for some functional tests. I can remove for now on.

tests/AppBundle/Utils/CartTest.php
alexsegura: Nope, because test classes are suffixed by `Test`

src/AppBundle/Entity/DeliveryAddress.php
```diff
@@ -3,6 +3,7 @@
 namespace AppBundle\Entity;
 
 
+use AppBundle\Action\Delivery\Deliver;
```
alexsegura: Bad import, you can remove it.

tests/Fixtures/RestaurantFixture.php
alexsegura: I was talking about the closing PHP tag 🙂

Atala:
@alexsegura I propose we close this one if we don't integrate it now.

If we want to do this dev later on we/I will start from scratch and to it clean. As a matter of fact, I think the freeze problem is not just for delivery addresses but also for MenuItem, MenuItemModifier, aso. Let's discuss it!

PR #83, opened 2017-10-05
Items deletion/unavailability

Atala:


alexsegura:
There are Doctrine extensions to manage soft delete out of the box https://github.com/Atlantic18/DoctrineExtensions 😉

Let me change it.

alexsegura:
@Atala I rebased the whole stuff into one commit. 

I rolled back the code you deleted, still ugly, but I think it works now 🙂
When a section is deleted, the `menu_id` is set to `null` and all items are soft deleted. 

Can you recheck?

PR #86, opened 2017-10-06
Create Delivery when Order is accepted

alexsegura:
This introduces the real order workflow : 

Customer orders ➡ Restaurant accepts ➡ Delivery is created

I added an `OrderManager` class to avoid repeating the code, in the API and in the controllers. 
Also, this introduces two-step payment with Stripe: funds are actually captured when the restaurant accepts the order.


Atala:
Also I don't think the correct way is to use transferGroups. I think we should credit the charge to the restaurant and use [application fee](https://stripe.com/docs/connect/direct-charges#collecting-fees) instead. But this trigger the following questions : who will be responsible to do refunds in case something goes wrong? Us or the restaurant?

alexsegura:
Ok, I remove dead code 👍

Regarding Stripe Connect params, we need to keep this for now, as we would need to initialize a demo with all restaurants having their own account right now.

alexsegura:
Regarding refunds, this is entirely managed by Stripe.

Issue #99, opened 2017-10-20
Translate our React Components


Atala:
We are using more and more React components on the front, but they are all monolanguage :'(

We should find a way to load some translations files in it.

Atala:
Done #381

PR #101, opened 2017-10-24
Early address selection

Atala:
- add address + date picker on restaurant list page
- use a react component on homepage with addresses suggestions


js/app/restaurant-list/RestaurantListFilter.jsx
```diff
@@ -0,0 +1,136 @@
...
+    const inputProps = {
+      value: address,
+      onChange: (value) => { this.onAddressChange(value) },
+      onBlur: () => { this.onAddressBlur() },
+      placeholder: 'Entrez votre adresse'
+    }
+
+    if (!wantsNow) {
+      selectElement = <DatePicker
+                        setDeliveryDate={ this.props.onDatePickerChange }
```
alexsegura: I have renamed it to `onChange` (naming things is hard haha)

js/app/restaurant-list/RestaurantListFilter.jsx
```diff
@@ -0,0 +1,136 @@
...
+                        availabilities={ this.props.availabilities }
+                        deliveryDate={ this.props.initialDate }
```
alexsegura: I have renamed it to `value` (same comment as above) 😄

alexsegura:
Looks good to me. 

Just fix 2 lines and I think we can :shipit:

assets/css/vars.scss
```diff
@@ -0,0 +1,15 @@
+// Colors
```
alexsegura: Nice to have introduced a file for variables 👍

js/app/cart/DatePicker.jsx
```diff
@@ -102,7 +103,7 @@ class DatePicker extends Component {
             {
               dates.map(date => (
                 <option key={ date } value={ date } >
-                  { moment(date).format('dddd DD MMM') }
+                  { moment(date).calendar(null, { sameDay: '[Aujourd\'hui]',  nextDay: '[Demain]', sameElse: 'dddd DD MMM'}) }
```
alexsegura: Can we use `window.__i18n`?

Atala:
Note: at the end, the ideal would be to have the same `DatePicker` in coopcycle-web and in coopcycle-js.

alexsegura:
@Atala yep, but on `coopcycle-js` there is Redux, and I had to adapt the code

js/app/cart/DatePicker.jsx
Atala: I think we should use a pure-JS solution (like [this](https://github.com/martinandert/counterpart) (didnt look into it) to translates our React component, `window.__i18n` is nice but don't scale.
I opened #99 to discuss this.

Atala:
Merged !

PR #110, opened 2017-10-29
fix: use MAX_DISTANCE to filter restaurants list

Atala:


alexsegura:
I'm not comfortable with using static values, this is a pain for unit testing. 
We may need to configure max distance per restaurant, I don't know. 

Maybe just introduce `MAX_DISTANCE_DEFAULT`, which is returned by `getMaxDistance` for the moment?

Atala:
Ok, that's the same for me.

Main issue here was that the filter on restaurant list was using `5000` as default value. (Note also that once it is configured per restaurant, we we'll need to rewrite these search functions).

Atala:
Semi-fixed by f2c8f88 , closing for now on.

PR #113, opened 2017-11-05
Add more info to install

florentk:


Atala:
Hey Florent,

You shouldn't need sudo to run the `make install` command. Which command causes problem?
 Which platform are you on?

Aloïs

Atala:
Solved in b712b5a

PR #116, opened 2017-11-05
add sudo to `docker-compose up` command

florentk:


Atala:
Hey Florent,

You shouldn't need `sudo` to run the `docker-compose` command. Which platform are you on?

Aloïs

florentk:
I have this error without `sudo` when I lauch  `docker-compose up` and `make install` :
`ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?`

florentk:
En faite, d'après la doc de Docker :+1: 

> If you would like to use Docker as a non-root user, you should now consider
> adding your user to the "docker" group with something like:
`sudo usermod -aG docker your-user`

Atala:
Added instructions in b712b5a

PR #121, opened 2017-11-06
Fix Issue #118 change field property

florentk:
change field property of AddressForm builder : "required => true" insteadof  "disabled => true"


Atala:
Hey Florent,

As stated in [this comment](https://github.com/coopcycle/coopcycle-web/issues/118#issuecomment-342123097) it is not the correct fix here.

Problem is the form is used two times with two different purposes : 
 - on order page : we don't want the user to be able to change his address (for now on at least, as it would need more validation)
 - on profile page, where we wan the user to be able to change (obviously)

Atala:
Also note that `required => true` is the default for Symfony so no need to specify it.

Issue #120, opened 2017-11-06
Default address when order


florentk:
I add one user address in my account. But  when I order, the address form is initialized with a other address. 

The bug is in RestaurantController.php:176, I not understand this code 

`
            // we need to compute the geohash, because of small differences between geodata from the back and from the front
            $address = $user->getAddresses()->filter(
                function ($address) use ($addressData, $geotools) {
                    $addressCoordinate = new Coordinate([$address->getGeo()->getLatitude(), $address->getGeo()->getLongitude()]);
                    $basketCoordinate = new Coordinate([$addressData['latitude'], $addressData['longitude']]);
                    return $geotools->geohash()->encode($addressCoordinate, 9)->getGeohash() ===
                        $geotools->geohash()->encode($basketCoordinate, 9)->getGeohash();
                }
            );
`

Why a filter is needed ? When I delete the filter, the address form is initialized with my first address.

Atala:
`But when I order, the address form is initialized with a other address.` 

On which page?

Atala:
A non-problem AFAIK, please reopen or comment if needed

PR #122, opened 2017-11-06
Issue #120 get first customer address when order

florentk:
Disable filter to get the first customer address


Atala:
I think you are misunderstanding the purpose of this code.

It is meant to be run **after** the user has chosen his address. The purpose of it is to know if we already have this address saved (so in this case no need to add something new to it's address book) or if it is a new one.

Atala:
Closing because not relevant, @florentk please ping me on Slack if you want to talk about this issue.

PR #123, opened 2017-11-07
Change design of restaurant interface.

alexsegura:
This changes the design of the restaurant dashboard for orders. 

Orders are pushed via WebSocket as before.
Browser history is updated using `window.pushState`

restaurant_ui


js/app/restaurant/OrderList.jsx
```diff
@@ -1,55 +1,99 @@
...
+    let className = ''
+    if (this.state.active && this.state.active['@id'] === order['@id']) {
+      className = 'active'
+    }
+
+    return (
+      <tr key={ order['@id'] } onClick={ () => this.props.onOrderClick(order) } style={{ cursor: 'pointer' }} className={ className }>
+        <td>#{ order['@id'].replace('/api/orders/', '') }</td>
+        <td><OrderLabel order={ order } /></td>
+        <td><i className="fa fa-clock-o" aria-hidden="true"></i>  { moment(order.delivery.date).format('lll') }</td>
```
Atala: IMHO we shouldn't show the delivery date but the date the dish has to be prepared (i.e. when the courrier will come).

For now on we have a fixed preparation time of 30 min and delivery time of 15 min [here](https://github.com/coopcycle/coopcycle-web/wiki/Business-assumptions) so you should show `deliveryDate - 30 min.`

src/AppBundle/Entity/OrderRepository.php
```diff
@@ -8,26 +8,27 @@
...
+
         $qb = $this->createQueryBuilder('o');
         $qb
             ->join(Delivery::class, 'd', Expr\Join::WITH, 'd.order = o.id')
             ->andWhere('o.restaurant = :restaurant')
             ->andWhere($qb->expr()->in('o.status', [
                 Order::STATUS_WAITING,
-                Order::STATUS_ACCEPTED
+                Order::STATUS_ACCEPTED,
+                Order::STATUS_REFUSED
```
Atala: I think we should show `STATUS_READY` too, because the restaurant owner still need it when the courier arrives to print the receipt.

I think we will show the ready orders that don't have been picked (`STATUS_PICKED` on delivery). Maybe we could only filter on delivery status? All items that are `STATUS_WAITING` or `STATUS_DISPATCHED` are in charge of the restaurant.

src/AppBundle/Entity/OrderRepository.php
```diff
@@ -8,26 +8,27 @@
...
             ]))
+            ->andWhere('d.date >= :date')
```
Atala: We should show orders that are "in the past". These are orders have an issue that should be addressed by someone, so they shouldn't disappear in our system. (can happen that the owner is late, courier late, aso.)

Maybe add a label for orders which are late - taking care of the [business assumptions](https://github.com/coopcycle/coopcycle-web/wiki/Business-assumptions)

alexsegura:
I made the changes you requested @Atala

Atala:
In 2abf9e7 , you have to split PREPARATION_AND_DELIVERY_DELAY in PREPARATION_DELAY (30 minutes) and DELIVERY_DELAY (15 minutes).

alexsegura:
👏

PR #132, opened 2017-11-09
No profile side menu

Atala:
Main change in this PR is that it is deleting the side menu in the profile section. All pages are accessible thanks to the header dropdown.

Reason for deleting the side-menu :
 - pages already accessible from top 
 - it is annoying on mobile
 **- more room to do nice things (for example the "tracking order" screen is nicer with more width, the "my orders" list too)**

@alexsegura it is a big change in design so I want to make sure it is ok with you


alexsegura:
Of course, I was too lazy to change it 😐

Issue #63, opened 2017-09-19
Use one database in Redis for all production data


alexsegura:
The best practice seems to be using always database 0 to store all production data, prefixing keys with namespaces.

Atala:
It's gonna be easy, actually `predis` comes with a `prefix` option

:)

alexsegura:
Fixed in #137

PR #137, opened 2017-11-10
fix: add prefixes to Redis

Atala:
F*x #63 

I think we just need to redeploy each instance and it will be fine, as data are always polled from Postgres.

Note that I did a small trick to prefix `subscribe` calls.

<strike>Please don't merge I have still a [question about this change](https://github.com/NodeRedis/node_redis/issues/1286)</strike>

### DIFF
--- a/js/app/cart/Cart.jsx
+++ b/js/app/cart/Cart.jsx
@@ -101,7 +101,6 @@ class Cart extends React.Component
             }
 
         $.post(this.props.addToCartURL, {
-          date: this.props.deliveryDate,
           address: address
         }).fail((e) => { this.handleAjaxErrors(e.responseText) })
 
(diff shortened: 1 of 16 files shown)


Atala:
I don't know what happened with the `fsevents` dependency, is it used for something?

alexsegura:
I'll check this week.

js/app/cart/Cart.jsx
```diff
@@ -101,7 +101,6 @@ class Cart extends React.Component
             }
 
         $.post(this.props.addToCartURL, {
-          date: this.props.deliveryDate,
```
alexsegura: Is this normal?

Atala: Yep (I dont remember why this change is here though)

alexsegura: I extracted it into a single commit 😉

PR #139, opened 2017-11-14
Improve validation on cart/orders

Atala:
Is it the correct way to do it in Symfony @alexsegura ?


src/AppBundle/Validator/Constraints/IsValidDeliveryDateValidator.php
```diff
@@ -0,0 +1,25 @@
...
+use Carbon\Carbon;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+class IsValidDeliveryDateValidator extends ConstraintValidator
+{
+
+    public function validate($object, Constraint $constraint)
+    {
+        $now = Carbon::now();
```
Atala: @alexsegura for the life of me I couldn't find a way to inject a date here for unit testing, and it's so easy with Carbon... :/

I don' t have  any other solution for now :/

alexsegura:
I'll check this week.

src/AppBundle/Validator/Constraints/IsValidDeliveryDateValidator.php
alexsegura: Looks good to me.

Issue #95, opened 2017-10-19
Ask for confirmation when beginning shopping in a new restaurant


Atala:
Use case is the following : 
 - shop in restaurant 1 (add some products)
 - go back to restaurant page
 - go to restaurant 2 page
 - click item

We should : 
 - ask the customer if he wants to shop in restaurant 2, telling him he started in restaurant 1
 - if yes, clear the basket and add the item to the basket
 - if no, do nothing

Atala:
The code which is triggered when a user add an item to the cart is [here](https://github.com/coopcycle/coopcycle-web/blob/master/js/app/cart/index.jsx#L32). So at the beginning of the handler we need to check if the id of the current restaurant is the same as the id of the restaurant of the cart. To access this data from javascript we need to pass the variables thanks to the PHP templating [here](https://github.com/coopcycle/coopcycle-web/blob/master/src/AppBundle/Resources/views/Restaurant/index.html.twig#L93-L105). One is `restaurant.id` (current restaurant), other one is `cart.restaurant.id` .
If the ids don't match, open a [bootstrap modal](https://getbootstrap.com/docs/3.3/javascript/#modals) with a warning.
The modal has two buttons : 
 - one that add the items as attended (means you need to save the menu item id that was clicked / handle the case when it should open a modifiers modal) (no need to clear the cart, it will clear itself)
 - one that does nothing (easy :) )

Ladias-Elie:
I am struggling a bit with this issue, here is how I would proceed, do you think it's a good way of doing ? 

1. Create a cartTotal components:  it will be stored in cartTotal.jsx and will render the session.cartTotal for non restaurant pages and the current cartTotal for pages where we can add items to the basket

2. Call the cartTotal components: Here I am not sure how, should we create a new js script that would be call in the menu.html.twig file ? If not should we use the index.js file ? But if so how do we handle the case when we are prompting something not in a restaurant page ?

3. When a new item is added: For the first one, prompt a bootstrap message box:  
	a) if the customer want to buy in this new restaurant: 	erase the old basket, add the new item and update the cartTotal componment
	b) if not do nothing

Atala:
Let's start with the (1). It is actually quite complicated.

What you want to do is to have two React components which are storing the same state (the same `cartItems`, the same `total`). So you would like that the `Cart` component render at two places when an element is added at the same time : in the right-column and in the header.

[...]

PR #143, opened 2017-11-21
Ask for confirmation when beginning shopping in a new restaurant #95

Ladias-Elie:
This PR address two issues:

- Ask for confirmation when a customer buy products in a new restaurant while having already products from another restaurant in his cart.

- Update the header cart on the restaurant page when a customer add a new product.

I still have to fix a bug with products that require to choose different existing option, but I will be happy to have a small review to be sure I am going in the right direction.


js/app/cart/Cart.jsx
```diff
@@ -186,11 +202,13 @@ class Cart extends React.Component
             </div>
           </div>
         </div>
+        {this.state.sameTopCart ? (<CartPortal><CartTop total={sum} i18n={this.props.i18n}/></CartPortal>):(<div></div>)}
```
Atala: Change for `{this.state.sameTopCart && (<CartPortal><CartTop total={sum} i18n={this.props.i18n}/></CartPortal>)}` which is more elegant ([ref](https://stackoverflow.com/questions/28258465/how-to-add-logical-if-statement-when-rendering-react-components?answertab=votes#tab-top))

js/app/cart/CartPortal.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import ReactDOM from 'react-dom';
+
+class CartPortal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.el = document.querySelector('[id="bs-example-navbar-collapse-1"]');
```
Atala: You can change the id attr of this element so it will be more explicit - change `bs-example-navbar-collapse-1` to `cart-top`

js/app/cart/index.jsx
```diff
@@ -9,6 +9,8 @@ let isXsDevice = $('.visible-xs').is(':visible')
 var cart = document.getElementById('cart');
 var cartComponent;
 
+var sameTopCart = window.AppData.restaurantId === window.AppData.sessionRestaurantId;
```
Atala: IMHO the variable name is misleading - it should be sthg like `isSessionCartForCurrentRestaurant`

js/app/cart/Cart.jsx
```diff
@@ -106,6 +118,10 @@ class Cart extends React.Component
     }).catch((err) => { console.log(err) });
   }
 
+  removeSessionTopCart() {
+    document.querySelector('a.btn-default').remove();
```
Atala: - the method name is misleading as you are not removing the cart in session, you are just removing an UI element. The session cart will be updated with the new restaurant ID when adding a new item.
- the selector used to get the element is too broad. Please give an ID to the element you want do remove and use `document.getElementbyId('<my_id>').remove()`

js/app/cart/CartTop.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+class CartTop extends React.Component
+{
+
+  render() {
+    return (
+      <a href={this.props.validateCartURL} className="btn btn-default navbar-btn navbar-right">
```
Atala: IMHO for now on we better redirect on restaurant page, instead on going for checkout right away when clicking on the top cart. It is because this icon is more used to resume shopping than to go to checkout. Please update `href` accordingly.

js/app/cart/index.jsx
```diff
@@ -18,6 +20,33 @@ var initialDate = localStorage.getItem('search__date') || window.AppData.Cart.da
     geohash = localStorage.getItem('search_geohash') || '',
     streetAddress = localStorage.getItem('search_address') || '';
 
+function addItemToBasket(event) {
```
Atala: 👍

src/AppBundle/Resources/translations/messages.fr.yml
```diff
@@ -144,6 +144,11 @@ demo.disclaimer.subtitle: Commandez, payez avec une carte de crédit de test et
 cart.widget.title: Vos détails de livraison
 cart.widget.button: Commander
 
+cart.modal.title: Attention
+cart.modal.content: Il n'est malheureusement pas possible de placer une commande dans plusieurs restaurants en même temps. Etes vous sur de vouloir continuer avec ce restaurant ?
```
Atala: Keep it simple : `Vous avez commencé votre commande dans un autre restaurant. Souhaitez vous commencer une nouvelle commande dans ce restaurant?"

src/AppBundle/Resources/translations/messages.en.yml
```diff
@@ -32,6 +32,12 @@ demo.disclaimer.subtitle: Follow the checkout process, pay with a test credit ca
 cart.widget.title: Your delivery details
 cart.widget.button: Order
 
+cart.modal.title: Warning
+cart.modal.content: Ordering in multiple restaurant at the same time is not possible. Are you sure you want to add this item in you cart ?
```
Atala: (same comment as in french)

js/app/cart/CartTop.jsx
Ladias-Elie: definitely better, but if we want to do like that we may have to change the <a href=https://github.com/coopcycle/coopcycle-web/blob/677839deb831159a66ff1d368b325a008a068f65/src/AppBundle/Controller/RestaurantController.php#L44> getCart </a> function. Because if the restaurant of the session cart and the one of the current restaurant page are the same then the cart is reseted which is not what we want in this situation.. What do you think ?

Atala:
Merged manually - thanks!

alexsegura:
There was still a problem in checkout. 

capture d ecran 2017-11-24 a 16 27 34

capture d ecran 2017-11-24 a 16 27 50

I rollbacked the merge & extracted code to the [no-silent-resto-change](https://github.com/coopcycle/coopcycle-web/compare/no-silent-resto-change) branch

PR #147, opened 2017-11-24
No silent resto change

Ladias-Elie:
I think the problem was coming from some unused span tag. 

I forgot to test all the way to the checkout sorry about that....


alexsegura:
Thanks! I'll check that during the weekend. 

Btw, I changed the base of the Pull Request to `master` to see all the changes.

js/app/cart/CartPortal.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import ReactDOM from 'react-dom';
+
+class CartPortal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.el = document.getElementById('cart-top');
```
alexsegura: Just to make sure I understand: the « portal » is what manages the modal, right? 
If so, shouldn't the container be the `<body>` tag?

Ladias-Elie: no, the portal manage the cart in the top right of the page. It make sure it share the same state as the other cart when a client is on a restaurant page.

alexsegura: Do we really need a portal for this?

Ladias-Elie: How would you do differently ? 

From what I understand, the portal allow you to render the element outside of the parent element tag. This is very interesting here because we want our TopCart element to share some state with the Cart element (total), but we want it to be rendered in the navigation bar.

You can find <a href=https://github.com/coopcycle/coopcycle-web/issues/95>here</a> more context about that issue.

Atala: I think that the portal is a neat solution for our problem. Both carts (the one on the right and the top one) share the same state, so the same total.

src/AppBundle/Resources/views/menu.html.twig
```diff
@@ -11,7 +11,7 @@
       <a class="navbar-brand" href="{{ path('homepage') }}">CoopCycle</a>
     </div>
 
-    <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
```
alexsegura: If you do that, you also need to modify the `data-target` of the toggle element.

alexsegura:
The problem in this PR is that is uses a workaround (the portal) to replace the element which is rendered on page load. IMHO let's go full-React for the cart in the navbar 🚀 

Some guidelines below.

#### Just use a `<div>` as a React container in `menu.html.twig`

We are going to render the cart in the navbar with React, so the following code should be replaced by a simple `<div id="top-cart"></div>` container:

```twig
{% set cart = app.session.get('cart') %}
{% set deliveryAddress = app.session.get('deliveryAddress') %}

{% if deliveryAddress is empty and cart is not empty %}
    <a id="top-cart" href="{{ path('restaurant', {'id': cart.restaurantId}) }}" class="btn btn-default navbar-btn navbar-right">
      {% trans %}Cart{% endtrans %}  <span class="glyphicon glyphicon-shopping-cart" aria-hidden="true"></span>  {{ cart.total }} €
    </a>
{% endif %}

{% if deliveryAddress is not empty %}
  <a id="top-cart" href="{{ path('order') }}" class="btn btn-default navbar-btn navbar-right">
    {% trans %}Order{% endtrans %}  <span class="glyphicon glyphicon-shopping-cart" aria-hidden="true"></span>  {{ cart.total }} €
  </a>
{% endif %}
```

#### Call a function when the cart is modified in `Cart.jsx`

When the cart is modified (i.e when an item is added/removed), call a function which is passed via `props`, and pass the total as a parameter. 

```javascript
removeItem(item) {
  // When the ajax call returns
  this.props.onCartChange(total);   
}

addMenuItemById(id, modifiers) {
  // When the ajax call returns
  this.props.onCartChange(total);
}
```

#### Rely on `state` instead of `props` in `CartTop.jsx`

The `CartTop` component now needs to get modified dynamically, so use `this.state.total` in the render method, and add a setter. 

```
constructor(props) {
  super(props);
  this.state = {
    total: props.total
  }
}
setTotal(total) {
  this.setState({ total })
}
```

#### Reference the `CartTop` component in `cart/index.jsx`:

For the `CartTop` component to be able to get notified of changes in the `Cart` component, they both need to be created in the `cart/index.jsx` file. 

```javascript
var topCartComponent = render(
  <CartTop restaurantUrl={restaurantUrl} total={total} />, 
  document.getElementById('top-cart')
);
```

#### Modify `CartTop` whenever `Cart` is modified

Now you can call the `setTotal` method when the `onCartChange` total is called 😉

```
<Cart
  onCartChange={total => topCartComponent.setTotal(total)} />
```

--- 

Of course, now the `js/cart.js` asset needs to be referenced on each page, not only on the restaurant page. 

I also missed the part about the modal, but let's do that first.

js/app/cart/CartPortal.jsx
alexsegura: Ok, I see what you did there 🤓
I added some guidelines on how to achieve this in the main conversation.

src/AppBundle/Resources/views/Restaurant/index.html.twig
```diff
@@ -100,9 +100,29 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+</div>
+
 {% endblock %}
 
 {% block scripts %}
+{% set sessionCart = app.session.get('cart') %}
```
Atala: This is not useful as the `cart` variable in the template [is the session cart](https://github.com/coopcycle/coopcycle-web/blob/2f547595a6a5bf1d0de89cb99b2b8a2f32658a4d/src/AppBundle/Controller/RestaurantController.php#L233)

src/AppBundle/Resources/views/menu.html.twig
Atala: Aggreed, I don't see why you need to change the ID here (I possibly misguided you on this).

Atala:
I partially agree: 
  - to render a Cart in the PHP template then remove it is ugly  and shouldn't be done.
I think however:
 - that React portals are the right way to go here. We need to have two elements which share the same state (i.e. CartTop is a Cart children). With your solution we have two elements with two states, which need to be kept in sync. With portal we don't have this issue
 - I don't think we need to add extra-JS on all other pages, as the cart in session can be rendered backend (no need to keep in sync while adding items in the Cart).

I propose something along these lines :  https://gist.github.com/Atala/8d17a895e08745714095c30153af519c

Ladias-Elie:
Thanks for the suggestion, maybe Alois solution is a good compromise and make us save a JS call. In the other hand I also like the idea of going full react, it may be more simple to understand when the cart logic in handle only in one place.

Anyway I don't have the knowledge to decide wish solution is the best so just tell me which one to implement =D

alexsegura:
Please, vote for me! 😬

Seriously, it's up to you. 

Maybe you could rebase the branch on master first?

src/AppBundle/Resources/views/Restaurant/index.html.twig
Ladias-Elie: I think it is. The cart variable from the template will be erased if the 'old' session cart is from a different restaurant. 
To be able to trigger the modal, and ask the user to choose if he want to continue with the old cart or create a new one we need to keep the old one.

Atala:
Hi Elie,

Let's go for @alexsegura option.

Steps to go : 
 - delete the things around the portal, keeping the `CartTop` element
 - use the code by @alexsegura to make it works on restaurant page
 - include the script `cart.js` on all pages (i.e. insert it on `base.html.twig`)
 - use the cart in session when the user is logged in to initialize the Cart component (means all the stuff like `window.AppData.Cart = ...` must go on `base.html.twig` page and used the in session cart, not the one passed as template parameter)

As always, ask for help in Slack if your stuck.

alexsegura:
@Ladias-Elie I rebased the branch on your fork (thank you for allowing me), and squashed everything into one commit to see clearly. Looking at the diff I think I didn't screw up everything 😅

I'm testing it right now.

src/AppBundle/Resources/views/base.html.twig
```diff
@@ -53,6 +53,35 @@
       window.AppData.countryIso = "{{ country_iso }}";
     </script>
     <script src="{{ asset('js/common.js') }}"></script>
+    {# Code for top cart #}
+    <script>
+    window.AppData = window.AppData || {};
+    {% set sessionCart = app.session.get('cart') %}
+    {% if sessionCart is not empty and sessionCart.restaurant %}
```
alexsegura: With the `if`, it means `window.AppData.Cart` can actually be `undefined` when there is no session. 
It is breaking some code which is relying on it. This is "normal" because in the previous code, the cart was always defined. 

I think it would be more reliable to always have `window.AppData.Cart` defined, even if there is no session. 
In this case, most of the properties would be initialized to `null` (like `date`)

alexsegura:
@Ladias-Elie you should rebase on master, I changed `Cart.jsx` again in 6c23ecb2cabff494488a54f0cc8443d187a9aa38. 

Git should manage the rebase easily.

Ladias-Elie:
@alexsegura I fixed the window.AppData problem by creating two different carts base on the fact that you are on a restaurant page or not. This way you always have a cart. 
This may be a lot of code duplication though...

@Atala I tried with your suggestion (https://gist.github.com/Atala/f38e5e80fbfb45e6b61b21b5430b9dc8), but it was causing some problems: For ex when initializing the cart with the session cart while we wanted it with the new restaurant.

I have tested the main scenario, it seems to work:

- redirecting to the corresponding restaurant when clicking on the top cart.
- pop up when shopping in a new restaurant
- top cart updated when adding an item to the cart

I still have a style problem, I can't manage to have a right margin for the top right cart. Some help would be great =) 
capture d ecran 2017-12-12 a 18 04 13

Maybe it would be also good to have the modal as a react element ??

And sorry for the new commit history mess up...

PR #158, opened 2017-12-03
Tax calculation

alexsegura:
This pull request is another prerequisite to be able to generate invoices. 

The most important change is that it introduces 3 new properties to `Order` & `Delivery`: 
- `totalExcludingTax`
- `totalTax`
- `totalIncludingTax`

They are calculated by `OrderManager::applyTaxes`. 
Also, the goal is to avoid knowing anything about how the prices are calculated, everything is processed by a listener before saving in database. 

Please note that `order.totalIncludingTax` & `delivery.totalIncludingTax` are kept separated, i.e. `order.totalIncludingTax` does **not** include the delivery price. 

To obtain the end user price, it is necessary to do:
```
order.totalIncludingTax + delivery.totalIncludingTax
```

### DIFF
--- a/src/AppBundle/Controller/OrderController.php
+++ b/src/AppBundle/Controller/OrderController.php
@@ -13,7 +13,6 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Component\HttpFoundation\Request;
 
-
 /**
  * @Route("/order")
  */
@@ -44,9 +43,7 @@ class OrderController extends Controller
 
         $delivery = new Delivery($order);
         $delivery->setDate($cart->getDate());
-        $delivery->setOriginAddress($restaurant->getAddress());
         $delivery->setDeliveryAddress($cart->getAddress());
-        $delivery->setPrice($restaurant->getContract()->getFlatDeliveryPrice());
 
         return $order;
     }
@@ -161,8 +158,8 @@ class OrderController extends Controller
      * @param Request $request
      *
      */
-    public function orderPublic($uuid, Request $request) {
-
+    public function orderPublic($uuid, Request $request)
+    {
         $orders = $this->getDoctrine()
             ->getRepository(Order::class)->findBy(['uuid' => $uuid]);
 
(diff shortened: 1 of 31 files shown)


src/AppBundle/Service/Taxation/FloatCalculator.php
```diff
@@ -0,0 +1,28 @@
...
+use Sylius\Component\Taxation\Model\TaxRateInterface;
+
+final class FloatCalculator implements CalculatorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function calculate(float $base, TaxRateInterface $rate): float
+    {
+        if ($rate->isIncludedInPrice()) {
```
Atala: Maybe add a comment ?
`if the base is all taxes included`

src/AppBundle/Entity/Order.php
```diff
@@ -275,6 +278,10 @@ public function setRestaurant(Restaurant $restaurant)
     {
         $this->restaurant = $restaurant;
 
+        if (null !== $this->delivery) {
+            $this->delivery->setPriceFromOrder($this);
```
Atala: Need to call `setOriginAddressFromOrder` as well?

src/AppBundle/Controller/OrderController.php
```diff
@@ -42,11 +41,11 @@ private function createOrderFromRequest(Request $request)
             $order->addCartItem($item, $menuItem);
         }
 
+        // TODO Use OrderManager::applyTaxes
```
Atala: Why not use it right away?

src/AppBundle/Entity/Order.php
alexsegura: Done. Also added a unit test.

src/AppBundle/Service/Taxation/FloatCalculator.php
alexsegura: This is already documented in `TaxRateInterface`

src/AppBundle/Controller/OrderController.php
alexsegura: Sorry wrong comment, this is actually useless here.

PR #159, opened 2017-12-04
feat: introduce docker-sync

Atala:
Introducing [docker-sync](http://docker-sync.io/). It is kind of shitty becaus it's a hack on top of docker but the dev experience is aaaaamzazing with it (response time ~200ms)

@alexsegura can you test it on your machine?

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -1,0 +1,19 @@
+version: '3'
+
+services:
+
+  webpack:
+    volumes:
+      - 'node_modules:/srv/coopcycle/node_modules'
+      - 'webpack-sync:/srv/coopcycle:nocopy'
+
+  php:
+    volumes:
+      - 'node_modules:/srv/coopcycle/node_modules'
+      - 'php-sync:/srv/coopcycle:nocopy'
+
+volumes:
+    php-sync:
+      external: true
+    webpack-sync:
+      external: true
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -72,6 +72,6 @@ COPY docker/php/start.sh /usr/local/bin/docker-app-start
 RUN chmod +x /usr/local/bin/docker-app-start
 
 # Needed to fix permissions error
-USER 1000
+USER 100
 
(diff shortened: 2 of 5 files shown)


docker/php/Dockerfile
```diff
@@ -72,6 +72,6 @@ COPY docker/php/start.sh /usr/local/bin/docker-app-start
 RUN chmod +x /usr/local/bin/docker-app-start
 
 # Needed to fix permissions error
-USER 1000
+USER 100
```
alexsegura: Doesn't this break standard `docker-compose`?

docker-compose-dev.yml
```diff
@@ -0,0 +1,19 @@
...
+services:
+
+  webpack:
+    volumes:
+      - 'node_modules:/srv/coopcycle/node_modules'
+      - 'webpack-sync:/srv/coopcycle:nocopy'
+
+  php:
+    volumes:
+      - 'node_modules:/srv/coopcycle/node_modules'
```
alexsegura: Why `node_modules`?

alexsegura:
It works, and it feels a bit faster, yes. 
I never had such performance problems on my machine. 

Side note: maybe we should specify in the `README` that it is needed to run `docker-compose build` when containers have changed?

:shipit:

docker/php/Dockerfile
alexsegura: Looks like `docker-compose` still works. 
I don't know why but it works 🤔

Atala:
I think even better would be sthg like [this](https://gist.github.com/mariusGundersen/8a4383f74a46ababc1de) -> run container build if Dockerfile as changed.

Should work with migrations, package.json, aso. too

docker-compose-dev.yml
Atala: That is a useless line.

docker/php/Dockerfile
Atala: I tested it work on my side.

The main issue I had with `docker-sync` is permission issues, it does not seem to work with an user ID of `1000` (I don't know why)

PR #160, opened 2017-12-05
Select address on cart

Atala:
Enable the address selection on Cart .

<strike>I still have an issue before merging : 
 - on restaurant list page the restaurant selection is based on a kilometer radius (`findNearby`)
 - on cart picker the validation is done with the "real" distance by bike

So a delivery address which is valid on restaurants list page may not be valid on the cart selector. Which is terrible?</strike>

@alexsegura do you get the issue? do you have any idea to deal with it?

### DIFF
--- a/src/AppBundle/Controller/RestaurantController.php
+++ b/src/AppBundle/Controller/RestaurantController.php
@@ -4,6 +4,7 @@ namespace AppBundle\Controller;
 
 use AppBundle\Entity\Address;
 use AppBundle\Entity\Base\GeoCoordinates;
+use AppBundle\Service\RoutingInterface;
 use AppBundle\Utils\Cart;
 use AppBundle\Entity\Menu\MenuItem;
 use AppBundle\Entity\Restaurant;
@@ -33,9 +34,10 @@ class RestaurantController extends Controller
 
     protected $logger;
 
-    public function __construct(LoggerInterface $logger)
+    public function __construct(LoggerInterface $logger, RoutingInterface $routing)
     {
         $this->logger = $logger;
+        $this->routing = $routing;
         $this->validator = Validation::createValidatorBuilder()
             ->enableAnnotationMapping()
             ->getValidator();
@@ -279,6 +281,10 @@ class RestaurantController extends Controller
 
         if ($request->request->has('address')) {
             $this->setCartAddress($request, $cart);
+            $origin = $cart->getRestaurant()->getAddress()->getGeo();
+            $destination = $cart->getAddress()->getGeo();
+            $distance = $this->routing->getDistance($origin, $destination);
+            $cart->setDistance($distance);
         }
 
         $errors = $this->validator->validate($cart, null, ["cart"]);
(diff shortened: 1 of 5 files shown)


Atala:
@alexsegura can you merge if it's working for you?

src/AppBundle/Controller/RestaurantController.php
```diff
@@ -274,12 +281,18 @@ public function addToCartAction($id, Request $request)
...
             $this->setCartAddress($request, $cart);
+            $origin = $cart->getRestaurant()->getAddress()->getGeo();
+            $destination = $cart->getAddress()->getGeo();
+            $distance = $this->routing->getDistance($origin, $destination);
+            $cart->setDistance($distance);
         }
 
         $errors = $this->validator->validate($cart, null, ["cart"]);
 
+//        var_dump($request->request->has('address'));
```
alexsegura: Please remove comment 💀

src/AppBundle/Resources/views/menu.html.twig
```diff
@@ -83,7 +83,7 @@
       {% set deliveryAddress = app.session.get('deliveryAddress') %}
 
       {% if deliveryAddress is empty and cart is not empty %}
-      <a href="{{ path('restaurant', {'id': cart.restaurantId}) }}" class="btn btn-default navbar-btn navbar-right">
+      <a href="{{ path('restaurant', {'id': cart.restaurant.id}) }}" class="btn btn-default navbar-btn navbar-right">
```
alexsegura: Causes an `Internal Server Error` when there is already a cart in session. 
We will need to delete sessions when deploying this.

src/AppBundle/Utils/Cart.php
```diff
@@ -51,23 +61,30 @@ class Cart
...
-        }
-
+        $this->restaurant = $restaurant;
         $this->address = new Address();
     }
 
     public function isForRestaurant(Restaurant $restaurant)
     {
-        return $this->restaurantId === $restaurant->getId();
+        return $this->getRestaurantId() === $restaurant->getId();
```
alexsegura: We can't compare instances directly because `$this->restaurant` is unserialized from session whereas `$restaurant` is managed by Doctrine, right?

src/AppBundle/Controller/RestaurantController.php
Atala: ☠️

src/AppBundle/Resources/views/menu.html.twig
Atala: I try to keep the old carts valid - looking into this.

If no solution I think it is ok to remove all in-session carts.

src/AppBundle/Utils/Cart.php
Atala: I try to keep the old carts valid - looking into this.

Atala:
I did 81487c28b0b7df094b9c68c86de92efadaa475f4 to fix this issue, but this is ugly. Do you have a better idea?

src/AppBundle/Controller/RestaurantController.php
Atala: Done

alexsegura:
Oh yes we will delete the sessions, don't worry.

Atala:
Done & merging .

PR #161, opened 2017-12-06
Introduce PDF invoices

alexsegura:
Damn simple introduction of PDF invoices based on calculated amounts with taxes. 

Can you test the PDF rendering @Atala ? 
You need to rebuild the PHP container `docker-compose build php`

- [x] Add properties in `LocalBusiness` entity
- [x] Add fields in `RestaurantType`
- [x] Add link to PDF in admin interface / user profile

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -12,6 +12,22 @@ RUN set -xe \
         icu-dev \
         zlib-dev \
         postgresql-dev \
+    && apk add --no-cache \
+            xvfb \
+            ttf-freefont \
+            fontconfig \
+            dbus \
+    && apk add qt5-qtbase-dev \
+            wkhtmltopdf \
+            --no-cache \
+            --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ \
+            --allow-untrusted \
+    && mv /usr/bin/wkhtmltopdf /usr/bin/wkhtmltopdf-origin \
+    && echo $'#!/usr/bin/env sh\n\
+Xvfb :0 -screen 0 1024x768x24 -ac +extension GLX +render -noreset & \n\
+DISPLAY=:0.0 wkhtmltopdf-origin $@ \n\
+killall Xvfb' > /usr/bin/wkhtmltopdf \
+    && chmod +x /usr/bin/wkhtmltopdf \
     && apk add --no-cache --virtual .build-deps $PHPIZE_DEPS \
     && docker-php-ext-configure pdo_pgsql --with-pgsql \
     && docker-php-ext-install \
@@ -48,10 +64,7 @@ WORKDIR /srv/coopcycle
 COPY composer.json ./
 COPY composer.lock ./
 
-RUN mkdir -p \
-        var/cache \
-        var/logs \
-        var/sessions \
+RUN mkdir -p var/cache var/logs var/sessions \
     && composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest \
     && composer clear-cache \
 # Permissions hack because setfacl does not work on Mac and Windows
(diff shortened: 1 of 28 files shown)


docker/php/Dockerfile
```diff
@@ -48,10 +64,7 @@ WORKDIR /srv/coopcycle
 COPY composer.json ./
 COPY composer.lock ./
 
-RUN mkdir -p \
-        var/cache \
-        var/logs \
-        var/sessions \
+RUN mkdir -p var/cache var/logs var/sessions \
```
Atala: Remove trailing `\`

src/AppBundle/Form/RestaurantType.php
```diff
@@ -34,27 +34,28 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
-                    'class' => 'AppBundle:Cuisine',
-                    'choice_label' => 'name',
-                    'query_builder' => function (EntityRepository $er) {
-                        return $er->createQueryBuilder('c')->orderBy('c.name', 'ASC');
-                    },
-                ),
-                'allow_add' => true,
-                'allow_delete' => true,
-            ))
+            // // FoodEstablishment
```
Atala: Why?

src/AppBundle/Resources/views/Order/pdf.html.twig
```diff
@@ -0,0 +1,87 @@
+<!DOCTYPE html>
```
Atala: I think the filename should be `invoice.html.twig` - it is actually a HTML tempalte.

src/AppBundle/Controller/AdminController.php
```diff
@@ -77,6 +79,24 @@ public function ordersAction(Request $request)
         );
     }
 
+    /**
+     * @Route("/admin/orders/{id}.pdf", name="admin_order_invoice", requirements={"id" = "\d+"})
```
Atala: I think this feature should be accessible to everyone (restaurants, customers too in `/profile`).
Lots of customers will need the pdf for their works invoiving.

src/AppBundle/Resources/views/Order/pdf.html.twig
```diff
@@ -0,0 +1,87 @@
...
+          </tbody>
+        </table>
+      </div>
+    </div>
+
+    <div style="clear: both;"></div>
+
+    <h2>{% trans %}Order{% endtrans %}</h2>
+
+    <table style="width: 100%;">
```
Atala: Should be sthg like `width: 800px;margin: 0 auto;` (nicer)

Atala:
Works for me -> nice job!

capture d ecran 2017-12-06 a 13 42 55

src/AppBundle/Controller/AdminController.php
alexsegura: Yep, this is in my TODOLIST

docker/php/Dockerfile
alexsegura: Why? This will break the multi-line 🤔

src/AppBundle/Resources/views/Order/pdf.html.twig
alexsegura: Yep.

alexsegura:
@Atala I have managed the SIRET number using [additional properties](http://schema.org/additionalProperty), as this is available only in France. 
Right now the implementation is kind of naïve, but at least differences between countries are managed. 

I have faced another problem: restaurant can't actually be saved because the `deliveryService` is `NULL` when using the default one. 
Also, errors display is not consistent when the form is displayed and when it is submitted. 
I'm investigating.

docker/php/Dockerfile
Atala: Forget it, I didn't look the whole file I thought it was the end of the line -_-

Atala:
Good luck with the Symfony forms! Ping if help needed

PR #164, opened 2017-12-09
Dispatch adress bug

Ladias-Elie:
I changed three things in the 'should dispatch order to closest courier' test case:

- The error is now raised as long as the wrong courrier receive a message. Before we were checking that the message type was 'order', I don't think that this is useful and furthermore if we want to check the message type I think it should be delivery.

- I removed the createRandomOrder function from the beforeEach one. In some test case we were creating two random order because of that. I am may have missed something here, not 100 % sure I should have removed it.

- I changed the courier location, see below

image

For the rest it is pretty straightforward, we were choosing the delivery address as the origin point while we should have chosen the restaurant address.

### DIFF
--- a/js/api/models/Courier.js
+++ b/js/api/models/Courier.js
@@ -83,11 +83,11 @@ Courier.prototype.toJSON = function() {
 
 Courier.nearestForDelivery = function(delivery, distance = 3500) {
 
-  var address = delivery.deliveryAddress;
+  var address = delivery.originAddress;
 
   return new Promise(function(resolve, reject) {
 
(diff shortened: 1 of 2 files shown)


js/api/models/Courier.js
```diff
@@ -83,11 +83,11 @@ Courier.prototype.toJSON = function() {
 
 Courier.nearestForDelivery = function(delivery, distance = 3500) {
 
-  var address = delivery.deliveryAddress;
+  var address = delivery.originAddress;
```
alexsegura: 😐

PR #165, opened 2017-12-11
Introduce StatsD, Graphite & Grafana

alexsegura:
This pull request introduces metrics managements with StatsD, Graphite & Grafana. 

Grafana is available through Docker on http://localhost:3000

grafana_example

**TODO**

- [x] Namespace metrics with app name 
- [x] Add some more useful metrics


alexsegura:
Waiting for this pull request to be merged to be able to decrement metrics 

M6Web/StatsdBundle#62

alexsegura:
⚠️ 

Actually, StatsD counters do not work in an 'intuitive" way, this can be confusing: StatsD flushes the counters every 10s (by default), so we can't use counters to "accumulate" values over time, unless we change the flush interval in Graphite. 

As I'm really new to StatsD/Graphite, I prefer using gauges instead.

alexsegura:
Until this issue is fixed we need to send data directly to StatsD (blocking but pretty fast) 

M6Web/Statsd#25

PR #168, opened 2017-12-20
Introduce manual dispatch.

alexsegura:
A really simple implementation of manual dispatch. 

**Administrator** can assign couriers to deliveries.

manual_dispatch_admin

**Courier** can see assigned deliveries assigned in personal account, and change the status. 

manual_dispatch_courier

This is just a first step before adding more cool stuff, but we need this at least to display it in the native app. 

ping @couleurmate

### DIFF
--- a/src/AppBundle/Resources/translations/messages.fr.yml
+++ b/src/AppBundle/Resources/translations/messages.fr.yml
@@ -128,6 +128,12 @@ Origin address: Adresse de départ
 Delivery address: Adresse de livraison
 Weight: Poids
 Show canceled orders: Afficher les commandes annulées
+Validate: Valider
+Please choose a courier in the list below: Veuillez choisir un coursier dans la liste ci-dessous
+The delivery will appear in his planning: La livraison apparaîtra dans son planning
+This delivery will be considered as picked: Cette livraison sera considérée comme récupérée
+This delivery will be considered as delivered: Cette livraison sera considérée comme livrée
+This action cannot be undone: Cette action ne peut pas être annulée
 
 profile.addresses.breadcrumb: Addresses
 profile.addresses.empty: Pas d'adresse sauvegardée pour le moment!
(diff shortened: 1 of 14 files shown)


src/AppBundle/Resources/translations/messages.fr.yml
```diff
@@ -172,11 +178,15 @@ order.status.CANCELED: Annulée
...
 delivery.status.WAITING: En attente
-delivery.status.DISPATCHED: Acceptée
+delivery.status.DISPATCHED: Assignée
 delivery.status.PICKED: Récupérée
 delivery.status.DELIVERED: Livrée
 delivery.status.CANCELED: Annulée
 
+delivery.status.DISPATCHED.verb: Assigner
+delivery.status.PICKED.verb: Récupérer
+delivery.status.DELIVERED.verb: Livrer
```
Atala: I think it should be `Marquer comme récupérée` et `Marquer comme livrée`

src/AppBundle/Resources/views/_partials/Delivery/actions.html.twig
```diff
@@ -0,0 +1,25 @@
+{% if is_granted('ROLE_ADMIN') and delivery.status == 'WAITING' %}
+  <button type="button" class="btn btn-xs btn-success"
+    data-toggle="modal"
+    data-target="#delivery-modal"
+    data-delivery="{{ delivery.id }}"
+    data-action="dispatch">
+    {% trans %}delivery.status.DISPATCHED.verb{% endtrans %}
+  </button>
+{% elseif is_granted('ROLE_COURIER') and delivery.status == 'DISPATCHED' %}
```
Atala: I think you should also allow admin to do this.

Use-case: an admin is checking all the deliveries at the end of the day and need to update the statuses

alexsegura: Actually `ROLE_ADMIN` inherits from `ROLE_COURIER` 😉 

https://github.com/coopcycle/coopcycle-web/blob/b05e11e7cc14f748a10e3eecd9efc8d8a94a8faa/app/config/security.yml#L13

couleurmate:
Better UX for a courier account to just see the dispatch on the homepage after login.

Atala:
Admin too could be redirected to dashboard page instead of homepage.

-> added to roadmap

PR #176, opened 2017-12-28
Save the cart in DB, not in session

Atala:
Positive side-effect : a new address is created with each cart so we now have a unique delivery address per order :)

### DIFF
--- a/app/DoctrineMigrations/Version20171228153019.php
+++ b/app/DoctrineMigrations/Version20171228153019.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types = 1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20171228153019 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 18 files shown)


app/DoctrineMigrations/Version20171228153019.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE cart_item DROP CONSTRAINT FK_F0FE25271AD5CDBF');
+        $this->addSql('DROP TABLE cart');
+        $this->addSql('DROP TABLE cart_item');
+        $this->addSql('COMMENT ON COLUMN zone.polygon IS \'(DC2Type:geojson)(DC2Type:geojson)\'');
```
alexsegura: Please remove. Still need to fix this... 😐

src/AppBundle/Resources/views/base.html.twig
```diff
@@ -59,7 +59,7 @@
     </script>
 
     <script src="{{ asset('js/common.js') }}"></script>
-    {% set cart = app.session.get('cart') %}
+    {% set cart = cart_provider.getCart() %}
```
alexsegura: 👍

app/DoctrineMigrations/Version20171228153019.php
Atala: ok

PR #186, opened 2018-01-18
fix: build for docker-sync

Atala:
Hi,

`docker-sync` use should not impact users not needing/using it. Can you test it doesnt break your local install ?


alexsegura:
Looks like this was merged in fa3340f35779e0a9edd52d935147f2d756fe25b8

PR #192, opened 2018-01-30
Link delivery and shop models

### LINKED ISSUE #188: Link Shop and Delivery models
- add a "deliveries list view" to shops
- add the possibility to link to a shop on `delivery/new` form page

Atala:
#Besides the linking, some improvements : 
fix: disable price selection for non-admin users
feat: enable admin to select a shop when creating a delivery
feat: display deliveries list for each store 

Fix #188


alexsegura:
I have fixed a few things and merged it, but there are still a few other things that will need to be fixed: 

- `/admin/stores/{id}/deliveries` displays the profile navbar
- On a store account, the user can fix the price freely if the price can't be calculated, it seems weird. Maybe we shouldn't even display the price, or have a fallback price? Let's think about that later. 
- I was thinking about a JOIN table `(store_id, delivery_id)` to avoid coupling Store & Delivery, but this may change in a near future so let's keep this like that.

PR #194, opened 2018-02-01
feat: disable automatic dispatch with a setting

Atala:


alexsegura:
Can you fix the tests? 
It looks like we need to initialize the database config for Travis.

alexsegura:
Actually, the problem is that on Travis, the automatic dispatch is tested by default, but it is disabled by default. Maybe for now we can disable those tests?

Atala:
I have the tests running locally but timeout on travis... -> I need to investigate (no clue at the moment)

alexsegura:
As we are going to deprecate this version of automatic dispatch, I close this PR. 

I hope you didn't lose to much time on this 😐

Atala:
The code is still relevant no?

alexsegura:
Yes, this is why I didn't delete the branch 😉

PR #198, opened 2018-02-09
Dashboard redux

Atala:
Still need to test the socket support


js/app/dashboard/store/reducers.js
```diff
@@ -0,0 +1,116 @@
...
+  let  userTasks = state[action.username],
+    newState = { ...state },
+    newUserTasks,
+    taskComparator = (taskA, taskB) => taskA['@id'] === taskB['@id']
+
+  switch(action.type) {
+    case 'ASSIGN_TASKS':
+      newUserTasks = userTasks.slice()
+      newState[action.username] = Array.prototype.concat(newUserTasks, action.tasks)
+      newState[action.username].duration = 0
```
alexsegura: Looks weird, is it an array or an object?

src/AppBundle/Controller/AdminController.php
```diff
@@ -230,15 +230,15 @@ public function dashboardFullscreenAction($date, Request $request)
...
-        });
-
-        usort($couriers, function (UserInterface $a, UserInterface $b) {
-            return $a->getUsername() < $b->getUsername() ? -1 : 1;
-        });
+        $couriers = $this->getDoctrine()
+            ->getRepository(ApiUser::class)
+            ->createQueryBuilder('u')
+            ->select("u.username")
+            ->where('u.roles LIKE :roles')
```
alexsegura: 💯

src/AppBundle/Resources/views/Admin/dashboardIframe.html.twig
```diff
@@ -62,40 +62,6 @@
     </aside>
 </div>
 
-<div class="modal fade" id="user-modal" tabindex="-1" role="dialog" aria-labelledby="user-modal-label">
```
alexsegura: Is it really better? Also, the modal has a different look'n'feel.

js/app/dashboard/store/reducers.js
```diff
@@ -0,0 +1,116 @@
...
+      break
+    case 'SAVE_USER_TASKS_SUCCESS':
+      newState[action.username].duration = action.duration
+      newState[action.username].distance = action.distance
+      break
+    case 'UPDATE_TASK':
+      if (action.task.assignedTo) {
+        userTasks = state[action.task.assignedTo]
+        let index =  _.findIndex(userTasks, action.task, taskComparator)
+        newState[action.task.assignedTo] = Array.prototype.splice(index, 1, action.task)
```
alexsegura: When the 2nd parameter is greater than 0, `Array.prototype.splice` _removes_ elements. Also, `Array.prototype.splice` returns the number of deleted elements, not the array. 

I think it should be:

```
userTasks.splice(index, 0, action.task)
newState[action.task.assignedTo] = userTasks
```

js/app/dashboard/store/reducers.js
```diff
@@ -0,0 +1,116 @@
...
+        _.intersectionWith(newState, action.tasks, taskComparator),
+        taskComparator
+        )
+      break
+    case 'REMOVE_TASKS':
+      newState = Array.prototype.concat(newState, action.tasks)
+      break
+    case 'UPDATE_TASK':
+      let index =  _.findIndex(state, action.task, taskComparator)
+      newState = Array.prototype.splice(index, 1, action.task)
```
alexsegura: Same problem as above with `Array.prototype.splice`

alexsegura:
Nice move 🚀

The websocket doesn't work (problem when updating tasks), and also the (badly-named) TaskList is never updated (the polyline never changes)

alexsegura:
I changed the queries to use `task.doneAfter`, because I created a task that should be completed between 23:30 and 00:30 on the next day, and they were not displayed. 

This is an extreme use case, but do you think it makes sense?

alexsegura:
Actually, what I did doesn't really make sense: `task.doneBefore` is actually most importent. 
I'll fix that in #200

js/app/dashboard/store/reducers.js
Atala: A JS array is indeed an object `typeof([]) -> "object`. You can assign arbitrary attributes to it (here distance, duration) and use it as an Array at the same time. I think it does the job nicely here.

src/AppBundle/Resources/views/Admin/dashboardIframe.html.twig
Atala: I think it's better to have everything inside the "react world" because the components should be reusable anywhere + handle modal state with Redux.

js/app/dashboard/store/reducers.js
alexsegura: Yes, but when you iterate over the values, how do you distinguish between tasks, and duration/distance?

Atala:
Fixing websocket issues

Atala:
I have fixed the issue with websocket updates. I checked the code and it appears we don't have automatic in-app refresh on the `TaskPage`. Should we add this ? It should be easy enough.

It seems to be that `TaskList` is correctly updated (at least now, but I didn't see a bug).

js/app/dashboard/store/reducers.js
Atala: `newState[action.username]` **is** a javascript array so if you iterate over it you will iterate over the tasks. It has also all the array methods. It has the extra-attributes `distance` and `duration`.

In pseudo-PHP code it may look like this :
```php
class TaskArray extends Array {
     protected duration;
     protected distance;
}
```

Atala:
I think everything's alright.

There is still one last "bug" (that was maybe here before) : each time we assign a task an `ASSIGN` event is generated for all task of the TaskList.

I will add a check before generating the event (= check that the last event was not an ASSIGN for the same user I think)

js/app/dashboard/store/reducers.js
alexsegura: I didn't know this was possible 🙂
I think it looks more "natural" like this: 

```
newState[action.username].tasks = action.tasks
newState[action.username].duration = action.duration
newState[action.username].distance = action.distance
```

alexsegura:
I fixed the problem in the task eventing system, it was dispatching deliveries twice. 
I'm improving some stuff, refreshing the polyline & let's merge!
I will need to rebase my pull request when this is merged.

alexsegura:
I ended managin `UserPanel` & the map with React/Redux. 
It feels simpler & more decoupled, now we can add more dashboard features without making spaghetti code 🙂

PR #209, opened 2018-02-13
fix: dispatch task events to concerned couriers

Atala:


src/AppBundle/Service/TaskManager.php
```diff
@@ -64,10 +63,10 @@ public function assign(Task $task, UserInterface $user, $position = null)
         });
 
         foreach ($tasks as $taskToAssign) {
-            $taskToAssign->assignTo($user, $position++);
-
             $isAssignedToSameUser = $taskToAssign->isAssigned() && $taskToAssign->isAssignedTo($user);
+
             if (!$isAssignedToSameUser) {
+                $taskToAssign->assignTo($user, $position++);
```
alexsegura: `TaskManager::assign` is also called when reordering tasks already assigned to the same user. 
`Task::assignTo` needs to be outside the `if`

js/api/dispatch-v2/index.js
```diff
@@ -17,15 +17,34 @@ console.log('NODE_ENV = ' + process.env.NODE_ENV)
...
-  metrics,
-  redis,
   pub,
-  sub,
   sequelize
 } = require('./config')(ROOT_DIR)
 
 const db = require('../Db')(sequelize)
 
+var ConfigLoader = require('../ConfigLoader')
```
alexsegura: Configuration is already loaded in `./config`, why did you need to do this?

js/api/dispatch-v2/index.js
```diff
@@ -43,14 +62,42 @@ var wsServer = new WebSocketServer({
     },
 })
 
-let isClosing = false
+let isClosing = false,
+    couriersList = {}
+
+const sub = require('../RedisClient')({
```
alexsegura: Same comment as above, Redis subscriber is created in `./config`.

js/api/dispatch-v2/index.js
```diff
@@ -43,14 +40,42 @@ var wsServer = new WebSocketServer({
...
+
+sub.on('subscribe', (channel, count) => {
+  if (count === channels.length) {
+    sub.on('message', function(channelWithPrefix, message) {
+      const parsedMessage = JSON.parse(message),
+            username = parsedMessage.user.username
+      _.each(channels, (channel) => {
+        if (sub.isChannel(channelWithPrefix, channel) && couriersList[username]) {
+          winston.debug(`Sending message ${message} to ${username}`)
+          couriersList[username].send(message)
```
alexsegura: I don't understand how you distinguish between assign/unassign on the client..
Maybe it would be better to also send the channel name?

```
couriersList[username].send({
  type: channel,
  data: message
})
```

Atala: Yes indeed I made this fix locally, forgot to push afterwards -_-

alexsegura: Push to master :shipit:

PR #221, opened 2018-02-20
Introduce task groups

alexsegura:
This PR introduces **task groups**: a task may belong to a group, to apply bulk actions on it. 
For now, groups can only be drag'n'dropped to assign all tasks at once. 

A new button allows to display grouped tasks, or all tasks at once. 

The drag'n'drop management has also been simplified: linked tasks are not grouped together via a DOM element, but they are highlighted when dragged. 

task_groups

### DIFF
--- a/assets/css/dashboard.scss
+++ b/assets/css/dashboard.scss
@@ -64,13 +64,20 @@ html, body {
     flex: 1;
     -webkit-flex-direction: column;
     flex-direction: column;
-    > h3, > h4 {
+    > h4 {
       padding: 15px;
       margin: 0;
       background-color: #95A5A6;
       color: #fff;
+      a {
+        color: #fff;
+        &:hover, &:active {
+          color: darken(#fff, 15%);
+          text-decoration: none;
+        }
+      }
     }
-    .dashboard__panel__scroll {
+    &__scroll {
       overflow: auto;
     }
   }
@@ -107,16 +114,6 @@ html, body {
       flex: 1;
       -webkit-flex-direction: column;
       flex-direction: column;
-    }
-
-    .dashboard__panel {
-      h4 > a {
-        color: #fff;
-        &:hover, &:active {
-          color: darken(#fff, 15%);
-          text-decoration: none;
-        }
-      }
     }
 
     .dropzone {
@@ -158,6 +155,10 @@ html, body {
   padding-right: 60px;
   cursor: pointer;
   color: #fff;
+  border-radius: 0;
+  &:last-child {
+    border-radius: 0;
+  }
   a {
     color: #fff;
   }
@@ -187,7 +188,6 @@ html, body {
   }
 }
 
-
 .task__icon {
   font-size: 16px;
   line-height: 16px;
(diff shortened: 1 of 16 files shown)


assets/css/dashboard.scss
```diff
@@ -64,13 +64,20 @@ html, body {
     flex: 1;
     -webkit-flex-direction: column;
     flex-direction: column;
-    > h3, > h4 {
+    > h4 {
```
Atala: We should avoid selectors that target HTML elements and bootstrap classes.

You could add on the folder list-items : 
```css
cursor: grab;
-webkit-cursor: grab;
```

Atala: We should avoid selectors that target HTML elements and bootstrap classes.

You could add on the folder list-items : 
```css
cursor: grab;
-webkit-cursor: grab;
```

js/app/dashboard/app.js
```diff
@@ -40,28 +40,35 @@ const drake = dragula({
 /**
  * Code to handle drag and drop from unassigned tasks to assigned
  */
-function configureOnDrop(allTasks, assignTasks) {
+function configureDrake(unassignedTasksContainer, allTasks, assignTasks) {
```
Atala: Maybe it would be nice to have the code just above inside this function. And then sthg like afterwards : `let drake = configureDragula()`

js/app/dashboard/app.js
```diff
@@ -40,28 +40,35 @@ const drake = dragula({
...
-          .toArray()
-      } else {
-        const task = _.find(allTasks, task => task['@id'] === $(element).data('task-id'))
-        tasks.push(task)
+      const isTask = element.hasAttribute('data-task-id')
+      const elements = isTask ? [ element ] : Array.from(element.querySelectorAll('[data-task-id]'))
+
+      let tasks = elements.map(el => _.find(allTasks, task => task['@id'] === el.getAttribute('data-task-id')))
+
+      // Make sure linked tasks are assigned too
```
Atala: Nice to have comment here - I think you should add more comments to your code ;)

<strike>I am not sure this code belongs here - shouldn't it be in Redux?</strike>

js/app/dashboard/components/TaskList.js
```diff
@@ -1,42 +1,89 @@
 import React from 'react'
-import { findDOMNode } from 'react-dom'
 import _ from 'lodash'
 import Task from './Task'
 import TaskGroup from './TaskGroup'
 import { connect } from 'react-redux'
-
+import { setTaskListGroupMode } from '../store/actions'
 
 class TaskList extends React.Component {
```
Atala: Maybe we should rename this to `UnassignedTaskList` ?

js/app/dashboard/app.js
alexsegura: This can't be in Redux: the `assignedTasksByUser` reducers only "knows" about the assigned tasks, not all of them. 
Ok for the comments ✅

js/app/dashboard/app.js
alexsegura: The code above doesn't need to know about Redux related stuff, this is why I did it like this.

js/app/dashboard/components/TaskList.js
alexsegura: The ["naming things is hard" label](https://github.com/coopcycle/coopcycle-web/labels/naming%20things%20is%20hard) is here for that 😍  

I created #223

PR #228, opened 2018-02-22
Write first test for Task entity

Ladias-Elie:


tests/AppBundle/Entity/TaskTest.php
```diff
@@ -68,4 +69,24 @@ public function testUnassign()
...
+
+    public function testHasEvent()
+    {
+        $event = new TaskEvent($this->task, "PICKUP");
+        $this->task->getEvents()->add($event);
+        $this->assertTrue($this->task->hasEvent("PICKUP"));
+        $this->assertFalse($this->task->hasEvent("DROPOFF"));
+    }
+
+    public function testGetFirstEvent()
```
alexsegura: 👍

alexsegura:
Thanks 😊

What would be nice would be testing `TaskManager`. 
For example, testing that `TaskManager::assign` doesn't retrigger a `TaskAssignEvent` when a `Task` is already assigned with the same user. 

Are you comfortable with mocking? If you want I can write a test to bootstrap everything.

PR #233, opened 2018-02-23
Introduce tagging system

alexsegura:
This PR introduces a tagging system that can be used to tag **any** resource. 

It is inspired by [anh/doctrine-extensions-taggable](https://github.com/hilobok/doctrine-extensions-taggable) & [fpn/doctrine-extensions-taggable](https://github.com/FabienPennequin/DoctrineExtensions-Taggable). 
I didn't use them because they are too old and look unmaintained. 

Both tasks & groups can be tagged.
When a task or a group is tagged, the marker takes the tag's color on the map. 
When there are several tags, the first one is picked to determine the color. 

@couleurmate what do you think? 
I tried to make something simple enough to manage any use case. 

**TODOLIST**

- [x] Add a screen to manage tags
- [x] Add a form field to edit tags in modal
- [x] Add Doctrine migration
- [x] Manage tags when importing tasks

task_tags

### DIFF
--- a/app/DoctrineMigrations/Version20180303201037.php
+++ b/app/DoctrineMigrations/Version20180303201037.php
@@ -1,0 +1,36 @@
+<?php declare(strict_types = 1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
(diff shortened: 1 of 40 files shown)


couleurmate:
Pretty neat ! It's all that we need. 

That could be great if we can set a tag for a specific resto/shop directly in its profile, and every new task get automatically that tag.

alexsegura:
@Atala you can review & merge if this is ok.

Atala:
Nice, will review tomorrow.

PR #241, opened 2018-02-28
TaskList is a TaskCollection

alexsegura:
This is the second step towards a clean generalization of tasks. 

The `TaskList` entity (i.e the tasks assigned to a courier on a given day) becomes a `TaskCollection`, 
It means the position is stored in the `TaskCollectionItem` entity, making the `TaskAssignement` entity useless. 
It also means the assigned user will be stored directly in the `Task` entity. 

The `TaskCollectionSubscriber` class has a new responsibility: update duration & distance when items of a `TaskCollection` have been modified. 

This is a big change, I still need to make sure nothing is broken, and introduce tests, at last. 

**TODO**

- [x] Add a `task.assigned_to` field
- [x] Drop table `task_assignment`
- [x] Detect assign / unassign events in Doctrine listener
- [x] Add `TaskCollectionInterface::getPolyline()`


alexsegura:
@Atala I think it is finished. 

More details on what has changed: 

The `TaskList` entity is now really more easy to use. 
It corresponds to the daily tasks for a courier, and inherits all the cool stuff from `TaskCollection`. 
Also, it means that when you add / remove a `Task` from a `TaskList`, it is assigned / unassigned. 

Everything is now managed through Doctrine listeners, I think it is cleaner. 

Also, now a `TaskList` corresponds to what is called a `UserPanel` on the dashboard components, when this is merged I will 

I have tested it, I think this is stable now. I'd like to merge this soon, so that we can continue. 
Can you test it and review? Thanks.

alexsegura:
I don't know why but Travis doesn't fails on [this build](https://travis-ci.org/coopcycle/coopcycle-web/builds/348275605)

Atala:
Nice, will review tomorrow.

alexsegura:
> For example Delivery could just link to TaskList one-to-on (maybe it needs some heavy refactoring?).

It can't be, because they could be sorted differently. 

In the example below, the tasks from `Delivery #1` & `Delivery #2` are reordered in the `TaskList`.
`Task #5` & `Task #6` are not associated with any `Delivery`, but they are in the `TaskList` because it makes sense it terms of dispatching. 

capture d ecran 2018-03-05 a 15 31 08

alexsegura:
Also, what may be surprising is that `duration`, `distance` & `polyline` are still stored in the `delivery` & `task_list` tables, but I didn't want to break everything at once 🙂
When this is merged I will move those shared properties in the `task_collection` table.

PR #275, opened 2018-03-14
Dashboard Improvments

Atala:


src/AppBundle/Entity/TaskCollectionItem.php
```diff
@@ -11,8 +11,9 @@
  *
  * @ORM\Entity
  * @ORM\Table(name="task_collection_item", uniqueConstraints={
- *   @ORM\UniqueConstraint(name="task_collection_item_unique", columns={"parent_id", "task_id"})}
- * )
+ *   @ORM\UniqueConstraint(name="task_collection_item_unique", columns={"parent_id", "task_id"}),
+ *   @ORM\UniqueConstraint(name="task_collection_position_unique", columns={"parent_id", "position"})
```
Atala: @alexsegura agree with this?

I think there are bugs that can cause items to have duplicate position in the same list (for instance two tasks at position 2)
-> need investigation and tests
```sql
select * from task_collection_item where task_id between 443 AND 450 order by position;
 id  | parent_id | task_id | position
-----+-----------+---------+----------
 278 |        24 |     446 |        0
 280 |        24 |     448 |        2
 276 |        24 |     444 |        4
 281 |        24 |     449 |        4
 279 |        24 |     443 |        4
 282 |        24 |     450 |        5
 301 |        24 |     447 |       24
```

src/AppBundle/Entity/TaskCollectionItem.php
```diff
@@ -11,8 +11,9 @@
  *
  * @ORM\Entity
  * @ORM\Table(name="task_collection_item", uniqueConstraints={
- *   @ORM\UniqueConstraint(name="task_collection_item_unique", columns={"parent_id", "task_id"})}
- * )
+ *   @ORM\UniqueConstraint(name="task_collection_item_unique", columns={"parent_id", "task_id"}),
```
Atala: As a side-note I think if we had separate models for `Delivery` and `TaskList` we would have a single unique constraint on `task_id` which will make things simpler in the long run (task can only be assigned once to a courier will be enforced at the database level).

src/AppBundle/Entity/TaskCollectionItem.php
```diff
@@ -3,7 +3,6 @@
 namespace AppBundle\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
-use Gedmo\Mapping\Annotation as Gedmo;
```
alexsegura: Annotations on properties should also be removed

tests/AppBundle/Entity/TaskListTest.php
```diff
@@ -32,4 +32,156 @@ public function testAddRemoveTaskChangesAssignedUser()
         $taskList->removeTask($task);
         $this->assertNull($task->getAssignedCourier());
     }
+
+    public function testUpdateTasksInsertTask()
```
alexsegura: 💯

src/AppBundle/Entity/TaskList.php
```diff
@@ -81,4 +81,30 @@ public function removeTask(Task $task)
 
         return parent::removeTask($task);
     }
+
+    public function updateTasks($tasksToAssign)
```
alexsegura: IMHO, having a `SplObjectStorage` as parameter is too specific. 
The `updateTasks` method should take a zero-indexed `array`, and use `SplObjectStorage` internally.

Atala: Ok will modify

src/AppBundle/Entity/TaskCollection.php
```diff
@@ -58,25 +58,65 @@ public function getItems()
...
+    public function getMaxPosition() {
+        $maxPosition = -1;
+        foreach ($this->getItems() as $i) {
+            $maxPosition = $i->getPosition() > $maxPosition ? $i->getPosition() : $maxPosition;
+        }
+
+        return $maxPosition;
+    }
+
     public function addTask(Task $task, $position = null)
```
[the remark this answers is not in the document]
alexsegura: Ok, I'll review tomorrow. 
At the beginning it was managed by hand, but I thought it would be easier to use the behavior, my bad 😐
I think it is the `Sortable` behavior which is buggy and unmaintained, this library is kind of a reference for other things.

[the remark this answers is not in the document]
Atala: [ref](https://github.com/Atlantic18/DoctrineExtensions/issues?utf8=%E2%9C%93&q=is%3Aopen+label%3Abug+sortable)

alexsegura:
I had the following React warning

capture d ecran 2018-03-22 a 20 30 13

So I fixed it in 223efe27ff6ead3162c2758cb540b44d29969907

Issue #155, opened 2017-12-01
Use YAML mapping for Doctrine instead of annotations


alexsegura:
We are going to introduce [Sylius](http://sylius.org/) to manage standard e-commerce stuff, and save us a lot of time. 

Sylius relies on [YAML mapping](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/yaml-mapping.html) to manage Doctrine entities, and be able to [customize models](http://docs.sylius.org/en/latest/customization/model.html). 

With the `doctrine:mapping:convert` command it should be very easy.

PR #276, opened 2018-03-15
Use Sylius to manage delivery orders

alexsegura:
This PR is a refactoring of what has been done in 29d967fdbc783374135baeccf023aeda65ad99ed to embed a form via an `<iframe>`, to order a delivery. 

It introduces [Sylius](https://github.com/Sylius/Sylius) to manage orders. 

Until #155 is fixed, we cannot properly extend the Sylius models (for example, to link a customer with `OrderInterface`, or even to have our own table names), so I created 2 temporary tables as a workaround: 
- a table to link a Sylius `OrderInterface` with a customer
- a table to link a Sylius `OrderItemInterface` with a delivery

Please note that at the moment, **Sylius orders only contain deliveries**. 
I have started moving Doctrine annotations to YAML, so we will be able to remove those tables shortly, and migrate data properly. 

The process is as before, with a few improvements.

When a customer places an order, it is creates a Sylius `OrderInterface` with `state = cart`, with one `OrderItemInterface` linked to a `Delivery`.

capture d ecran 2018-03-15 a 13 32 04

Administrators are notified and need to confirm the order.

capture d ecran 2018-03-15 a 13 32 42

--- 

When the order is confirmed, the customer receives an email with a link to a page where the order can be paid by credit card. 

capture d ecran 2018-03-15 a 13 49 00


app/AppKernel.php
```diff
@@ -13,8 +13,11 @@ public function registerBundles()
             new Symfony\Bundle\TwigBundle\TwigBundle(),
             new Symfony\Bundle\MonologBundle\MonologBundle(),
             new Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle(),
-            new Sylius\Bundle\TaxationBundle\SyliusTaxationBundle(),
             new Sylius\Bundle\ResourceBundle\SyliusResourceBundle(),
+            new Sylius\Bundle\TaxationBundle\SyliusTaxationBundle(),
+            new Sylius\Bundle\MoneyBundle\SyliusMoneyBundle(),
```
Atala: Multi-currency support

app/DoctrineMigrations/Version20180314160032.php
```diff
@@ -0,0 +1,39 @@
...
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180314160032 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE TABLE delivery_order (order_id INT NOT NULL, user_id INT NOT NULL, PRIMARY KEY(order_id, user_id))');
```
Atala: Nice to separate delivery/eshop order into tables

app/config/state_machine.yml
```diff
@@ -0,0 +1,68 @@
...
+            refunded: ~
+        transitions:
+            create:
+                from: [cart]
+                to: new
+            process:
+                from: [new]
+                to: processing
+            complete:
+                from: [new, processing]
```
Atala: Maybe match with stripe statuses ? Sthg like `captured` or so

app/DoctrineMigrations/Version20180314160032.php
alexsegura: This is actually a workaround 🙂
In the base Sylius Order component, an Order is not linked to anything. 
However, it totally makes sense to link an Order with a Customer. 

When #155 is fixed, we will be able to extend Sylius models, for example to implement [`CustomerAwareInterface`](https://github.com/Sylius/Sylius/blob/f2042d449c5ae3b68df145910cb42fae6c90d254/src/Sylius/Component/Customer/Model/CustomerAwareInterface.php)

src/AppBundle/Controller/PublicController.php
```diff
@@ -0,0 +1,123 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * @Route("/{_locale}/pub", requirements={ "_locale": "%locale_regex%" })
+ */
+class PublicController extends Controller
+{
+    /**
+     * @Route("/d/{number}", name="public_delivery")
```
Atala: `uuid` is better maybe?

app/AppKernel.php
alexsegura: Yep, but it is just there because `Sylius\Payment` needs it. 
Later, we will be able to manage currencies properly 🙂

src/AppBundle/Entity/DeliveryOrder.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
Atala: Why not put it in the `Entity/Delivery` folder?

src/AppBundle/Resources/views/Pdf/delivery.html.twig
```diff
@@ -0,0 +1,94 @@
+<!DOCTYPE html>
+<html>
```
Atala: I think to do clean invoices we will need the settings for admin/coop details (adress, siret, aso.).

Atala:
Hey !
Looks nice to me :) I added some comments but didn't do the full review yet.
I will test the process tomorrow on my machine

src/AppBundle/Controller/PublicController.php
alexsegura: When an `OrderInterface` goes from `state=cart` to `state=new`, it uses the default [`OrderNumberAssigner`](https://github.com/Sylius/Sylius/blob/f2042d449c5ae3b68df145910cb42fae6c90d254/src/Sylius/Bundle/OrderBundle/NumberAssigner/OrderNumberAssigner.php) class. 
I didn't change anything to the defaults, but for sure it can be overriden, and the `order.number` field is a `VARCHAR(255)`. It actually generates _strings_ like `000000001`.

Atala:
First of all contragulations for the big amount of work put into this!
I hit two major bugs in the ordering process:
 - bug "phone number is null" in admin before validating delivery (cf screenshot in slack)
 - ~as a customer I didn't received the second mail ("delivery order confirmed") altough everything seems fine in admin~  there is a problem with delivery order confirmation, after clicking the button "confirmed" I lose the button but dont receive the mail + still asked to confirm. Screenshots : 
capture d ecran 2018-03-16 a 16 47 44
capture d ecran 2018-03-16 a 16 47 52


Other remarks : 
- Wording : Nous allons vous envoyer un email d'ici quelques minutes. -> Nous allons vous envoyer un email d'ici quelques minutes, dès que la commande est validée.
- Visual feedback in admin when delivery order is confirmed (green alert : "Commande confirmée : à vos vélos!")
- No breadcrumb needed (or make it not clickable) as the user is not supposed to go back in the order process (I think? or does it work?)

I also think the UX for the admin is kind of confusing (what is a delivery? a delivery order? how to access both?), we should think twice about it - but not a problem right now.

alexsegura:
> bug "phone number is null" in admin before validating delivery (cf screenshot in slack)

Fixed, I was not storing the phone number at all, my bad. 

> after clicking the button "confirmed" I lose the button but dont receive the mail + still asked to confirm

Fixed, once confirmed the warning alert turns into an info alert

capture d ecran 2018-03-17 a 13 21 43

> I also think the UX for the admin is kind of confusing (what is a delivery? a delivery order? how to access both?), we should think twice about it - but not a problem right now.

Yes, sure, this is the next step. 
Of course the goal is to manage all kind of orders (meals & logistics) the same way, & display them on the same admin screen. 
It should be easy to distinguish between orders using [Single Table Inheritance](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html#single-table-inheritance).

PR #277, opened 2018-03-17
rollback from latest version of osrm to 5.9.2

Ladias-Elie:


alexsegura:
Oh you are right, we have freezed the version in production. 

However we are using `5.16.0` in production, can you set this version?

Also, the file `package-lock.json` shouldn't have changed.

Ladias-Elie:
I tried several of them but 5.9.2 seems to be the only one working for me...

Here is the log for the 5.16.0:

```
osrm_1      | [warn] Missing/Broken File: /data/data.osrm.maneuver_overrides
osrm_1      | [error] Required files are missing, cannot continue
coopcycleweb_osrm_1 exited with code 1
```

any suggestions ?

alexsegura:
Have you tried compiling it from scratch?

```
docker-compose build osrm
make osrm
```

Ladias-Elie:
ok it worked. I close this PR, I will include the 5.16.0 in my next one.

PR #285, opened 2018-04-03
fix: fix name collision when twice same tax rate

Atala:


alexsegura:
Merged in fe27c3ec8fe36f5ef7d2d5a480f2649a3620eb55

PR #299, opened 2018-04-10
fix: add missing transition to payment

Atala:
Can you check it's ok?


Atala:
It's a mistake, closing!

PR #315, opened 2018-04-17
handle failed payments

Atala:
I have to deconstruct the cascade thing.

I don't think it was possible to retry payments with this state machine, because we need to dissociate order creation and payment creation. Plus creating the order at the step before it's cleaner, we then have an unique payment URL per order, we may implement later a [forgotten cart feature](http://shopifynation.com/marketing/shopify-abandoned-cart-orders-feature/)

### DIFF
--- a/app/config/state_machine.yml
+++ b/app/config/state_machine.yml
@@ -42,14 +42,6 @@ winzou_state_machine:
                     do: ["@coopcycle.order_manager", "capturePayment"]
                     args: ["object"]
             after:
-                after_create_create_payment:
-                    on: ["create"]
-                    do: ["@sm.callback.cascade_transition", "apply"]
-                    args: ["object.getLastPayment('cart')", "event", "'create'", "'sylius_payment'"]
-                after_create_authorize_payment:
-                    on: "create"
-                    do: ["@coopcycle.order_manager", "authorizePayment"]
-                    args: ["object"]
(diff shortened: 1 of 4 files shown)


alexsegura:
I checked how it is done in Sylius Core. 

In the [`sylius_payment.yml`](https://github.com/Sylius/Sylius/blob/9cd567a1dee978af0fa775aabc0401073e64dd40/src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_payment.yml#L38), when a payment fails, they call [`OrderPaymentProcessor::process`](https://github.com/Sylius/Sylius/blob/190a314d7e84d36093ec2736111dabc713c6b7bf/src/Sylius/Component/Core/OrderProcessing/OrderPaymentProcessor.php#L52)
The service is configured [here](https://github.com/Sylius/Sylius/blob/f7d42d2ce64288407372775e0ed421debcd50cd3/src/Sylius/Bundle/CoreBundle/Resources/config/services/order_processing.xml#L64)

Their implementation of `OrderPaymentProcessor` does this: 
- Each instance is configured with a "target state" for the payment. 
- If the last payment for the order is not in the expected target state, it creates a new payment. 

The implementation I added is really naïve, it only adds a payment in state `cart`
I need to think about it a little more, all of the stuff with state machine is becoming complex and we have barely no tests 😥

PR #318, opened 2018-04-20
Improve list display for deliveries

Atala:
Use colors to show clearly linked tasks (a.k.a deliveries for the moment)

capture d ecran 2018-04-20 a 18 24 44

capture d ecran 2018-04-20 a 18 24 19

### DIFF
--- a/src/AppBundle/Entity/Delivery.php
+++ b/src/AppBundle/Entity/Delivery.php
@@ -45,6 +45,8 @@ class Delivery extends TaskCollection implements TaskCollectionInterface
     const VEHICLE_BIKE = 'bike';
     const VEHICLE_CARGO_BIKE = 'cargo_bike';
 
+    const COLORS_LIST = ['#213ab2', '#b2213a', '#5221b2', '#93c63f', '#b22182', '#3ab221', '#b25221', '#2182b2', '#3ab221', '#9c21b2', '#c63f4f', '#b2217f', '#82b221', '#5421b2', '#3f93c6', '#21b252', '#c6733f'];
+
     /**
      * @Groups({"delivery"})
      */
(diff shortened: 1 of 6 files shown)


src/AppBundle/Entity/Delivery.php
```diff
@@ -45,6 +45,8 @@ class Delivery extends TaskCollection implements TaskCollectionInterface
     const VEHICLE_BIKE = 'bike';
     const VEHICLE_CARGO_BIKE = 'cargo_bike';
 
+    const COLORS_LIST = ['#213ab2', '#b2213a', '#5221b2', '#93c63f', '#b22182', '#3ab221', '#b25221', '#2182b2', '#3ab221', '#9c21b2', '#c63f4f', '#b2217f', '#82b221', '#5421b2', '#3f93c6', '#21b252', '#c6733f'];
```
alexsegura: This is too much coupling between the view & the model. 
This should be done in Twig / JavaScript side.

src/AppBundle/Resources/config/doctrine/TaskCollectionItem.orm.yml
```diff
@@ -27,7 +27,7 @@ AppBundle\Entity\TaskCollectionItem:
         task:
             targetEntity: AppBundle\Entity\Task
             cascade:
-                - persist
+                - all
```
alexsegura: This doesn't seem related to this PR.

alexsegura:
The UI is cool ⭐️

src/AppBundle/Entity/Delivery.php
Atala: Advantage to having it here is that it is easy to serialize `color` in the API (so we have exactly same colors everywhere).

Maybe having a service that get the color is a better solution?

PR #382, opened 2018-06-15
Introduce additional product options

### LINKED ISSUE #377: Introduce "optional" product options
Right now, we only manage mandatory options, i.e which appear as **radio buttons**. 

However, some options may not be mandatory, for example "add a slice of bacon", "add cheese", "add sauce"... 
Such options should appear as **checkboxes**. 

To manage this, we need to modify the `ProductOption` class to know how the group should behave. 
A few possible implementations: 
- Add a boolean property `isMultiple` or `isOptional` (quite redundant)
- Add a range `minChoices` / `maxChoices`: if `minChoices = maxChoices = 1`, then this is a radio buttons
- Mimic how this is done in Symfony [`ChoiceType`](https://symfony.com/doc/3.4/reference/forms/types/choice.html#select-tag-checkboxes-or-radio-buttons)

⚠️ The "optional options" **should not belong to the variant**. 
See `createProductVariant` in [`RestaurantController`](https://github.com/coopcycle/coopcycle-web/blob/f5b02c8eb2382aeae6a371c0d21ab2b6ec66b012/src/AppBundle/Controller/RestaurantController.php#L89)

alexsegura:
Fixes #377 

capture d ecran 2018-06-15 a 15 22 27

### DIFF
--- a/src/AppBundle/Controller/RestaurantController.php
+++ b/src/AppBundle/Controller/RestaurantController.php
@@ -344,8 +344,18 @@ class RestaurantController extends Controller
 
             $optionValues = [];
             foreach ($options as $optionCode => $optionValueCode) {
-                $optionValue = $productOptionValueRepository->findOneByCode($optionValueCode);
-                $optionValues[] = $optionValue;
+
+                $optionValueCodes = [];
+                if (is_array($optionValueCode)) {
+                    $optionValueCodes = $optionValueCode;
+                } else {
+                    $optionValueCodes[] = $optionValueCode;
+                }
+
+                foreach ($optionValueCodes as $optionValueCode) {
+                    $optionValue = $productOptionValueRepository->findOneByCode($optionValueCode);
+                    $optionValues[] = $optionValue;
+                }
             }
 
             $nonExistingOption = $this->matchNonExistingOption($product, $optionValues);
--- a/src/AppBundle/Resources/translations/messages.fr.yml
+++ b/src/AppBundle/Resources/translations/messages.fr.yml
@@ -484,6 +484,8 @@ form.product_option.add_value.label: Ajouter une valeur
 form.product_option.values.title: Valeurs
 form.product_option_value.value.label: Valeur
 form.product_option_value.price.label: Prix
+form.product_option.additional.label: Cette option est un supplément
+form.product_option.additional.help: Le client pourra choisir plusieurs valeurs, sans être obligé d'en choisir au moins une
 product_option.strategy.free: Gratuit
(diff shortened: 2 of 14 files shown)


src/AppBundle/Resources/translations/messages.fr.yml
```diff
@@ -484,6 +484,8 @@ form.product_option.add_value.label: Ajouter une valeur
 form.product_option.values.title: Valeurs
 form.product_option_value.value.label: Valeur
 form.product_option_value.price.label: Prix
+form.product_option.additional.label: Cette option est un supplément
```
Atala: I don't think this text is clear enough.
"Cette option est facultative."

src/AppBundle/Resources/views/Restaurant/index.html.twig
```diff
@@ -126,20 +126,34 @@
                   {{ option.name }}
                   {% if option.strategy == 'option' %} - {{ option.price|price_format }}€{% endif %}
                 </h4>
-                <div class="list-group" data-product-option>
+                <div class="list-group" {% if not option.additional %}data-product-option{% endif %}>
```
Atala: I think you should add a text on the modal intended to the customer. Like "optional" next to the modifier's name.

src/AppBundle/Controller/RestaurantController.php
```diff
@@ -344,8 +344,18 @@ public function addProductToCartAction($id, $code, Request $request)
 
             $optionValues = [];
             foreach ($options as $optionCode => $optionValueCode) {
-                $optionValue = $productOptionValueRepository->findOneByCode($optionValueCode);
-                $optionValues[] = $optionValue;
+
+                $optionValueCodes = [];
+                if (!is_array($optionValueCode)) {
```
GaspardPO: Inverting the if / else will remove the "!" in the condition.

src/AppBundle/Resources/translations/messages.fr.yml
alexsegura: The problem is it translates to "this option is optional" 😅

src/AppBundle/Resources/views/Restaurant/index.html.twig
alexsegura: I think the option name can be used in this case, for example "Would you like an extra?"

alexsegura:
@Atala nice catch, the code was not working when there were _only_ additional options. This is fixed. 
Maybe some day we should wrap all of this in a React component, but I merge this for now.

PR #394, opened 2018-06-28
Add setting to switch stripe payment types

Atala:


package-lock.json
```diff
@@ -2724,7 +2724,7 @@
       "requires": {
```
alexsegura: `package-lock.json` shouldn't be modified in this PR

src/AppBundle/Service/OrderManager.php
```diff
@@ -186,20 +190,19 @@ public function capturePayment(OrderInterface $order)
...
 
         try {
 
             $charge = Stripe\Charge::retrieve(
                 $stripePayment->getCharge(),
-                array(
-                    'stripe_account' => $restaurantStripeId
-            ));
+                $stripeOptions
+            );
```
alexsegura: I think there is something missing after that: we still need to create a Transfer to the Stripe account of the restaurant. See faf97fa5d5a52ba9c9188f9ea526e3c2b58e43ea

ccoreilly: There is no need to create a transfer when using destination charges. Before https://github.com/coopcycle/coopcycle-web/commit/faf97fa5d5a52ba9c9188f9ea526e3c2b58e43ea we were using "separate charges and transfers": see https://stripe.com/docs/connect/charges

package-lock.json
ccoreilly: Removed

src/AppBundle/Resources/views/Restaurant/form.html.twig
```diff
@@ -164,10 +164,11 @@
   {% endif %}
 
   {% if form.contract is defined %}
-    {{ form_row(form.contract.minimumCartAmount) }}
+    {{ form_row(form.contract.minimumCartAmount, { help: 'restaurant.contract.flatDeliveryPrice.help' }) }}
```
Atala: I think there is a typo here, help text shouldn't have been added.

app/DoctrineMigrations/Version20180627124854.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180627124854 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+        $this->addSql('ALTER TABLE contract ADD restaurant_pays_stripe BOOLEAN DEFAULT \'true\' NOT NULL');
```
Atala: Cool to have default set to `true` for existing contracts

src/AppBundle/Entity/Contract.php
```diff
@@ -46,6 +46,13 @@ class Contract
      */
     private $feeRate;
 
+    /**
+     * @var bool
+     * Restaurant pays Stripe fee?
+     * @Assert\Type("bool")
+     */
+    private $restaurantPaysStripeFee = true;
```
Atala: Not a huge fan to have this field directly linked to Stripe use..

src/AppBundle/Resources/views/Restaurant/form.html.twig
```diff
@@ -164,10 +164,11 @@
...
     {{ form_row(form.contract.flatDeliveryPrice, { help: 'restaurant.contract.flatDeliveryPrice.help' }) }}
     {{ form_row(form.contract.customerAmount, { help: 'restaurant.contract.customerAmount.help' }) }}
     {{ form_row(form.contract.feeRate, { help: 'restaurant.contract.feeRate.help' }) }}
+    {{ form_row(form.contract.restaurantPaysStripeFee) }}
```
Atala: Can you add an help text as well ?
Also as you suggested can you hide this field by default ?

src/AppBundle/Service/OrderManager.php
```diff
@@ -186,20 +190,19 @@ public function capturePayment(OrderInterface $order)
...
         }
 
         try {
 
             $charge = Stripe\Charge::retrieve(
                 $stripePayment->getCharge(),
-                array(
-                    'stripe_account' => $restaurantStripeId
-            ));
+                $stripeOptions
```
Atala: Can you remove this extra-param (and code related to it), as we don't need it to retrieve the charge ?

ccoreilly: I removed the param and tried it out but the charge is not retrieved correctly because it is not associated with the coop's Stripe ID. We have to pass the restaurant's Stripe ID when using Direct Charges.

Atala: I gave it more thought, and it actually makes perfect sense. When the restaurant pays the fee, the created Stripe charge is not linked to our account at all (actually you don't see anything from the platform's stripe dashboard before capture and seeing the application fee then). you need the extra-parameter to "connect as" the restaurant owner.

ccoreilly: Yes, exactly.

src/AppBundle/Service/OrderManager.php
```diff
@@ -142,7 +142,10 @@ public function authorizePayment(OrderInterface $order)
         $stripeOptions = array();
 
         if (!is_null($stripeAccount)) {
+
             if($restaurantPaysStripeFee) {
+                // needed only when using direct charges (the charge is linked to the restaurant's Stripe account)
+                $stripePayment->setStripeUserId($stripeAccount->getStripeUserId());
```
ccoreilly: This solves https://github.com/coopcycle/coopcycle-web/issues/398 perfectly, well thought!

PR #397, opened 2018-06-30
Drop support of predefined locations in AddressPicker

alexsegura:
This Pull Request drops support of "predefined locations" in AddressPicker. 
We now use the base library instead of a fork.


alexsegura:
Well, actually I was too confident 😞

The component heavily relies on its internal state, so it seems impossible to correctly add the desired functionality. 

I think I'm going to create an address picker from the ground up using `react-autosuggest`

PR #405, opened 2018-07-21
Add endpoint to create deliveries.

### LINKED ISSUE #374: API endpoint to create deliveries for shop orders
This is a proposal & an RFC to reintroduce an API endpoint to create deliveries. 
There used to be such an endpoint, but right now deliveries are [read-only](https://github.com/coopcycle/coopcycle-web/blob/823643d2d1a561be898350faf2e1ff4a340272dd/src/AppBundle/Entity/Delivery.php#L17)

The goal is to allow external systems (i.e, an e-commerce website) to create deliveries
This way, merchants having an existing e-commerce website could delegate the shipping to a coop using CoopCycle 
As an example, there is some draft work in [coopcycle-wp](https://github.com/coopcycle/coopcycle-wp) to add a CoopCycle Shipping Method in WooCommerce

### General idea

A `Store` in CoopCycle represents the external store. 
It provides a user account to the store owner, and allows grabbing an API Token. 
When an API call is received with a Store token, it creates a `Delivery` with a pickup address corresponding to the store address defined in the back-office. 

### Authentication

I'm not sure we can use JWT authentication as-is for those tokens, as they need to have a long TTL. 
Maybe we will need to use a custom token system & authenticator in Symfony Security?

### Keep track of the external order

I think we will need to keep track of the order on the external system. 
Maybe we could create a "fake" order in our system too?

### JSON payload

Here is an example of JSON payload 

```
POST /api/deliveries
{
  pickup: {
    before: "2018-06-08 12:00:00"
  }, 
  dropoff:  {
    address: {
      streetAddress: "67, rue de la Sorbonne", 
      firstName: "Jane",
      lastName: "Doe", 
      telephone: "+33612345678"
    },
    before: "2018-06-08 14:00:00"
  }, 
  order: {
    code: "00000001"
  }
}
```

alexsegura:
Fixes #374 

This is the first draft to add an endpoint to create deliveries. 
As explained in the original issue, each store has a token associated, which allows to identify the originator. (Please note this Pull Request does not take care of creating or displaying store tokens). 

The payload is really simple: 

```
{
  "pickup": {
     // Optional, if not pickup address is provided, use store address
    "address": "24, Rue de la Paix",
    "doneBefore": "tomorrow 13:00"
  },
  "dropoff": {
    "address": "48, Rue de Rivoli",
    "doneBefore": "tomorrow 13:30"
  }
}
```

What do you think? This is just a starting point, we can iterate from this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ install:
 	@printf "\e[0;32mPopulating schema..\e[0m\n"
 	@docker-compose run php bin/console doctrine:schema:create --env=dev
 	@docker-compose run php bin/demo --env=dev
-	@docker-compose run php bin/console doctrine:migrations:version --add --all
+	@docker-compose run php bin/console doctrine:migrations:version --no-interaction --quiet --add --all
 
 osrm:
 	@mkdir -p var/osrm
--- a/app/DoctrineMigrations/Version20180802111729.php
+++ b/app/DoctrineMigrations/Version20180802111729.php
@@ -1,0 +1,31 @@
+<?php declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20180802111729 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE TABLE store_token (id SERIAL NOT NULL, store_id INT DEFAULT NULL, token TEXT NOT NULL, created_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, updated_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, PRIMARY KEY(id))');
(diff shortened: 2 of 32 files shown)


Atala:
I think we need at least a second endpoint that returns the delivery's price (to show it to the customer for example).

I am not sure a store should be able to set an arbitrary pickup address, it should be the store's address -> what do you think?

And then we'll be good to go.

ccoreilly:
@Atala The reason Alex allowed an arbitrary pickup address is to allow platforms that promote P2P deliveries. At La Pájara we are wanting to start working with such a platform that allows people to sell cooked food to other people, in this case their platform would create a delivery using the endpoint developed here setting the pickup address to the "cook's" address.

Do we need a Store token or could we user the user JWT used in the app for authentication? If the authenticated user is a store owner then a delivery can be created.

alexsegura:
Yes exactly. This is a general purpose endpoint. 

Regarding authentication, yes, we can use JWTs

However, we need to change slightly the mechanism: i.e, we need to store them in database, in order to make them revokable. Actually, JWTs are not really revokable, they just expire. 

In this case, I think we need to have JWTs that don't expire, while we need to be able to revoke them, because if a token which allows creating deliveries is compromised, it could be used to flood the platform with fake data. 

I think omitting the [`exp` claim](https://tools.ietf.org/html/rfc7519#section-4.1) is sufficient to have JWTs that don't expire. Actually, we can encode in the JWT the intended behavior, i.e. we can say "this JWT is allowed to set arbitrary pickup addresses". 

Ok, I will change to JWTs 👍

alexsegura:
More about revoking tokens:
- https://github.com/lexik/LexikJWTAuthenticationBundle/issues/58#issuecomment-89641970
- https://stackoverflow.com/questions/21978658/invalidating-json-web-tokens#23089839

PR #409, opened 2018-08-07
Introduce users search with ElasticSearch

alexsegura:
For another branch, I need a quick & reliable way of searching users easily (i.e., by typing the first letters of their firstname, lastname, email... or whatever). 
Using `LIKE` queries is not reliable, and does not handle typos. 

This is why I finally decided to introduce ElasticSearch 🚀
I think it will have a small memory footprint, with a few thousands of documents. 

So let's begin with users search, so that we can learn how ElasticSearch works (I don't know much of it myself, it is quite complex). 

In production, we can run it easily using Docker, with only one node. 
When we have more money and experience, we can create a real cluster with several nodes and lots of RAM 😁

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ services:
   - postgresql
   - docker
   - redis-server
+  - elasticsearch
 
 addons:
   postgresql: '9.4'
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -1,6 +1,9 @@
 imports:
     - { resource: config_dev.yml }
 
+parameters:
+    elasticsearch_prefix: '%database_name%_test'
+
 framework:
     test: ~
     session:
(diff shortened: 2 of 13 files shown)


Atala:
For which branch do you need fuzzy text search?

I disagree with introducing Elastic Search in our stack as it is yet another component and hard to manage (for example its dumps on disk are huge).

I removed it from the stack in my previous job far a text search based on Postgres functionalities. I used [this way](https://gearheart.io/blog/how-to-use-django-postgresql-for-full-text-search/) but [this way](https://www.rdegges.com/2013/easy-fuzzy-text-searching-with-postgresql/) seems also promising

anayrat:
Loog at pg_trgm : https://www.postgresql.org/docs/current/static/pgtrgm.html

alexsegura:
Fixed in 05deb7b1d02f5929d1c0b288944d178d40308a22

PR #417, opened 2018-08-25
Introduce order timeline, allow delaying orders. 

alexsegura:
This Pull Request introduces an "Order Timeline", i.e a database table containing probabilistic times for each step of the workflow. It calculates the time the restaurant should start preparing the meal based on the order total. 

This will allow triggering alerts, and opens the way to detect orders which will probably be late, or need to be accepted shortly. 

Also, it adds a new button allowing to delay an order (hardcoded to 10 minutes).


js/app/restaurant/OrderDetails.jsx
```diff
@@ -155,7 +162,7 @@ class OrderList extends React.Component {
           <p>
             <span className="text-left"><OrderLabel order={ order } /></span>
             <strong className="pull-right text-success">
-                { moment(order.shippedAt).format('lll') }  <i className="fa fa-clock-o" aria-hidden="true"></i>
+                { moment(order.preparationExpectedAt).format('LT') }  <i className="fa fa-clock-o" aria-hidden="true"></i>
```
Atala: Maybe add a small text on the left saying what this time means

PR #445, opened 2018-09-22
Allergens & restricted diets

alexsegura:
This introduces allergens & restricted diets management for meals. 

⚠️ Please note this is just an introduction, to be able to link meals with allergens, and for the customer to see meals containing allergens. 
Features such as filtering will be discusses & managed later, if needed. 

### Backend

On the product form, admins can check allergens & restricted diets for each product. 

capture d ecran 2018-09-22 a 13 44 35

### Frontend

Allergens are displayed with a warning icon. 
Restricted diets are displayed with a checkmark icon.

capture d ecran 2018-09-22 a 13 20 20

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -379,7 +379,13 @@ services:
     tags: [ form.type ]
 
   'AppBundle\Form\ProductType':
-    arguments: [ '@sylius.factory.product_variant', '@sylius.product_variant_resolver.default' ]
+    arguments:
+      - '@sylius.factory.product_variant'
+      - '@sylius.product_variant_resolver.default'
+      - '@sylius.repository.product_attribute'
+      - '@sylius.factory.product_attribute_value'
+      - '@sylius.locale_provider'
+      - '@translator'
     tags: [ form.type ]
 
   my.jwt_listener:
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -401,7 +401,9 @@ a.list-group-item.menu-item--unavailable {
   width: 80%;
 }
 
-.menu-item .menu-item-description {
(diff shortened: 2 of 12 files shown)


alexsegura:
@AmelieA this is the list of 14 allergens that must always be labelled in pre-packed and non-prepacked foods in the EU. They cover more than 80% of all the food allergies. 
So, this way, we are compliant with the legislation. 

Having a complete taxonomy would require more effort for administrators. 
But this can be changed later, for example to help administrators detect food allergies with a complete list of ingredients, for example. 

> maybe a "I garanti this product does not contain any allergène" box to be sure people don't skip this step by mistake ?

This is a good idea. I prefer to display a warning instead, to trigger attention on the list.

AmelieA:
Seems like I forgot to link the document : https://docs.google.com/document/d/1TTIL2j_SBf3jD9sp7oO1VG1Yt3cw9ZxuCjGzpq38z0w/edit
The problème with the covering only 80% from the start is, when we do covers 99%, you won't be sure that every restaurant did the change and added the allergen. And that would be playing with people lives.
Also, if you enter, red meat as an allergen, and alcool, like on the document, your life will become way easier when you have to include diets.

For the warning display, people tend to not read them (specially for repetitive task). But it would not be our responsibility anymore...

alexsegura:
@AmelieA I understand your concerns, but let's iterate with small steps. 
As the filtering issue, the admin UX is a whole subject by itself. We could imagine a more comprehensive "wizard" when creating a product, as the page is beginning to be quite bloated, but we'll do this later during a general redesign of the admin UI.

AmelieA:
The allergens are missing from the options. Like if your option is eatable, you also need it there

Issue #391, opened 2018-06-19
Selenium testing


alexsegura:
It should be great to setup Selenium testing, locally & on Travis using Docker

- [Consistent Selenium testing with Docker](https://medium.com/@garrensmith/consistent-selenium-testing-with-docker-f2d5a24a1bc5)
- [How to integrate Travis with Selenium Tests using Docker](http://witkowskibartosz.com/blog/how_to_integrate_travis_with_selenium_tests_using_docker.html)

alexsegura:
Also see Panther 

https://symfony.com/blog/introducing-symfony-panther-a-browser-testing-and-web-scrapping-library-for-php

PR #482, opened 2018-10-15
Add selenium & tests

freelibre:
This is a Work in progress for issue:
https://github.com/coopcycle/coopcycle-web/issues/391

With this you can log in a Selenium grid console with 2 browers (chrome & firefox):
http://localhost:4444/grid/console
An exemple test case is added in side format (exported from firefox selenium IDE)
A new npm package is added to launch side test from CLI

Next steps proposal:
- try to launch side-runner tests automatically and get result in grid thru VNC
- try to add it in Travis 
- try to add existing behat test


alexsegura:
Thank you for this 👏

IMHO, the final goal is to have browser tests written in Behat/Gherkin language. 
I think this is more maintainable, and we can have access to the database, to create fixtures, for example. 

Selenium IDE can be cool, so that "random" users can help us write tests. 
I found [salmabenhadjali/gherkinator](https://github.com/salmabenhadjali/gherkinator) that allows to convert `*.ide` files to Gherkin language. 
A pretty nice workflow would be: 
- People create `*.ide` files without any knowledge of Gherkin
- We convert them to Gherkin language automatically

Actually, I would like to give a try to [symfony/panther](https://github.com/symfony/panther), it looks like it is way easier to use than full-stack Selenium. 
Let me try to setup `symfony/panther` on the project. It looks like it is not supported out of the box in Behat (see Behat/Symfony2Extension#145), but it should be easy.

freelibre:
Thank you for your feeback. I think `*.ide` file will be deprecated soon so I've proposed `*.side` files generated by the new IDE version. It should be possible to then use [selenium-side-runner](https://www.npmjs.com/package/selenium-side-runner) to run tests. You are right such files generated are indeed more for people willing to contribute by adding some End 2 End scenario without lot of coding knowledge as it just needs the browser extension you can grab [firefox extension](https://addons.mozilla.org/fr/firefox/addon/selenium-ide/) for example and such tests could be generated and updated very quickly. I'm trying  behat/mink facebook\php webdriver to interface with selenium hub but your solution seems much wiser.

alexsegura:
While digging around I came across [Zalenium](https://opensource.zalando.com/zalenium/)

- [zalando/zalenium](https://github.com/zalando/zalenium)
- [docker-selenium](https://github.com/elgalu/docker-selenium)

freelibre:
Yes you are right Zalenium add more features. If you want I will change to Zalenium. I've just push the VNC support to current selenium install so now you can connect to a node `URL:5900` with a VNC client like VNCtiger or Remmina (the login password is : secret) and then you will be able to see the tests playing in a browser.

alexsegura:
See #502

PR #388, opened 2018-06-19
WIP: Windows compliance

4x10m:
Related issues:
https://github.com/4x10m/coopcycle-web/pull/1
https://github.com/4x10m/coopcycle-web/pull/2
https://github.com/4x10m/coopcycle-web/pull/3

* [ ] Deploying on windows using `make install` working
* [ ] Deploy on windows using `powershell make.ps1` working
* [ ] Deploying on linux always working

For continuation of this work: see #392 

WIP
Don't merge plz

PR #508, opened 2018-11-30
Make Docker work on Windows

alexsegura:
Supersedes #388

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,10 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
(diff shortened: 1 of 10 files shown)


alexsegura:
Thank you @stfbohn, @4x10m & @fedegos  🎉

A few things:
- I removed `make.ps1`, because it seemed useless
- I reverted the changes on `app/config/app_dev.php` for `assets_base_urls`, we need to find a way to make it dynamic (should be possible to guess the Docker host via an environment variable)
- I reverted the changes in `Makefile` commenting out the OSRM build, we need to find a workaround for Windows

Issue #435, opened 2018-09-13
Missing internationalisation on new restaurant page


eliascodes:
Hey guys, 

Looks like the hours section on the new restaurants page has a couple of missing translations:

<img width="284" alt="screen shot 2018-09-12 at 17 35 50" src="https://user-images.githubusercontent.com/12719553/45490024-f1b7db80-b75d-11e8-8a3d-e6c619cfec00.png">

PR #510, opened 2018-12-03
Fix issue #435 Translate Open hours "table title" and "Add" buton

freelibre:
Fix issue #435: "Missing internationalisation on new restaurant page"


alexsegura:
Thank you!

I had to fix the Zalenium version, see zalando/zalenium#783

PR #572, opened 2019-06-18
Fix issue571

arnomasson:


alexsegura:
Thank you! 👏

I cherry-picked 60c314158efb4c2cca53cabc0763bd931ccb863c into `master`, because this Pull Request contains unnecessary changes on `package-lock.json`

I opened #573 to check if we can find a way to fix this problem.

PR #577, opened 2019-06-21
Resend email

### LINKED ISSUE #576: Add button to send registration email again
Sometimes, the email to confirm registration is not sent. 
After registration, on the `check_email.html.twig` template, we should show a button to send email again.

arnomasson:
Fixes #576

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -65,7 +65,7 @@ services:
                   - '@apns_php.logger'
 
 # Uncomment to send emails in dev
-#fos_user:
+# fos_user:
 #   registration:
 #       confirmation:
 #           enabled: true
--- a/src/AppBundle/Controller/UserController.php
+++ b/src/AppBundle/Controller/UserController.php
@@ -4,12 +4,15 @@ namespace AppBundle\Controller;
 
 use AppBundle\Entity\ApiUser;
 use FOS\UserBundle\Model\UserManagerInterface;
+use FOS\UserBundle\Mailer\MailerInterface;
 use Laravolt\Avatar\Avatar;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+
 
 class UserController extends AbstractController
 {
@@ -72,5 +75,22 @@ class UserController extends AbstractController
         $response->headers->set('Content-Type', 'image/png');
 
         return $response;
+    }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer, SessionInterface $session)
+    {
+        if ($request->request->has('email')) {
+            $email = $request->request->get('email');
+            $user = $userManager->findUserByEmail($email);
+            if ($user !== null) {
+                $mailer->sendConfirmationEmailMessage($user);
+                $session->set('fos_user_send_confirmation_email/email', $email);
+                return $this->redirectToRoute('fos_user_registration_check_email');
+            } 
+        }
+        return $this->redirectToRoute('fos_user_security_login');   
     }
 }
--- a/src/AppBundle/Resources/translations/messages.en.yml
+++ b/src/AppBundle/Resources/translations/messages.en.yml
@@ -141,7 +141,6 @@ localBusiness.form.telephone: Telephone
 localBusiness.form.openingHours: Opening hours
 localBusiness.form.deliveryPerimeterExpression: Zone or max distance for delivery
 
-
 restaurant.available: Available
 restaurant.unavailable: Unavailable
 restaurant.disabledWarning: |
@@ -168,6 +167,7 @@ restaurant.contract.flatDeliveryPrice.help: This amount will not be displayed in
 restaurant.contract.customerAmount.label: Amount paid by the customer
 restaurant.contract.customerAmount.help: This amount will be displayed in the cart and will increase the total amount
 restaurant.contract.feeRate.label: Fee rate
+
 restaurant.contract.feeRate.help: This percentage will be retained by the platform on the total of the products
 restaurant.contract.restaurantPaysStripeFee.label: Responsible for the cost of the Stripe fees, refunds, and chargebacks
 restaurant.contract.restaurantPaysStripeFee.help: Choose who will be responsible for the cost of the Stripe fees and managing refunds and chargebacks
@@ -549,6 +549,8 @@ form.credit_note.amount.label: Amount
 form.credit_note.name.label: Name
 form.credit_note.name.help: The name of the promotion that will be displayed in the cart
 
+form.resend_confirmation_email.submit.label: Send email again
+
 form.embed_settings.with_vehicle.label: Add vehicle choice
 
 promotion_coupon.used.label: Number of uses
@@ -712,6 +714,8 @@ restricted_diets.LOW_LACTOSE_DIET: Low lactose
 restricted_diets.LOW_SALT_DIET: Low salt
 restricted_diets.VEGAN_DIET: Vegan
 restricted_diets.VEGETARIAN_DIET: Vegetarian
+
+resend_confirmation_email.disclaimer: If you have not received the confirmation email, click on the button.
 
 sylius.ui.code: Code
 sylius.form.promotion_coupon.per_customer_usage_limit: Per customer usage limit
--- a/src/AppBundle/Resources/translations/messages.fr.yml
+++ b/src/AppBundle/Resources/translations/messages.fr.yml
@@ -570,6 +570,8 @@ form.credit_note.name.help: Le nom de la réduction qui sera affiché dans le pa
 
 form.embed_settings.with_vehicle.label: Ajouter le choix du type de véhicule
 
+form.resend_confirmation_email.submit.label: Renvoyer l'email
+
 promotion_coupon.used.label: Nombre d'utilisations
 
 product_option.strategy.free: Gratuit
@@ -772,6 +774,8 @@ restricted_diets.LOW_LACTOSE_DIET: Sans lactose
 restricted_diets.LOW_SALT_DIET: Sans sel
 restricted_diets.VEGAN_DIET: Vegan
 restricted_diets.VEGETARIAN_DIET: Végétarien
+
+resend_confirmation_email.disclaimer: Vous n'avez pas l'email de confirmation ? Cliquez sur le bouton ci-dessous.
 
 sylius.ui.code: Code
 sylius.form.promotion_coupon.per_customer_usage_limit: Limite d'utilisation par client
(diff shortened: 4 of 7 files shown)


src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,19 @@ public function avatarAction($username, Request $request)
 
         return $response;
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email")
```
alexsegura: The route should be limited to `POST`
See https://symfony.com/doc/current/routing/requirements.html#adding-http-method-requirements

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,19 @@ public function avatarAction($username, Request $request)
...
+     * @Template()
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer)
+    {
+        $email = $request->request->get("email");
+        $user = $userManager->findUserByEmail($email);
```
alexsegura: The variable `$email` may contain the email of a non-existing user.
You should make sure that `$user` is not `NULL`.

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,21 @@ public function avatarAction($username, Request $request)
...
         return $response;
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     * @Template()
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer)
+    {
+        if($request->request->get("email") !== NULL) {
```
alexsegura: There should be a space after `if`

alexsegura: You should assign `$request->request->get("email")` into a variable (like below), and use this variable instead of calling the function twice

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,21 @@ public function avatarAction($username, Request $request)
...
+            $email = $request->request->get("email");
+            $user = $userManager->findUserByEmail($email);
```
alexsegura: Problem still there: you are not checking if `$user` exists. `$user` may be NULL if the email does not exist in database.

src/AppBundle/Controller/UserController.php
alexsegura: To check if the request contains a parameter, you can do `$request->request->has('email')`: it returns a boolean if the parameter exists or not.

alexsegura: Please use simple quotes instead of double quotes

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,21 @@ public function avatarAction($username, Request $request)
...
+            $mailer->sendConfirmationEmailMessage($user);
+        }
+
+        return $this->redirectToRoute('fos_user_registration_check_email');
```
alexsegura: It works because you have modified locally `RegistrationController` 🙂 

https://github.com/FriendsOfSymfony/FOSUserBundle/blob/5d2145777fa1d61ba27ba4c1e7016f3c3e7ee7ba/Controller/RegistrationController.php#L106-L124

Before redirecting, you need to set something in session, like in `EmailConfirmationListener`

https://github.com/FriendsOfSymfony/FOSUserBundle/blob/5d2145777fa1d61ba27ba4c1e7016f3c3e7ee7ba/EventListener/EmailConfirmationListener.php#L59-L75

app/config/config_dev.yml
```diff
@@ -65,7 +65,7 @@ services:
                   - '@apns_php.logger'
 
 # Uncomment to send emails in dev
-#fos_user:
-#   registration:
-#       confirmation:
-#           enabled: true
+fos_user:
```
alexsegura: You shouldn't have committed this change

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,24 @@ public function avatarAction($username, Request $request)
...
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     * @Template()
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer)
+    {
+        $emailExists = $request->request->has('email');
+        $userExists = $request->request->has('user');
```
alexsegura: There is no parameter called `user`

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,24 @@ public function avatarAction($username, Request $request)
...
+        if ($emailExists && $userExists == true) {
```
alexsegura: Why difference of style in the same instruction (`$var` VS `$var == true`)?

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,24 @@ public function avatarAction($username, Request $request)
...
+            $email = $request->request->get('email');
+            $user = $userManager->findUserByEmail($email);
+            $mailer->sendConfirmationEmailMessage($user);
+        }
+
+        $this->session->set('fos_user_send_confirmation_email/email', $user->getEmail());
```
alexsegura: So close... 😬
`$this->session` is `NULL` here. Check how to get the session in the [docs](https://symfony.com/doc/current/controller.html#managing-the-session)

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +74,24 @@ public function avatarAction($username, Request $request)
 
         return $response;
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     * @Template()
```
alexsegura: The `@Template` annotation is not necessary

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +76,25 @@ public function avatarAction($username, Request $request)
...
         return $response;
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer, SessionInterface $session)
+    {
+        $emailExists = $request->request->has('email');
+        if($emailExists == true) {
```
alexsegura: - There should be a space after `if`
- You should use strict equality (`===`)

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +76,25 @@ public function avatarAction($username, Request $request)
...
+            $email = $request->request->get('email');
+            $user = $userManager->findUserByEmail($email);
+        }   
+        if ($user !== null) {
```
alexsegura: This `if` block should be nested inside the previous `if` block.

src/AppBundle/Resources/translations/messages.en.yml
```diff
@@ -168,6 +168,27 @@ restaurant.contract.flatDeliveryPrice.help: This amount will not be displayed in
 restaurant.contract.customerAmount.label: Amount paid by the customer
 restaurant.contract.customerAmount.help: This amount will be displayed in the cart and will increase the total amount
 restaurant.contract.feeRate.label: Fee rate
+
```
alexsegura: Unnecessary spaces

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +76,23 @@ public function avatarAction($username, Request $request)
...
         return $response;
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer, SessionInterface $session)
+    {
+        $emailExists = $request->request->has('email');
+        if ($emailExists === true) {
```
alexsegura: And what if the `email` parameter does not exist?

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +76,27 @@ public function avatarAction($username, Request $request)
...
         return $response;
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer, SessionInterface $session)
+    {
+        $urlCheckEmail = $this->generateUrl('fos_user_registration_check_email');
+        $urlLogin = $this->generateUrl('fos_user_security_login');
```
alexsegura: `$urlCheckEmail` & `$urlLogin` variables are not used, remove them.

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +76,27 @@ public function avatarAction($username, Request $request)
...
+            $email = $request->request->get('email');
+            $user = $userManager->findUserByEmail($email);
+            if ($user !== null) {
+                $mailer->sendConfirmationEmailMessage($user);
+                $session->set('fos_user_send_confirmation_email/email', $email);
+                return $this->redirectToRoute('fos_user_registration_check_email');
+            } else {
+                return $this->redirectToRoute('fos_user_security_login');
+            }
+        } else {
```
alexsegura: There is no need to put an `else` here, you can put the `return` outside of the `if`

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +76,27 @@ public function avatarAction($username, Request $request)
...
+        $urlLogin = $this->generateUrl('fos_user_security_login');
+        $emailExists = $request->request->has('email');
+        if ($emailExists === true) {
+            $email = $request->request->get('email');
+            $user = $userManager->findUserByEmail($email);
+            if ($user !== null) {
+                $mailer->sendConfirmationEmailMessage($user);
+                $session->set('fos_user_send_confirmation_email/email', $email);
+                return $this->redirectToRoute('fos_user_registration_check_email');
+            } else {
```
alexsegura: This `else` seems useless

src/AppBundle/Resources/translations/messages.fr.yml
```diff
@@ -773,5 +775,7 @@ restricted_diets.LOW_SALT_DIET: Sans sel
 restricted_diets.VEGAN_DIET: Vegan
 restricted_diets.VEGETARIAN_DIET: Végétarien
 
+resend_confirmation_email.disclaimer: Si vous n'avez pas reçu l'email de confirmation, cliquez sur le boutton.
```
alexsegura: `boutton` → `bouton`

I think the wording would be nicer with something like: 
```
Vous n'avez pas l'email de confirmation ? Cliquez sur le bouton ci-dessous.
```

src/AppBundle/Controller/UserController.php
```diff
@@ -73,4 +76,22 @@ public function avatarAction($username, Request $request)
...
         return $response;
     }
+
+    /**
+     * @Route("/register/resend-email", name="register_resend_email", methods={"POST"})
+     */
+    public function resendRegistrationEmailAction(Request $request, UserManagerInterface $userManager, MailerInterface $mailer, SessionInterface $session)
+    {
+        $emailExists = $request->request->has('email');
+        if ($emailExists === true) {
```
alexsegura: You can do directly `if ($request->request->has('email'))`, the `$emailExists` is not used anywhere else.

PR #582, opened 2019-06-26
fix too long list orders in panels

### LINKED ISSUE #386: Cart with long list of items
When there is a very long list of items / adjustments, the cart element stays fixed but does not push the page content. 

rayon_9_cart

### LINKED ISSUE #546: big command
When the command is too expensive, a problem appear for restaurants 
Capture
They can't see everything

arnomasson:
Fixes #386
Fixes #546

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -214,7 +214,7 @@ h1, h2, h3, h4, h5, h6 {
 // ------------------ Cart ------------------
 
 #cart {
-  margin-top: $base-margin;
+  margin-top: ($base-margin + 20px);
 }
 
 // $panel-heading-padding is 10px 15px !default;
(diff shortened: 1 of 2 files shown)


assets/css/styles.scss
```diff
@@ -305,6 +305,21 @@ $cart-heading-line-height: ($navbar-height - ($panel-heading-padding-vertical *
   }
 }
 
+/* fix too long list order*/
+.sticky-inner-wrapper {
+ overflow-y: auto; 
+ height: 85%; 
+}
+
+.panel.panel-default {
```
alexsegura: This selector targets too much elements! It should be more restrictive.
Moreover, I think it doesn't change anything.

PR #585, opened 2019-07-01
Restaurant "pledges"

### LINKED ISSUE #491: Allow users to show interest in unlisted restaurants
# Feature context
Despite all the good values from CoopCycle it can be hard for restaurants to make the switch (or just use it in addition to others providers).  
They can be skeptical, cautious or just waiting to see the development of the service in their town. Yet this is critical for cooperative to be able to generate sufficient revenues and survive, as the more restaurants they get the more work they get.
One way to encourage them is to show them that clients are already waiting for their restaurant to order food. A bit similar in the idea to what Liberapay [is doing](https://en.liberapay.com/explore/pledges)

## User Stories :
- Oliver, a client, is looking for it's favorite vegan restaurant in Lyon. He use the search box and type his home address but the restaurant doesn't seem to use this service yet he would like to let them know that he would be willing to order from them
- Camille, an associate of the cooperative, is looking to prospect for new restaurants for her users but doesn't know which one she should go first. She goes to the admin panel and look under the new section "Pledges" and sees a list of restaurants her users are interested in
- Camille, is prospecting a new restaurant that she previously spotted. The owner seems interested but hesitant because of the ratio hassle/gain. She tells him about the benefits and growth of her cooperative and to support her claim she shows him that already 20 clients are interested in his restaurant (optional: with an average of 200€ revenue per week)

## UI Stories
- At the bottom of the restaurants page and the search results page, add a link "Looking for a restaurant that is not listed ? Click here to let them know you are interested".
- From the previous link show a form with two sections, one for the restaurant and one for the pledge and a submit button "I'm interested". Restaurant section : Show a list of already submitted pledged restaurants to select from and a link "Not in this list ? Click here to add it" that make a form for the restaurant appear with two fields (name and address).
(optional) Pledge section : Show a form with the average amount of the client's cart and the frequency of his orders
- In the admin panel add a link to a new section called "Pledges", there show a list of the restaurants pledged by users orderable by the number of interested users (optional, by average revenue per week)

arnomasson:
Fixes #491


arnomasson:
Fixes #585

src/AppBundle/Controller/RestaurantController.php
```diff
@@ -602,8 +602,32 @@ public function mapAction(Request $request, SlugifyInterface $slugify)
...
+            $pledge->setCreatedAt(new \DateTime());
+            $pledge->setState(1);
+            $pledge->setUser($this->getUser());
+            $manager->persist($pledge);
+            $manager->flush();
+
+            return $this->redirectToRoute('restaurants_suggest_submitted');
+    }
+        return $this->render('@App/restaurant/restaurant_pledge.html.twig', [
+        'formPledge' => $form->createView()
```
alexsegura: In Twig, the coding style is to use snake case, so the variable should be `form_pledge`

app/DoctrineMigrations/Version20190705153145.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+final class Version20190705153145 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE TABLE restaurant_pledge_vote (id SERIAL NOT NULL, user_id INT NOT NULL, pledge_id INT DEFAULT NULL, voted_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, PRIMARY KEY(id))');
+        $this->addSql('CREATE UNIQUE INDEX UNIQ_9A6FB9B1A76ED395 ON restaurant_pledge_vote (user_id)');
```
alexsegura: Problem here is that the generated SQL contains `CREATE UNIQUE INDEX` on `user_id`.
It means users can vote once. We want them to vote once **per pledge**. 
Probably the mapping is wrong in `Restaurant.PledgeVote.orm.yml`. Maybe the association mapping between `PledgeVote` & `ApiUser` should be `ManyToMany`?

app/DoctrineMigrations/Version20190708092901.php
```diff
@@ -0,0 +1,38 @@
...
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20190708092901 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE restaurant ADD pledge_id INT DEFAULT NULL');
```
alexsegura: We shouldn't need to add a column to both tables. A `restaurant_id` column in the `restaurant_pledge` table should be enough. I think you need to add `mappedBy` & `inversedBy` in `Restaurant.Pledge.orm.yml` & `Restaurant.orm.yml`. Check [Doctrine docs](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/association-mapping.html#one-to-one-bidirectional), it only adds a column in one table.

src/AppBundle/Controller/RestaurantController.php
```diff
@@ -253,6 +257,16 @@ public function indexAction($id, $slug, Request $request, SlugifyInterface $slug
             }
         }
 
+        if ($restaurant->getState() === 'pledge') {
+
+            $numberOfVotes = count($restaurant->getPledge()->getVotes());
+
+            return $this->render('@App/admin/restaurant_pledge_accepted.html.twig', [
```
alexsegura: The path of this Twig template is wrong, it is not an admin area template. 
It should be `@App/restaurant/pledge.html.twig`

src/AppBundle/Resources/translations/emails.fr.yml
```diff
@@ -42,6 +42,25 @@ admin.order_foodtech.created.body: |
...
+  L'Equipe CoopCycle.
+user.pledge.subject: Votre suggestion est désormais sur CoopCycle !
+user.pledge.body: |
+  Bonjour,
+  <br>
+  grâce à vous, %pledge.name% est désormais sur CoopCycle !
+  <br>
+  Merci,
+  <br>
+  L'Equipe CoopCycle.user.pledge.subject: Votre suggestion est désormais sur CoopCycle !
```
alexsegura: Typo here, remove "user.pledge.subject: Votre suggestion est désormais sur CoopCycle !"

src/AppBundle/Resources/views/admin/restaurant_pledge_accepted.html.twig
```diff
@@ -0,0 +1,36 @@
...
+        <div class="row">
+        	<div class="col-xs-4 col-xs-push-4">
+                <h5>{{ restaurant.address.streetAddress }}</h5>
+                <h5>{{ restaurant.address.postalCode }}</h5>
+                <h5>{{ restaurant.address.addressLocality }}</h5>
+            </div>
+        </div>
+        <div class="row">
+        	<div class="col-xs-4 col-xs-push-4">
+                <form method="post" action="{{ path('restaurant_vote', { id: restaurant.id }) }}">
```
alexsegura: When the user has already voted, we shouldn't show the button.

app/DoctrineMigrations/Version20190708144248.php
```diff
@@ -0,0 +1,179 @@
...
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20190708144248 extends AbstractMigration
```
alexsegura: This migration shouldn't have been committed.

app/DoctrineMigrations/Version20190709124816.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20190709124816 extends AbstractMigration
```
alexsegura: This migration shouldn't have been committed, it is the same as `Version20190709120618`

app/DoctrineMigrations/Version20190709120618.php
```diff
@@ -0,0 +1,35 @@
...
+        $this->addSql('DROP INDEX uniq_eb95123f65effccf');
+        $this->addSql('ALTER TABLE restaurant DROP pledge_id');
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
```
alexsegura: Should be removed.

src/AppBundle/Resources/views/restaurant/restaurant_pledge_accepted.html.twig
```diff
@@ -0,0 +1,36 @@
...
+        <div class="row">
+        	<div class="col-xs-4 col-xs-push-4">
+                <h5>{{ restaurant.address.streetAddress }}</h5>
+                <h5>{{ restaurant.address.postalCode }}</h5>
+                <h5>{{ restaurant.address.addressLocality }}</h5>
+            </div>
+        </div>
+        <div class="row">
+        	<div class="col-xs-4 col-xs-push-4">
+                <form method="post" action="{{ path('restaurant_vote', { id: restaurant.id }) }}">
```
alexsegura: When the user has already voted, we shouldn't show the button.

app/DoctrineMigrations/Version20190709130229.php
```diff
@@ -0,0 +1,185 @@
...
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20190709130229 extends AbstractMigration
```
alexsegura: You should remove the unnecessary lines in this migration.

src/AppBundle/Controller/RestaurantController.php
```diff
@@ -261,9 +261,29 @@ public function indexAction($id, $slug, Request $request, SlugifyInterface $slug
 
             $numberOfVotes = count($restaurant->getPledge()->getVotes());
 
+            $userId = $this->getUser()->getId();
```
alexsegura: This whole piece of code checking if a user has already voted could be refactored by adding a method to the `Pledge` class. 

```php
public function hasVoted(UserInterface $user): bool
{
  // Loop on votes
  // If a vote has the same $user, return true
  // Else return false
}
```

src/AppBundle/Entity/Restaurant.php
```diff
@@ -722,4 +724,26 @@ public function addPreparationTimeRule($preparationTimeRule)
...
+    }
+
+    /**
+     * @param mixed $pledge
+     *
+     * @return self
+     */
+    public function setPledge($pledge)
+    {
+        $this->pledge = $pledge;
```
alexsegura: Actually, the linked restaurant isn't saved in database. 
You need to add `$pledge->setRestaurant($this);` here.

src/AppBundle/Resources/config/doctrine/Restaurant.PledgeVote.orm.yml
```diff
@@ -0,0 +1,33 @@
...
+            targetEntity: AppBundle\Entity\Restaurant\Pledge
+            inversedBy: votes
+            cascade:
+                - all
+            joinColumns:
+                pledge_id:
+                    referencedColumnName: id
+            orphanRemoval: false
+    manyToMany:
+        user:
```
alexsegura: Actually, **both** `pledge` & `user` associations are `manyToOne`

src/AppBundle/Entity/Restaurant/Pledge.php
```diff
@@ -0,0 +1,221 @@
...
+    public function setRestaurant($restaurant)
+    {
+        $this->restaurant = $restaurant;
+
+        return $this;
+    }
+
+    public function addVote(UserInterface $user)
+    {
+        if ($this->hasVoted($user) === false) {
```
alexsegura: Prefer syntax `if (!$this->hasVoted($user))`

src/AppBundle/Entity/Restaurant/Pledge.php
```diff
@@ -0,0 +1,221 @@
...
+    {
+        foreach ($this->getVotes() as $vote) {
+            if ($vote->getUser() === $user) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public function accept()
```
alexsegura: Add return type here `public function accept(): Restaurant`

alexsegura: If `$pledge` is not in state "new", this method should throw an Exception. 
Please add a unit test for this use case (i.e write a test that expects an Exception).

web/img/SuggererUnRestaurant.svg
```diff
@@ -0,0 +1,160 @@
+<?xml version="1.0" encoding="utf-8"?>
```
alexsegura: Please use dashes (`-`) in filenames, and English

`web/img/SuggererUnRestaurant.svg` → `web/img/suggest-restaurant.svg`

src/AppBundle/Resources/translations/messages.en.yml
```diff
@@ -438,6 +441,29 @@ form.settings.administrator_email.label: Administrator email
 form.settings.administrator_email.help: Notifications will be sent to this email
 form.settings.phone_number.label: Phone number
 form.settings.phone_number.help: This number will be displayed to customers
+
+form.suggest.name.label: name
+form.suggest.title: Suggest a restaurant
+form.suggest.submit.label: Register
```
alexsegura: "Enregistrer" is not "Register" in English

src/AppBundle/Resources/translations/messages.en.yml
```diff
@@ -6,6 +6,9 @@ navbar.myAccount: My Account
 base.footer.contactUs: Contact us
 base.footer.help: Help
 base.footer.legal: Legal notice
+base.footer.suggest: Suggest a restaurant
```
alexsegura: `base.footer.suggest` & `form.suggest.title` are the same sentence, [DRY](https://fr.wikipedia.org/wiki/Ne_vous_r%C3%A9p%C3%A9tez_pas)

tests/AppBundle/Entity/Restaurant/PledgeTest.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public function testAcceptThrowsException() {
+        $pledge = new Pledge();
+        $address = new Address();
+        $pledge->setState('new');
+        $pledge->setName('test');
+        $pledge->setAddress($address);
+
+        $this->expectException(Pledge::class);
```
alexsegura: The test is not passing. 
The expected exception is not `Pledge::class` 🙃
Also, you are trying to make the `accept` method throw an Exception here: accepting a Pledge in state "new" is totally ok.

src/AppBundle/Validator/Constraints/IsActivableRestaurantValidator.php
```diff
@@ -29,53 +29,55 @@ public function validate($object, Constraint $constraint)
...
-        ]);
-        if (count($contractErrors) > 0) {
-            $this->context->buildViolation($constraint->contractMessage)
-                ->atPath('contract')
-                ->addViolation();
-        }
-
-        // The validations below only make sense when the restaurant is created
-        if (null !== $object->getId()) {
+        if ($object->getState() !== 'pledge') {
```
alexsegura: Ok, we are changing an important piece of code here: we could allow enabling restaurants that are not ready for that (for example, with a missing Stripe account). I think it's the opportunity to **add a unit test** for this `IsActivableRestaurantValidator` class.

Check how it's done in the `OrderValidatorTest` class: you will need to extend `ConstraintValidatorTestCase` to have helper methods to test the validator. 

You should at least test that the validator adds a violation on the "enabled" field, when the restaurant is not in state "pledge".

PR #638, opened 2019-07-30
Allow configuring refundable product

### LINKED ISSUE #632: Allow configuring refundable containers for each product
It should be possible to enable refundable containers per product. 

Add a column `refundable_containers_enabled` (boolean) on the `product` table.
Also, add a `refundable_containers_unit` (float), which allows to configure how many containers are needed by a product (increment by 0.5)

On the admin interface, on the product screen, it should be possible to toggle options, and configure the units.

arnomasson:
Fix #632

### DIFF
--- a/js/app/product/form.js
+++ b/js/app/product/form.js
@@ -19,3 +19,11 @@ $(function() {
     size: [ 256, 256 ]
   })
 })
+
+$('#product_reusablePackagingEnabled').click(function(){
+  if($(this).prop("checked")) {
+    $('#reusable-unit-enabled').show();
+  } else {
+    $('#reusable-unit-enabled').hide();
+  }
+});
--- a/src/AppBundle/Form/ProductType.php
+++ b/src/AppBundle/Form/ProductType.php
@@ -27,6 +27,7 @@ use Symfony\Component\Form\Extension\Core\Type\MoneyType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
+use Symfony\Component\Form\Extension\Core\Type\NumberType;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Translation\TranslatorInterface;
@@ -39,7 +40,6 @@ class ProductType extends AbstractType
     private $productAttributeValueFactory;
     private $localeProvider;
     private $hasChangedName = false;
-
     public function __construct(
         ProductVariantFactoryInterface $variantFactory,
         ProductVariantResolverInterface $variantResolver,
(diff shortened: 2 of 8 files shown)


src/AppBundle/Resources/config/doctrine/Sylius.Product.orm.yml
```diff
@@ -18,6 +18,18 @@ AppBundle\Entity\Sylius\Product:
         deletedAt:
             type: datetime
             nullable: true
+        refundableContainersEnabled:
```
alexsegura: Please remove unnecessary options. You only need `type` & `options.default`

src/AppBundle/Form/ProductType.php
```diff
@@ -95,6 +95,21 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+        $builder
+            ->add('refundableContainersEnabled', CheckboxType::class, [
+                'required' => false,
+                'label' => 'form.product.refundable_enabled.label',
+            ])
+            ->add('refundableContainersUnit', NumberType::class, [
+                'label' => 'form.product.refundable_unit.label',
+                'html5' => true,
+                'attr'  => array(
+                'min'  => 0,
```
alexsegura: Missing indentation.

src/AppBundle/Resources/views/restaurant/product.html.twig
```diff
@@ -41,6 +41,20 @@
...
 
+  <div class="row">
+    <div class="col-sm-6">
+      {{ form_row(form.refundableContainersEnabled) }}
+      {% if form.refundableContainersEnabled.vars.data == true %}
+      <span id="refund-unit-enabled">
+        {{ form_row(form.refundableContainersUnit, {'attr': {'class': 'form-control input-number'}}) }}
+      </span>
+      {% else %}
+        <span id="refund-unit-enabled" style="display: none">
```
alexsegura: Unclosed `<span>` tag here. Moreover, I think it should be a `<div>` (because `<span>` should only contain [inline elements](https://developer.mozilla.org/fr/docs/Web/HTML/%C3%89l%C3%A9ments_en_ligne))

js/app/product/form.js
```diff
@@ -19,3 +19,11 @@ $(function() {
     size: [ 256, 256 ]
   })
 })
+
+$('#product_refundableContainersEnabled').click(function(){
```
alexsegura: Nice! 🙂

src/AppBundle/Form/ProductType.php
```diff
@@ -95,6 +95,21 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         // While price & tax category are defined in ProductVariant,
         // we display the fields at the Product level
         // For now, all variants share the same values
+        $builder
```
alexsegura: The comments on lines 95-97 aren't related to this code block, so this is confusing. 

Also, I'm thinking that those fields should be added only when the restaurant has `depositRefundEnable = true` 🤔 
So, this should be moved inside the `FormEvents::POST_SET_DATA` listener (below)

alexsegura:
Closing in favor of #640 & #641

PR #654, opened 2019-08-08
fix wrong phone number input

### LINKED ISSUE #645: client phone number problem
When people create an account with their facebook account, the first number of their phone number ( a 0 in France ) doesn't appear. So when you want to call their, you need to correct manually before calling.

arnomasson:
Fix #645


alexsegura:
Thank you @arnomasson 👍

PR #655, opened 2019-08-08
allow admin to remove tags

### LINKED ISSUE #552: Admin should be able to delete tags

arnomasson:
Fix #552

### DIFF
--- a/src/AppBundle/Controller/AdminController.php
+++ b/src/AppBundle/Controller/AdminController.php
@@ -596,7 +596,7 @@ class AdminController extends Controller
         if ($request->isMethod('POST') && $request->request->has('delete')){
             $id = $request->request->get('tag');
             $tag = $this->getDoctrine()->getRepository(Tag::class)->find($id);
-            $tagManager->unTagAll($tag);
+            $tagManager->untagAll($tag);
             $this->getDoctrine()->getManagerForClass(Tag::class)->remove($tag);
             $this->getDoctrine()->getManagerForClass(Tag::class)->flush();
 
--- a/src/AppBundle/Controller/AdminController.php
+++ b/src/AppBundle/Controller/AdminController.php
@@ -73,6 +73,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\Routing\Annotation\Route;
+use AppBundle\Service\TagManager;
 
 class AdminController extends Controller
 {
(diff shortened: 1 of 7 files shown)


src/AppBundle/Resources/views/admin/tags.html.twig
```diff
@@ -22,6 +22,12 @@
   <td><a href="{{ path('admin_tag', { slug: tag.slug }) }}">#{{ tag.id }}</a></td>
   <td>{{ tag.name }}</td>
   <td><span class="label label-default" style="background-color: {{ tag.color }};">{{ tag.color }}</span></td>
+  <td>
```
alexsegura: - You should also add a (empty) `<th>` in the `<thead>` section so that number of columns match
- Would look better with the button right-aligned (use Bootstrap's `text-right` class)

src/AppBundle/Controller/AdminController.php
```diff
@@ -588,8 +589,18 @@ public function editTaxationAction($code, Request $request)
     /**
      * @Route("/admin/settings/tags", name="admin_tags")
      */
-    public function tagsAction(Request $request)
+    public function tagsAction(Request $request, TagManager $tagManager)
     {
+        if ($request->isMethod('POST') && $request->request->has('delete')){
+            $id = $request->request->get('tag');
+            $tag = $this->getDoctrine()->getRepository(Tag::class)->find($id);
+            $tagManager->unTagAll($tag);
```
alexsegura: It should be `untagAll` (the `T` is in capital letters here). 
Class methods are case-insensitive, but let's be strict 🙂

src/AppBundle/Entity/Tag.php
```diff
@@ -10,6 +12,9 @@ class Tag
 
     private $slug;
 
+    /**
```
alexsegura: PHP DocBlocks are like this (stars are aligned vertically):
```
/**
 *
 */
```

src/AppBundle/Resources/views/admin/tags.html.twig
```diff
@@ -22,6 +22,12 @@
...
+  	<form method="POST">
+  	<input type="hidden" name="tag" value="{{ tag.id }}">
+  	<button class="btn btn-danger" type="submit" name="delete" onclick="return confirm('Are you sure to delete this tag ? all elements rely to it will not be able to show it or be grouped by it anymore.')">{{ 'basics.delete'|trans }}</button>
```
alexsegura: Can you translate the confirm message too please?

alexsegura: Please use a small button (`btn-xs`) to stay coherent with the rest of the app

PR #696, opened 2019-09-19
allow admin to export tags in csv tasks file

### LINKED ISSUE #688: export tags column
to isolate some tasks ; exporting tags column will be usefull.

arnomasson:
Fix #688


src/AppBundle/Form/TaskExportType.php
```diff
@@ -14,6 +14,7 @@
 use Symfony\Component\Form\FormError;
 use Symfony\Component\Form\Extension\Core\Type as FormType;
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Doctrine\Common\Collections\ArrayCollection;
```
alexsegura: Useless

PR #697, opened 2019-09-20
Fix

### LINKED ISSUE #639: Price should not be calculated when store has no pricing rule set
When a store has no pricing rule set, the price is calculated anyway. 

Moreover, when there are no pricing rules sets defined, we should show a better error message inviting to create one, instead of a blank dropdown.

arnomasson:
related to #639


src/AppBundle/Resources/translations/messages.fr.yml
```diff
@@ -122,6 +122,8 @@ restaurant.temporaryClosedWarning: Ce restaurant est fermé jusqu'au <strong>%da
   Vous pouvez commander pour plus tard !
 restaurant.noMenuWarning: 'Ce restaurant n''a pas de menu pour l''instant
 
+  '
+restaurant.form.noPrincingRule: 'Aucune règle de tarification n''a été instaurée, <a target="_blank" href="%editPricingRule%">cliquez-ici pour en créer une</a>.
```
alexsegura: Please remove `target="_blank"`

PR #706, opened 2019-09-26
Fix front for btn basics save on restaurant profile if not admin

arnomasson:
no issue created for this problem


r0xsh:
Fixed by 29a49bb1f65597510d8421c611e99e77a68e85bf

PR #747, opened 2019-10-10
Add CSV delimiter detection and set delimiter for CSV reader

### LINKED ISSUE #734: Support semicolons in CSV import
Excporting .CSV from Excel in Europe only provides ";" 

We have a lot of recurring tasks and imports are essential to manage them.

KaiCMueller:
Fixes #734 

User request was that CSV files using semicolon as separator should be supported.
The [Spout Libary](https://github.com/box/spout) is used for reading CSV files, which is expecting the delimiter to be set if it is not comma (default).
I added a short logic to detect the delimiter (most used character in CSV headline) in the CSV file and setting the delimiter in the reader.
Test for CSV with semicolon was added.
The following separators are supported now and will be auto-detected: comma, semicolon, tab, pipe

### DIFF
--- a/src/AppBundle/Utils/TaskSpreadsheetParser.php
+++ b/src/AppBundle/Utils/TaskSpreadsheetParser.php
@@ -165,7 +165,7 @@ class TaskSpreadsheetParser
         $mimeType = mime_content_type($filename);
 
         if (in_array($mimeType, self::MIME_TYPE_CSV)) {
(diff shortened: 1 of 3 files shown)


alexsegura:
Thank you @KaiCMueller 🙂 

thank-you

KaiCMueller:
@alexsegura My pleasure ❤️

Issue #684, opened 2019-09-10
Structured filename when exporting picture/signature


couleurmate:
Exempli gratia : `ID_address.name_date.jpg`

alexsegura:
It's done here

https://github.com/coopcycle/coopcycle-web/blob/47a0f82104d3107a10658847b441f1858a56d775/src/AppBundle/Controller/Utils/AdminDashboardTrait.php#L302-L327

KaiCMueller:
@alexsegura please assign me :) What name schema would you suggest?

alexsegura:
Well like @couleurmate is suggesting, he's one of our power users and knows best 😎

alexsegura:
Fixed in #749

PR #749, opened 2019-10-10
Create a structured filename for image download

KaiCMueller:
#684

User requested a structured filename when downloading task images.

Filename is now created in the following structure which is hopefully correct:
{image_id}_{slugify(address.name)}_{task.createdAt.format("Y-m-d")}.{extension}

If address.name is present, slugify is used to create a normalized name for replacing white spaces and other characters that might cause problems when downloading images depending on the client.

A test was added to test the filename generation. Since the method itself is protected and hidden in a trait, did some reflection magic to get it tested.

Found and fixed a bug in \AppBundle\Entity\Task::addImage

Please be aware I was not able to manually test the image upload and download since I did not find the functionalities in the web application. Probably it is meant to be used from the App or some other client.


src/AppBundle/Controller/Utils/AdminDashboardTrait.php
```diff
@@ -320,9 +326,38 @@ public function downloadTaskImage($taskId, $imageId, StorageInterface $storage)
...
+        );
+
+        return $fileName;
+    }
+
+    /**
+     * @required
+     * @param SlugifyInterface $slugify
+     */
+    public function setSlugify(SlugifyInterface $slugify)
```
alexsegura: You don't need this with dependency injection.

src/AppBundle/Controller/Utils/AdminDashboardTrait.php
```diff
@@ -27,6 +29,10 @@
 
 trait AdminDashboardTrait
 {
+
+    /** @var SlugifyInterface */
+    protected $slugify;
+
     protected function redirectToDashboard(Request $request)
```
alexsegura: Thanks to dependency injection, you can just add a `SlugifyInterface $slugify` parameter to this method, and it will automagically be injected 😉

src/AppBundle/Entity/Task.php
```diff
@@ -426,7 +426,7 @@ public function setImages($images)
 
     public function addImage($image)
     {
-        $this->images->add($images);
+        $this->images->add($image);
```
alexsegura: Nice catch 😱

alexsegura:
Thank you @KaiCMueller it's nice! I requested some changes to make it perfect 🙂 

> Please be aware I was not able to manually test the image upload and download since I did not find the functionalities in the web application. Probably it is meant to be used from the App or some other client.

Yes, you are right, you can't upload in the webapp. 
What you can do, is: 
- Rename some random image from the web to `123456.jpg`
- Copy it to a folder named `web/images/tasks/12/34` (2 levels of folders, first 2 chars, next 2 chars of the filename)
- Create a task in the admin dashboard
- Insert manually a new row in the `task_image` table
```
docker-compose exec postgres psql -U postgres -d coopcycle -c 'insert into task_image (task_id, image_name, created_at) values (<your task id>, '123456.jpg', CURRENT_TIMESTAMP)'
```

KaiCMueller:
Thanks for the feedback, I will implement the changes tomorrow. I thought I didn't want to extend the application too much just for this small feature, but I prefer the way with a separate class with public methods and dependency injection!

I did already test the downloading of images the way you described, but I do not know the image format and name that is actually used when uploaded. I will go through your documentation to get a better understanding of how task images are used.

Have a nice evening!

KaiCMueller:
@alexsegura did the changes :) let me know what you think.

src/AppBundle/Controller/Utils/AdminDashboardTrait.php
```diff
@@ -320,9 +322,16 @@ public function downloadTaskImage($taskId, $imageId, StorageInterface $storage)
...
-            basename($imagePath)
+            $this->getImageDownloadFileName($image, $slugify)
         );
 
         return $response;
     }
+
+    protected function getImageDownloadFileName(TaskImage $taskImage, SlugifyInterface $slugify)
+    {
+        $taskImageNamer = new TaskImageNamer();
```
alexsegura: I prefer constructor injection, but it's because I'm an object-oriented psycho 🤪
I mean, `getImageDownloadFileName($taskImage)` interface method is like "I give you an object, give me it's filename". The fact that it internally uses some object to slugify is implementation details.

```
$taskImageNamer = new TaskImageNamer($slugify);
return $taskImageNamer->getImageDownloadFileName($taskImage);
```

But there are no interfaces, it's overkill, and totally fine like this too 👍

alexsegura:
Thank you @KaiCMueller

PR #750, opened 2019-10-10
Use new Symfony events.

### LINKED ISSUE #719: Use new Symfony events
Like in 47a0f82104d3107a10658847b441f1858a56d775, we should replace soon-to-be deprecated events

jbdelhommeau:
fix #719

Kiss; I love open source.

### DIFF
--- a/src/AppBundle/Api/EventSubscriber/MaintenanceSubscriber.php
+++ b/src/AppBundle/Api/EventSubscriber/MaintenanceSubscriber.php
@@ -7,7 +7,7 @@ use Predis\Client as Redis;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\Translation\TranslatorInterface;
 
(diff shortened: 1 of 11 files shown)


src/AppBundle/Security/Firewall/TokenBearerListener.php
```diff
@@ -25,15 +25,14 @@ public function __construct(
...
-    )
-    {
+    ) {
         $this->tokenStorage = $tokenStorage;
         $this->authenticationManager = $authenticationManager;
         $this->jwtTokenAuthenticator = $jwtTokenAuthenticator;
         $this->httpMessageFactory = $httpMessageFactory;
     }
 
-    public function handle(GetResponseEvent $event)
```
alexsegura: Unfortunately, `ListenerInterface` is [still declared](https://github.com/symfony/symfony/blob/b1ea2cd7a6b7f644a528f78f7814e01a7fabe181/src/Symfony/Component/Security/Http/Firewall/ListenerInterface.php) with `GetResponseEvent` 😢 

```
PHP Fatal error:  Declaration of AppBundle\Security\Firewall\TokenBearerListener::handle(Symfony\Component\HttpKernel\Event\RequestEvent $event) must be compatible with Symfony\Component\Security\Http\Firewall\ListenerInterface::handle(Symfony\Component\HttpKernel\Event\GetResponseEvent $event)
```

This interface is [deprecated since Symfony 4.3](https://github.com/symfony/symfony/blob/b1ea2cd7a6b7f644a528f78f7814e01a7fabe181/src/Symfony/Component/Security/Http/Firewall/ListenerInterface.php), they say "turn listeners into callables instead".
So you can try to update the usage, but it's totally fine if you just revert your changes 😉

jbdelhommeau: OK in fact. I revert changes for the `TokenBearerListener.php`. :)

PR #751, opened 2019-10-11
Followup for #749: Refactor class instantiation

KaiCMueller:
Followup to https://github.com/coopcycle/coopcycle-web/pull/749


KaiCMueller:
@alexsegura I was confused but now got your point. Couldn't let this mistake leave in the application, so here is the fix :)

Issue #596, opened 2019-07-10
Fix OneToOne associations


alexsegura:
Some associations are wrongly configured using `OneToOne`, but they should be `OneToMany/ManyToOne` [Stack Overflow](https://stackoverflow.com/questions/8205606/doctrine-onetoone-incorrectly-generating-a-unique-index)

For example, `Restaurant` & `Address`. The same address cannot appear twice in the restaurant table.

johnnybigoode:
Noticed that this was mentioned @ https://github.com/coopcycle/coopcycle-web/pull/766#discussion_r336172690 
Is this still an issue?

If yes, what relations should be one to many?

- restaurants can have several cuisines
- addresses can have several restaurants (like the food court of a mall)

PR #766, opened 2019-10-17
#596 | WIP: Fix Restaurant and Cuisine associations

dollyd:
PR for #596: Fix associations between Restaurant and Cuisine

### DIFF
--- a/src/AppBundle/Resources/config/doctrine/Cuisine.orm.yml
+++ b/src/AppBundle/Resources/config/doctrine/Cuisine.orm.yml
@@ -20,4 +20,8 @@ AppBundle\Entity\Cuisine:
             unique: false
             nullable: true
             precision: 0
+    manyToMany:
+      owners:
+        targetEntity: AppBundle\Entity\Restaurant
+        mappedBy: servesCuisine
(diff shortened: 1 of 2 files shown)


src/AppBundle/Resources/config/doctrine/Cuisine.orm.yml
```diff
@@ -20,4 +20,8 @@ AppBundle\Entity\Cuisine:
             unique: false
             nullable: true
             precision: 0
+    manyToMany:
+      owners:
```
alexsegura: This generates the following error:
```
Property AppBundle\Entity\Cuisine::$owners does not exist
```
It should be `restaurants`, I suppose, but you would need to add the corresponding property to the `Cuisine` class.
Just to be sure: why did you do this? To be able to lookup all restaurants matching a certain cuisine?

dollyd: Yes - restaurants can have several cuisines

alexsegura:
Not relevant since df29827f62bb2e4ce8d2e9eff8bac99f97c6903d

Issue #789, opened 2019-10-25
Docker compose - php container exited with code 1


nodesoccoop:
Hi,
when i start my instance i got this error:
```
php_1          | ocramius/package-versions:  Generating version class...
php_1          | ocramius/package-versions: ...done generating version class
php_1          | > Incenteev\ParameterHandler\ScriptHandler::buildParameters
php_1          | Updating the "app/config/parameters.yml" file
php_1          | > @php bin/console cache:clear --no-warmup
php_1          |
php_1          |  // Clearing the cache for the dev environment with debug
php_1          |  // true
php_1          |
php_1          |  [OK] Cache for the "dev" environment (debug=true) was successfully cleared.
php_1          |
php_1          | > @php bin/console assets:install --symlink --relative web
php_1          |
php_1          |  Trying to install assets as relative symbolic links.
php_1          |
php_1          |  --- ----------------------------- ------------------
php_1          |       Bundle                        Method / Error
php_1          |  --- ----------------------------- ------------------
php_1          |   ✔   ApiPlatformBundle             relative symlink
php_1          |   ✔   CraueConfigBundle             relative symlink
php_1          |   ✔   FOSJsRoutingBundle            relative symlink
php_1          |   ✔   JMoseCommandSchedulerBundle   relative symlink
php_1          |  --- ----------------------------- ------------------
php_1          |
php_1          |  [OK] All assets were successfully installed.
php_1          |
php_1          | + php bin/console doctrine:database:create --if-not-exists '--env=dev'
php_1          | Database "coopcycle" for connection named default already exists. Skipped.
php_1          | + php bin/console doctrine:query:sql 'CREATE EXTENSION IF NOT EXISTS postgis' '--env=dev'
php_1          | int(0)
php_1          | + php bin/console doctrine:query:sql 'CREATE EXTENSION IF NOT EXISTS postgis_topology' '--env=dev'
php_1          | int(0)
php_1          | + php bin/console doctrine:query:sql 'CREATE EXTENSION IF NOT EXISTS pg_trgm' '--env=dev'
php_1          | int(0)
php_1          | + php bin/console doctrine:database:create --if-not-exists '--env=test'
php_1          | Database "coopcycle_test" for connection named default already exists. Skipped.
php_1          | + php bin/console doctrine:query:sql 'CREATE EXTENSION IF NOT EXISTS postgis' '--env=test'
php_1          | int(0)
[...]

PR #801, opened 2019-10-30
Cleanup/simplify Dockerfile

Pandoraaa:
May fix some problems on Linux (#789)

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -60,14 +60,6 @@ RUN set -xe \
     && mv composer.phar /usr/local/bin/composer \
     && apk del .fetch-deps
 
-# https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
-ENV COMPOSER_ALLOW_SUPERUSER 1
-
-RUN composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative \
-    && composer clear-cache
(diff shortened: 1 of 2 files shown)


alexsegura:
Thank you for this. 

Definitely, I think the `umask(0000)` is the fix we need. A Symfony 4 front controller file looks like this: 

https://github.com/symfony/recipes/blob/7be9ee0bbaf2ee0a3fdcfffdb1567917290a36c3/symfony/framework-bundle/4.2/public/index.php

```

if ($_SERVER['APP_DEBUG']) {
    umask(0000);
    Debug::enable();
}
```

alexsegura:
See #803

alexsegura:
@Pandoraaa I rebased your commits and added some more.

Can you try again to see if it fixes your problems?

```
rm -rf vendor
rm -rf var/jwt
docker-compose down -v
docker-compose build php
docker-compose up
```

alexsegura:
@Pandoraaa I think all remaining problems are fixed in #816

Pandoraaa:
@alexsegura Oh yeah now everything works fine! 
Happy to see that the umask solution did the job :+1: 

I think the docker build could be optimized, it was sooooo long... :timer_clock:

alexsegura:
Yeah what is long is compiling the PHP extensions. The simplest way to speed up the start of the containers would be to pre-build it, and upload it to Docker Hub.

PR #1229, opened 2020-03-14
Allow storing arbitrary data in tasks

### LINKED ISSUE #1178: Allow attaching data to tasks
For deposit-refund, we need to all messengers to report how many containers were collected. 
Basically, the app needs to present the messenger with a number input, and to post this number to the backend. 

To implement this is a generic and reusable manner, we need to introduce a way to attach arbitrary data payload to tasks, as well as a way to render a form with those inputs in the app.

#### Fields definition

Each task can have a `TaskField` collection, which indicates what kind of entry it accepts

`GET /api/tasks/1`

```json
{
  "fields": [
    {
      "type": "number", 
      "name":"containers_returned",
      "label":"How many containers?"
    }
  ]
}
```

On the app, depending on the `fields` property, a dynamic form with a number input is rendered on the screen to complete a task.

#### Sending data

When completing a task, it is possible to specify a `data` payload

`PUT /api/tasks/1/done`

```json
{
  "data": [
    {"name": "containers_returned", "value": 4}
  ]
}
```

alexsegura:
Fixes #1178

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,10 @@ osrm:
 	@docker-compose run --rm osrm osrm-partition /data/data.osrm
 	@docker-compose run --rm osrm osrm-customize /data/data.osrm
 
-phpunit:
+test-db-sync:
 	@docker-compose exec php php bin/console doctrine:schema:update --env=test --force --no-interaction --quiet
+
+phpunit:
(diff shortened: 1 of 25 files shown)


alexsegura:
Superseded by #2524

PR #1256, opened 2020-03-26
android: send both 'notification+data' and 'data-only' (notification) messages

vova1vova:
This change would allow android app to handle 'notifications' (if they are sent as 'data-only' messages) in the background and later on (when implemented on the app side) display a notifcation to the user locally without relying on firebase SDK, which gives the app more flexibility. 

This change should not affect current app users, because current version does not handle 'data-only' messages at all, and in a new version: 

- "tasks:changed" event would be handled the same way as now
- "order:created" event would be handled the same in the foreground and in the background a user would see a current notification + the app would be opened if possible (only on android)

### DIFF
--- a/src/AppBundle/Service/RemotePushNotificationManager.php
+++ b/src/AppBundle/Service/RemotePushNotificationManager.php
@@ -7,22 +7,26 @@ use AppBundle\Entity\RemotePushToken;
 use Kreait\Firebase\Factory as FirebaseFactory;
 use Kreait\Firebase\Exception\ServiceAccountDiscoveryFailed;
 use Kreait\Firebase\Messaging\CloudMessage;
+use Psr\Log\LoggerInterface;
 
 class RemotePushNotificationManager
 {
     private $firebaseFactory;
     private $apns;
     private static $enabled = true;
+    private $logger;
 
     public function __construct(
         FirebaseFactory $firebaseFactory,
         \ApnsPHP_Push $apns,
-        string $apnsCertificatePassPhrase)
+        string $apnsCertificatePassPhrase,
+        LoggerInterface $logger)
     {
         $this->firebaseFactory = $firebaseFactory;
 
         $apns->setProviderCertificatePassphrase($apnsCertificatePassPhrase);
         $this->apns = $apns;
+        $this->logger = $logger;
     }
 
     public static function isEnabled()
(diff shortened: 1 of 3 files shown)


src/AppBundle/Service/RemotePushNotificationManager.php
```diff
@@ -185,7 +189,16 @@ public function send($message, $recipients, $data = [])
...
-        $this->apns($message, $apnsTokens, $data);
+        //todo temporary send both "notification+data" and "data-only" messages on android
+        // reasons:
+        // 1. backward compatibility with the old app-versions
+        // 2. app does not always handle the case when "notification+data" is delivered in
+        // the background
+        // when #2 is done and #1 is not relevant any more we should only send "data-only" messages
+
+        $this->fcm($textMessage, $fcmTokens, $data); // send "notification+data" message
+        $this->fcm(null, $fcmTokens, $data); // send "data-only" message
```
alexsegura: Does this send the notification twice?

vova1vova: yes, it does, but they would be of two different types "notification+data" and "data-only". There would be the following scenarios:

- current version: the behavior would stay the same, it does not handle "data-only" messages
- new version (in the foreground): the behavior would also be the same, it would handle "data-only" messages and ignore "notification+data" messages
- new version (in the background): "notification+data" message would be shown in the notification center, "data-only" would be handled only for "order:created" event - the app would be opened.

To sum up, it would be an issue only in one case, when "order:created" event is received in the background: the app would be opened and there would be also a notification in the notification center. 

It's not ideal, but I don't have any ideal solution in mind. First of all, we still need to send "notification+data" for the current app version (it would be cool to know not only platform but also an app version for each push token, but it's quite some work to make it happen), and the second, some changes are required on the app side to be able to get rid from "notification+data" message (the app needs to show notification in the notification center, when it receives a "data-only" message)

src/AppBundle/Service/RemotePushNotificationManager.php
```diff
@@ -7,22 +7,26 @@
...
     private $apns;
     private static $enabled = true;
+    private $logger;
 
     public function __construct(
         FirebaseFactory $firebaseFactory,
         \ApnsPHP_Push $apns,
-        string $apnsCertificatePassPhrase)
+        string $apnsCertificatePassPhrase,
+        LoggerInterface $logger)
```
alexsegura: Can you fix the unit tests please? 😉 
You can just inject a `NullLogger` in tests (there are examples in other tests)

vova1vova: fixed

alexsegura:
Thank you @vova1vova

PR #1503, opened 2020-06-22
Full pt-BR support and include BRL currency

miguelpeixe:
Some notes on this PR: 

 - I used pt-BR naming instead of pt_BR because of html lang attibute standard
 - I had to change how antd locale is imported because of naming conflicts with actual pt_BR js messages locale. Doesn't look pretty. I assume would be better if antd locales were imported directly to an object with `require()` instead, avoiding this type of problem in the future. I can make this PR as well.


alexsegura:
Thank you @miguelpeixe 

This makes me think that it's quite painful to add a new language 🙁 
There should be a central file defining the locales
Also to reduce the bundle size I suppose it would be better to be able to configure/restrict the locales that are bundled with Webpack

miguelpeixe:
Thanks for accepting my PR :)

Finding the translation files wasn't really the problem. Defining the locales throughout the code was quite difficult. I still don't know how the languages are populated inside `sylius_locales`, I had to reset my database to make sure it was inserted to the table. Aside from that, once everything is setup, weblate is a great option to keep it updated.

alexsegura:
The Sylius locales are creating via a console command

```
bin/console coopcycle:setup
```

PR #1505, opened 2020-06-22
Add moment pt-BR locale

miguelpeixe:
I actually missed a spot for the pt-BR locales. Moment locales setup was pretty hidden hehe


alexsegura:
Thank you @miguelpeixe 

Yeah this is what I was saying about the pain to add a new locale, too much files to change

PR #1581, opened 2020-07-15
Translations

coopcycle-weblate:


alexsegura:
Thank you @jfoclpf for adding pt-BR 🇵🇹

jfoclpf:
Hi @alexsegura, do not thank only to me, but to the members of the cooperative of Lisbon who did a great collective job. They already applied for the federation and thus we are advancing the todo work `:)`

PR #1611, opened 2020-07-27
add registration ask for company

flug:


src/Resources/translations/messages.fr.yml
```diff
@@ -427,6 +428,12 @@ form.address.floor.label: Étage
 form.address.description.label: Instructions
 form.address.description.placeholder: 2ème porte à gauche…
 form.address.company.label: Entreprise
+form.address.company_name.label: Nom de l'entreprise
+form.address.company.business_reference.label: Contact de la personne référente pour l’entreprise
+form.address.company.function.label: Poste
+form.address.company.collaborator_number.label: Numbre de collaborateur
```
alexsegura, line 434: `s/Numbre/Nombre`

src/Resources/translations/messages.fr.yml
```diff
@@ -427,6 +428,12 @@ form.address.floor.label: Étage
...
+form.address.company.meal_estimate.label: Numbre de collaborateur
```
alexsegura, line 435: `s/Numbre de collaborateur/Estimation du nombre de repas (par jour)`

src/Resources/translations/messages.fr.yml
alexsegura, line 434: `s/collaborateur/collaborateur-ices`

src/Resources/translations/messages.en.yml
```diff
@@ -2,6 +2,7 @@ global.changesSaved: Your changes have been saved.
 navbar.login: Login
 navbar.myAccount: My Account
 base.footer.contactUs: Contact us
+base.footer.companyRegister: Become a partner
 base.footer.help: Help
 base.footer.legal: Legal notice
 base.footer.suggest: Suggest a restaurant
```
alexsegura, line 8: Can you also add the `form.address.company*` keys to `messages.en.yml` please? The `en` language is the fallback/default, so translations should be present here so that they can be crowd-translated [here](http://translate.coopcycle.org/)

src/Form/Company/RegistrationType.php
```diff
@@ -0,0 +1,67 @@
...
+use AppBundle\Form\AddressType;
+use AppBundle\Message\Company\RequestRegistration;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\DataMapperInterface;
+use Symfony\Component\Form\Extension\Core\Type\IntegerType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class RegistrationType extends AbstractType implements DataMapperInterface
```
alexsegura, line 15: Just curious: why do we need to implement `DataMapperInterface`?
Usually I do this:

```php
public function configureOptions(OptionsResolver $resolver)
{
    $resolver->setDefaults(array(
        'data_class' => RequestRegistration::class,
    ));
}
```

flug: "RequestRegistration" has elements in the constructor. If I pass it as you put it, I would have to remove the arguments and use setter, which makes the class anemic.

flug: ```DataMapperInterface``` allows you to implement 2 methods to match the need that the class has to function properly.

PR #1625, opened 2020-07-31
divided the config docker test

flug:


docker-compose.test.yml
```diff
@@ -0,0 +1,26 @@
...
+      - './docker/nginx_test/conf.d:/etc/nginx/conf.d:ro'
+      - './web:/var/www/html/web:ro'
+      - './vendor:/var/www/html/vendor:ro'
+
+  stripe_mock:
+    image: stripemock/stripe-mock:latest
+    ports:
+      - '12111:12111'
+
+  browserless:
```
alexsegura, line 21: The `browserless` container is not used for tests, but to generate PDFs

PR #1620, opened 2020-07-30
[WIP] add relationship to customer -> customer group -> organization

flug:
@alexsegura just the beginning of the PR to see if it fits the criteria?

PR #1648, opened 2020-08-18
add cart possibility to order via b2b

flug:


src/Controller/UserController.php
```diff
@@ -135,4 +136,16 @@ public function confirmInvitationAction(Request $request, string $code,
             'form' => $form->createView()
         ]);
     }
+
+    /**
+     * @Route("/mode-pro", name="pro_account_cart")
+     */
+    public function proCartAction(Request $request, OrderFactory $orderFactory)
+    {
+        $orderFactory->createNew();
```
alexsegura, line 145: Why?

src/Command/InitDemoCommand.php
```diff
@@ -457,7 +457,7 @@ private function createRestaurant(Entity\Address $address, TaxCategoryInterface
         $contract->setCustomerAmount(350);
         $contract->setFeeRate(0);
 
-        $restaurant = new Entity\Restaurant();
+        $restaurant = new Entity\LocalBusiness();
```
alexsegura, line 460: Can you please extract this change to a separate commit?

src/Command/SetupCommand.php
```diff
@@ -63,7 +63,8 @@ class SetupCommand extends Command
 
     private $channels = [
         'web' => 'Web',
-        'app' => 'App'
+        'app' => 'App',
+        'pro' => 'Pro'
```
alexsegura, line 67: Can you please extract this change to a separate commit?

templates/_partials/user/account_dropdown.html.twig
```diff
@@ -8,6 +8,13 @@
         {% trans %}navbar.myAccount{% endtrans %}
       </a>
     </li>
+    {% if context_channel.channel.code == 'pro' %}
+      <li><a href="{{ path('pro_account_cart', {'redirect_to': app.request.uri, '_channel_code' : 'web'}) }}">{% trans %}navbar.lambdaCart{% endtrans %}</a></li>
```
alexsegura, line 12: Using the `_channel_code` parameter only works in dev, because the [FakeChannelCodeProvider](https://github.com/Sylius/Sylius/blob/8e50539ce5015d98403b6c25637e4b6bb1b6ddf9/src/Sylius/Bundle/ChannelBundle/Context/FakeChannel/FakeChannelCodeProvider.php) is injected (see how the [dependency injection](https://github.com/Sylius/Sylius/blob/8e50539ce5015d98403b6c25637e4b6bb1b6ddf9/src/Sylius/Bundle/ChannelBundle/DependencyInjection/SyliusChannelExtension.php#L34-L38) works). This won't work in production.

src/Controller/UserController.php
```diff
@@ -135,4 +136,12 @@ public function confirmInvitationAction(Request $request, string $code,
             'form' => $form->createView()
         ]);
     }
+
+    /**
+     * @Route("/mode-pro", name="pro_account_cart")
+     */
+    public function proCartAction(Request $request)
```
alexsegura, line 143: Instead of adding a new action, you should use the mechanism of Sylius for resolving/storing the channel. 
See [how it's configured in DI](https://github.com/Sylius/Sylius/blob/8e50539ce5015d98403b6c25637e4b6bb1b6ddf9/src/Sylius/Bundle/ChannelBundle/Resources/config/services.xml#L40-L60), and see [HostnameBasedRequestResolver](https://github.com/Sylius/Sylius/blob/8e50539ce5015d98403b6c25637e4b6bb1b6ddf9/src/Sylius/Component/Channel/Context/RequestBased/HostnameBasedRequestResolver.php) for an example of resolving the channel based on the request. 

Just look at how [FakeChannelContext](https://github.com/Sylius/Sylius/blob/8e50539ce5015d98403b6c25637e4b6bb1b6ddf9/src/Sylius/Bundle/ChannelBundle/Context/FakeChannel/FakeChannelContext.php) works (also, see the [service definition](https://github.com/Sylius/Sylius/blob/8e50539ce5015d98403b6c25637e4b6bb1b6ddf9/src/Sylius/Bundle/ChannelBundle/Resources/config/services/integrations/debug.xml#L26-L31)). It should be implemented in the same way. 

---

Actually, we already have [a service tagged with `sylius.context.channel.request_based.resolver`](https://github.com/coopcycle/coopcycle-web/blob/458c4549c9dc3c860177bc0107df4946851286ef/app/config/services.yml#L622-L624), which resolves the channel from a HTTP header.

src/EventListener/ChangeChannelListener.php
```diff
@@ -0,0 +1,25 @@
...
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+        if (!$event->getRequest()->query->has('change_channel')) {
+            return;
+        }
+
+        $response = $event->getResponse();
+        $newChannel = $event->getRequest()->query->get('change_channel');
```
alexsegura, line 21: It should not be possible to inject the channel via a request parameter without sanitizing input. 

In the current implementation, if you do `?change_channel=notexists`, it stores the `notexists` string in cookies, and `ProChannelContext ` then returns a non-existing channel, leading to the following error:

```
Return value of AppBundle\Sylius\Channel\ProChannelContext::getChannel() must implement 
interface Sylius\Component\Channel\Model\ChannelInterface, null returned
```

Then it goes into an infinite error 500.

---

At least, the value of `$newChannel` should only allow a finite list of values (`web` or `pro`). We don't want customers to switch to the `app` channel. 
For other values, it should just do nothing.

src/Sylius/Channel/ProChannelContext.php
```diff
@@ -0,0 +1,34 @@
...
+    public function __construct(ChannelRepositoryInterface $channelRepository, RequestStack $request)
+    {
+
+        $this->channelRepository = $channelRepository;
+        $this->masterRequest = $request->getMasterRequest();
+    }
+
+    public function getChannel(): ChannelInterface
+    {
+        return $this->channelRepository->findOneByCode($this->masterRequest->cookies->get('channel_cart', 'web'));
```
alexsegura, line 32: When the channel does not exist, the following error occurs:

```
Return value of AppBundle\Sylius\Channel\ProChannelContext::getChannel() must implement 
interface Sylius\Component\Channel\Model\ChannelInterface, null returned
```

When the channel does not exist, it should throw a `ChannelNotFoundException`, for the chain to execute properly.
See [CompositeChannelContext](https://github.com/Sylius/Sylius/blob/34fcad9732bb53eab1d02a721900bb2b3e1edda2/src/Sylius/Component/Channel/Context/CompositeChannelContext.php)

alexsegura, line 32: As the cookie is written during the `kernel.response` event, `? change_channel=pro` does not have effect immediately, but on the subsequent request. The value in the request parameter should have a priority over the cookie. See how it's done in [FakeChannelCodeProvider](https://github.com/Sylius/Sylius/blob/34fcad9732bb53eab1d02a721900bb2b3e1edda2/src/Sylius/Bundle/ChannelBundle/Context/FakeChannel/FakeChannelCodeProvider.php)

alexsegura (changes requested):

alexsegura (dismissed):

src/Sylius/Channel/ProChannelContext.php
alexsegura: This is still a problem @flug

alexsegura:
@flug can you rebase please?

alexsegura (changes requested):
Here is a visual explanation of the problem explained in https://github.com/coopcycle/coopcycle-web/pull/1648#discussion_r473038047

change_channel_bug

templates/_partials/user/account_dropdown.html.twig
```diff
@@ -8,6 +8,19 @@
...
+    {% if context_channel.channel.code == 'pro' %}
+      {% set queryParams = queryParams|merge({change_channel: 'web'}) %}
+      {% set accountType = 'navbar.lambdaCart' %}
+    {% else %}
+      {% set queryParams = queryParams|merge({change_channel: 'pro'}) %}
+      {% set accountType = 'navbar.proCart' %}
+    {% endif %}
+    {% set currentPath = path(app.request.attributes.get('_route'),
+      app.request.attributes.get('_route_params')|merge(queryParams)) %}
+    <li><a href="{{ currentPath }}">{{ accountType|trans }}</a></li>
```
alexsegura, line 22: Only "authorized" users should be able to switch channels, i.e users belonging to a group associated with an Organization

alexsegura (changes requested):

alexsegura (changes requested):
No, the commit 58ea23961aea9a284160f69619a66bd4894bdc56 shouldn't be there. 

The way to know if the user is has the feature enabled, is to check if the user belongs to a group associated to an organization, as implemented in #1620

Basically, it means (in SQL): 

```sql
SELECT
  CASE WHEN COUNT(*) > 0 THEN 'in group' ELSE 'not in group' END
FROM api_user u 
JOIN sylius_customer c ON u.customer_id = c.id
JOIN organization_config oc ON oc.group_id = c.customer_group_id
```

PasLoin:
@flug @alexsegura 

We have a demand from one of our customer that ask if it's possible to limit to b2b order on the plateform (with identified final client linked to the restorant). Is it in the scope of this branch ?

PR #1696, opened 2020-09-02
Allow selecting an area on the map

### LINKED ISSUE #1341: Assign tasks by selecting an area on the map screen
For particular deliveries, the same product is dropped at lots of addresses (free press, prospectus or letters, ...). There are no differences between the tasks exept the adress. 

One way to dispatch theses tasks is to assign them to courriers according to their preferred neighborhoods or by separating deliveries with natural limits (river, highway, railways, ...).

It would be facilitating to select an area on the map -and the adresses inside this area- and assign it to a courier.

alexsegura:
Fixes #1341 

area-select


rayon9coop:
🤩👍👏

PR #1705, opened 2020-09-03
use shared frontend library

vladimir-8:
only refactoring, code that shows a colored line near linked tasks is moved into the [shared library ](https://github.com/coopcycle/coopcycle-frontend-js)to be re-used in the mobile app https://github.com/coopcycle/coopcycle-app/pull/748

Screen Shot 2020-09-03 at 17 41 18


vladimir-8:
replaced with https://github.com/coopcycle/coopcycle-web/pull/1754

Issue #1311, opened 2020-04-11
Introduce "shops"


alexsegura:
#### The problem

One of the main design problems we have, is that "B2B customers" (i.e, customers that **don't** have a customer-facing page) are represented as a `Store` entity, under the `/api/stores` endpoint. 

The have a slightly different behaviour & features. 
For example, a `Store` may have a `PricingRuleSet` & a `TimeSlot`.

Actually, a `Store` & a `Restaurant` are pretty similar. They are both a `LocalBusiness`, and bike messenger companies deliver their products. The bike messenger company may deliver their products to customers, or businesses (or both). 

The biggest difference is that a `Store` doesn't have customer-facing page. 

No matter if this is a `Store` or a `Restaurant`, the main question is: 
> Do they sell goods online via our platform?
> Do they expose their catalog of products online publicly?

So it's more a matter of "enabling" online shopping for a `Restaurant` or a `Store`.

#### Introducing Shop concept

For a better separation of concerns, a customer-facing page should become a first-class citizen. 
A pretty simple way to do this is to extract all the logic into a `Shop` class. 
A `Shop` just exposes the catalog of a `LocalBusiness` online. 

A `LocalBusiness`, no matter if it is a `Store` or a `Restaurant`, may have **several** `Shop`. This would easily fix the problem of franchises, where the same restaurant brand has several locations in the same city, with more or less the same catalog (currently, they are duplicated). 

This would also allow to easily add online shopping later for a B2B customer. Since the beginning of the COVID-19 pandemic, we have seen lots of businesses switch from B2B to B2C.

#### Refactoring

The `Restaurant` & `Store` entities should be unified under a single `local_business` table in database, with inheritance mapping. All the entries in the `store` table need to be moved to the `local_business` table. We will lose the previous ids but YOLO. 

A `LocalBusiness` may have one or many `Shop`

The catalog of products should be abstracted instead of being named a `Menu`. We can still expose a `/api/restaurants/{id}/menu` API operation but only for restaurants, for stores it should return a 404. 

On the opposite, `Restaurant` should be closer to the way `Store` behaves, regarding deliveries for example. Currently, deliveries _may_ be associated to a `Store`, but not directly to a `Restaurant` (the relationship is restaurant <-> order <-> delivery).

#### API endpoints (to be confirmed)

- `/api/shops` should slowly replace `/api/restaurants`
- `/api/restaurants` should be kept for backward compatibility
[...]

Issue #690, opened 2019-09-17
Allow inserting another linked stop through dashboard


alexsegura:
When there are 2 linked tasks on the dashboard, it should be possible to add another stop (in the middle, at the end...). 

There are several possibilities for the UI/UX: right-click & context menu, drag'n'drop...

alexsegura:
We could use [combining items](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/combining.md)

ERAMANrepartos:
This seems like its part of mult-dropoff deliveries and / or routes, but either way, is being dealt with elsewjhere

PR #1729, opened 2020-09-15
Attach tasks to organizations

### LINKED ISSUE #1402: Ability to assign manually created tasks to a Store.
I see it as two tasks (1 pickup and 1 dropoff) that have been created before a Store existed on the platform, they can be assigned to a specific Store created at a later date and by assigning those two tasks an order is created.

alexsegura:
Fixes #1402

Under the hood, this pull request introduces Organizations (#1311) as the top level entity to represent a customer. 
Technically, tasks are attached to an Organization, and Stores & Restaurants too. However, the user interface still shows Stores/Restaurants, for backward compatibility. 

**TODO**

- [x] Attach task to store when created via API

This introduces a new view for tasks as a list, where it's possible to attach tasks to stores

multi-select-tasks

--- 

On the dashboard, when the task is attached to an organization (a store or a restaurant), it is displayed like this:

Capture d’écran 2020-09-15 à 13 42 28

### DIFF
--- a/app/DoctrineMigrations/Version20200915100034.php
+++ b/app/DoctrineMigrations/Version20200915100034.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20200915100034 extends AbstractMigration
+{
+    public function getDescription() : string
+    {
+        return '';
+    }
+
+    private function linkToRestaurants()
+    {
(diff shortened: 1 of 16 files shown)


agichim:
If tasks are not linked together as an order they can still be attached under a store but will still be unlinked, right?

alexsegura:
@agichim yes, they would still be unlinked, it would not "reconstruct" a delivery. 
Should we also allow to link tasks together into a delivery? It's #690

agichim:
#690 is a great issue! I guess the ideal would be to link tasks into an order or not if you don't want to. But, as we discussed about not having tasks without stores, maybe tasks should always be linked into an order. I'm not sure yet about that.

alexsegura:
@agichim no, tasks shouldn't enforced to be attached to an order, it couples too much things together between e-commerce & dispatch.
But being able to select some tasks, and "convert" them into an order, seems good.

agichim:
Makes sense then!

agichim (approved):

alexsegura:
Let's open another issue about being able to create orders from tasks.
It needs to be discussed a little bit to make it right 🙂

agichim:
I'll check if there is an issue opened already. I feel I might have done it myself?

agichim:
Here's the issue #1731

Issue #436, opened 2018-09-16
Installation : make install lead to runtime exception error on Linux


freelibre:
**System information**
Fedora 28
Linux 4.18.7-200.fc28.x86_64 

**Steps to reproduce**
`git clone https://github.com/coopcycle/coopcycle-web`
`cd coopcycle-web/`
`make install`

**Expected result:**
No error

**Actual result:**
[RuntimeException]                                              
  /srv/coopcycle/vendor does not exist and could not be created. 

**Detail logs:**
[MakeInstall.log](https://github.com/coopcycle/coopcycle-web/files/2386887/MakeInstall.log)

Atala:
We have this workaround for now on (if you find the proper fix it will be better!)
```
Hi @florent_k I had the same exact issue on Debian 7 and had to change the Dockerfile to use uid 82 instead of www-data as well as chmodding var/cache and vendor to 777 (unsafe but I guess it is ok for testing)
```

Atala:
Could you try this workaround? (or even find a proper fix?)
`had to change the Dockerfile to use uid 82 instead of www-data as well as chmodding var/cache and vendor to 777 (unsafe but I guess it is ok for testing)`

freelibre:
Thank you for your feedback. Sorry I don't understand yet what you mean by "to change the Dockerfile to use uid 82 instead of www-data" also I don't see any vendor directory or file in the repository.
I have tried to change permission of var/cache to 777 before running a make install but it still through the same error message output in red.
`[RuntimeException]                                              
  /srv/coopcycle/vendor does not exist and could not be created.  `
I've also tried to create manually  /srv/coopcycle/vendor but I'm still getting the same error message.
I will investigate further and will let you know.

freelibre:
May be a fix could be to add user 82 to www-data group with:
usermod -u 82 www-data

Senzaki:
Or simply:

```
sudo chown -R $(id -u):82 coopcycle-web
sudo chmod -R g+w coopcycle-web
```

A bit of a hack but it works.

freelibre:
Thank you, yes this hack works for me too so I could continue further in the make process but then I get a DB error (see attached logs). By the way yesterday (without the new merge) I was asked my API GCP and today with a clean install no key is asked...I will keep looking if I can find something to help you with.

alexsegura:
@freelibre Looking at your log file, it says the database _already exists_, which explains why you weren't asked for GCP API key

[...]

Issue #141, opened 2017-11-14
Add a make update command


alexsegura:
Useful when pulling changes. It should do at least: 

```
composer install
docker-compose run php bin/console doctrine:migrations:migrate
```

Atala:
It can even `git pull --rebase` ;)

johnnybigoode:
Can I take a shot at this? 

Besides the `make update`, I feel that a `make fresh` would be nice. Idea from this [reply](https://github.com/coopcycle/coopcycle-web/issues/436#issuecomment-437593541). Probably needs a big warning message since it will delete everything. 

Is this a bad idea?

alexsegura:
Yes sure 🙂

PR #1757, opened 2020-09-27
Add make fresh for cleaning up

johnnybigoode:
Related to issues #436 and #141 

Notice that #141 is already fixed with `make migrations-migrate`

But `make fresh` basically helps cleaning up everything docker for a fresh install.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ migrations-diff:
 	@docker-compose exec php php bin/console doctrine:migrations:diff
 
 migrations-migrate:
+	@composer install
 	@docker-compose exec php php bin/console doctrine:migrations:migrate
 	@docker-compose exec php php bin/console doctrine:schema:update --env=test --force --no-interaction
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,3 +31,10 @@ migrations-migrate:
 
 email-preview:
 	@docker-compose exec php php bin/console coopcycle:email:preview > /tmp/coopcycle_email_layout.html && open /tmp/coopcycle_email_layout.html
+
+fresh:
+	-docker ps -a -q | xargs docker stop
+	-docker ps -a -q | xargs docker rm
+	-docker rmi $(docker images -q)
```
alexsegura, line 38: Wouldn't it be simpler to run `docker-compose down`?

johnnybigoode: It didn't even cross my mind that docker-compose would have such a function. 

[Their docs state that](https://docs.docker.com/compose/reference/down/)nNetworks and volumes defined as external are never removed.

I looked over the project and it seems that no volumes are defined as external. I went ahead and fixed the make to docker-compose down, but it seems pointless now.

alexsegura: It's cool. This way the `Makefile` acts as an straightforward interface.

PR #1851, opened 2020-10-21
use CREATE_TASK_LIST_* reducers from the shared library

vladimir-8:
refactoring only, and change the way how shared library is imported to make it easier to replace it with the local version during development


alexsegura:
Thank you @vladimir-8

PR #1875, opened 2020-10-30
Implement Route Optimization Handling

### LINKED ISSUE #1791: Introduce route optimization
After looking around, it looks like [VROOM](http://vroom-project.org/) is the best candidate for this, because it plays nicely with our existing stack (it uses [OSRM](http://project-osrm.org/) under the hood)

Here are some guidelines to implement it. 

Related issues:
- #217 
- #887 
- #4

--- 

### Add `vroom-docker` to `docker-compose.yml`

VROOM is a command-line program, but it's simpler for us to use its API flavour, [vroom-express](https://github.com/VROOM-Project/vroom-express)
There is a Docker image running `vroom-express`, [vroomvrp/vroom-docker](https://hub.docker.com/r/vroomvrp/vroom-docker).
There is an example [docker-compose.yml configuration](https://github.com/VROOM-Project/vroom-docker/blob/c46c4f518c81fb40ac999206577dcbf3a42d67f1/docker-compose.yml) in the repository.
Of courser, `VROOM_ROUTER` should be `osrm`

### Create a custom Symfony Normalizer to convert a `Task` to a `Job`

The VROOM API expects a problem JSON payload, and returns a solution. 
The JSON payload of the problem is described in the [usage](https://github.com/VROOM-Project/vroom-express#usage) section of vroom-express, and also on the [API docs](https://github.com/VROOM-Project/vroom/blob/master/docs/API.md#input) of VROOM. 

The problem payload contains a `jobs` key, which correspond to an array of [Task](https://github.com/coopcycle/coopcycle-web/blob/d6c7dbeef92494f7195d3c642145ec95a966bdcb/src/Entity/Task.php) objects in our system. 

We have our own serialization format, but the VROOM format is different. 
We use [Symfony Normalizer](https://symfony.com/doc/4.4/serializer.html) component, and so we need to create a [custom Normalizer](https://symfony.com/doc/4.4/serializer/custom_normalizer.html)

⚠️ It's not clear in the docs what is the difference between `jobs` and `shipments`, it needs more investigation on how to map data. 

We want to solve the following [Vehicle Routing Problems](https://en.wikipedia.org/wiki/Vehicle_routing_problem#VRP_variants)
- Vehicle Routing Problem with Pickup and Delivery (VRPPD)
- Vehicle Routing Problem with Time Windows (VRPTW)

In our system, a `Task` has a type, than can be either `pickup` or `dropoff`

### Create a `RouteOptimizer` class

Basically, what we want in the end, is a class that takes an array of `Task` objects, and returns it optimized.
A trivial interface would be:

```php
/**
 * @param Task[] $tasks
 * @return Task[]
 */
public function optimize(array $tasks): array;
```

The implementation loops over the Tasks, converts them to a VROOM problem payload, decodes the VROOM solution, and returns the list sorted. 

It will need some dependencies, like a HTTP client to interact with the VROOM API. 
We can create a HTTP client using [CsaGuzzleBundle](https://github.com/coopcycle/coopcycle-web/blob/d6c7dbeef92494f7195d3c642145ec95a966bdcb/app/config/config.yml#L436-L458), and inject it to the constructor of `RouteOptimizer`

andrewcunnin:
Fixes #1791


app/config/config.yml
```diff
@@ -460,6 +460,10 @@ csa_guzzle:
             config:
                 base_uri: "%router.request_context.scheme%://%router.request_context.host%"
                 lazy: true
+        vroom:
+            config:
+                base_uri: "%env(VROOM_HOST)%"
```
alexsegura, line 465: Can you add this variable to the `env.dist` file please?

src/Entity/RoutingProblem.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+namespace AppBundle\Entity;
```
alexsegura, line 3: Could you change the namespace, please? The `AppBundle\Entity` namespace is better suited for classes that are persisted in database. Maybe a `AppBundle\Vroom` package?

alexsegura:
Thank you @andrewcunnin 

I added a button to bring your work to life ⚡️

optimize

PR #1881, opened 2020-11-01
update max-width and width of modal container

agichim:
On mobile browsers the popup is too shrunk.

image

The code update should have it look like this on mobile:
image

And on computers should look like this:
image


amadsalmon (approved):
You are right, it is better in this case to define the width of the popup in percent of the containing block rather than with a pixel-fixed size.
It is better for the user.

PR #1899, opened 2020-11-05
refactor redux store structure

vladimir-8:
- rename dispatch -> lastmile

- refactoring to start using a new redux store structure in the lastmile (aka dispatch) part:

```
{
    lastmile: {
        simpleDomainData1: {....},
        simpleDomainData2: {....},
        entities : {
            entityType1 : {....},
            entityType2 : {....}
        },
        ui : {
            uiSection1 : {....},
            uiSection2 : {....}
        } 
...
    },  
```

- tests


js/app/coopcycle-frontend-js/lastmile/redux.js
```diff
@@ -0,0 +1,5 @@
+// published version
+export * from 'coopcycle-frontend-js/lastmile/redux'
```
alexsegura, line 2: Actually `dispatch` or `logistics` would have been more accurate but that's fine 🙂

alexsegura:
Thank you @vladimir-8 🎉

alexsegura:
@vladimir-8 sorry, but I had to revert the changes (we can restore them later, no worries). 

Dispatchers are reporting problems:

> the orders assigned to ppl are randomly changing order in their list

> not fixed apparently. or something. they are telling me that they put them in order for a REP, but if they refresh the page they are no longer in that order

Also, the following errors occur:

```
Cannot read property '@id' of undefined
Cannot read property 'address' of undefined
Cannot read property 'username' of undefined
```

alexsegura:
It seems to happen when multiple dispatchers are using the dashboard at the same time. 
Some empty item is inserted into the state, and causes the errors.

PR #1968, opened 2020-12-11
refactor redux store structure (fix)

vladimir-8:
- rename dispatch -> logistics

- refactoring to start using a new redux store structure in the logistics (aka dispatch) part:

```
{
    logistics: {
        simpleDomainData1: {....},
        simpleDomainData2: {....},
        entities : {
            entityType1 : {....},
            entityType2 : {....}
        },
        ui : {
            uiSection1 : {....},
            uiSection2 : {....}
        } 
...
    },  
```

- tests


alexsegura:
Thank you @vladimir-8

js/app/dashboard/redux/taskListEntityReducers.js
```diff
@@ -0,0 +1,106 @@
...
+      //     task: '/api/tasks/21',
+      //     position: 0
+      //   },
+      //   {
+      //     task: '/api/tasks/18',
+      //     position: 1
+      //   }
+      // ],
+      let taskCollectionItems = action.taskList.items
+      let itemIds = taskCollectionItems.map(item => item.task)
```
alexsegura, line 68: I think there's a problem here. The previous code was like this:

```es6
case TASK_LIST_UPDATED:

  taskListIndex = _.findIndex(state.taskLists, taskList => taskList['@id'] === action.taskList['@id'])

  if (-1 === taskListIndex) {

    return state
  }

  newTaskLists.splice(taskListIndex, 1, {
    ...state.taskLists[taskListIndex],
    distance: action.taskList.distance,
    duration: action.taskList.duration,
    polyline: action.taskList.polyline,
  })

  return {
    ...state,
    taskLists: newTaskLists,
  }
```
  
i.e it was **NOT** touching the items. 

This event is generated by an async job, which recalculates the distance/duration of the items in the list.
It means that depending on the order of job scheduling, which is unpredictable, we may be receiving an event for a **previous version** of the list (with other tasks). 

I think this is was caused problems with tasks being "visually" unassigned.

js/app/dashboard/redux/taskListEntityReducers.js
```diff
@@ -0,0 +1,106 @@
...
+      let newEntity = taskListUtils.replaceTasksWithIds(action.taskList)
+      let newItems = taskListEntityUtils.addOrReplaceTaskList(state.byId, newEntity)
+
+      return {
+        ...state,
+        byId: newItems,
+      }
+    }
+    case TASK_LIST_UPDATED: {
+      let entityByUsername = taskListEntityUtils.findTaskListByUsername(state.byId, action.taskList['username'])
```
alexsegura, line 51: Problem here. In this case, the task list should be matched by `@id` (not `username`). 

Previous code was like this: 
```es6
taskListIndex = _.findIndex(state.taskLists, taskList => taskList['@id'] === action.taskList['@id'])
```

It should be matched by `@id`, because the `username` of the event may have been generated from a task list for another day. 

It is actually very tricky... each time an address is updated, an async job checks if the address belongs to a task list, and if yes, it recalculates the route for the task list. What can happen, is that someone modifies an address, that belong to a task list from a few days ago. If the same messenger is assigned today, let's say, if we match by username, we get completely wrong data.

Issue #1662, opened 2020-08-24
Add more rows in the example spreadsheet to apply to more use cases


AdrienSICKLO:
The current example spreadsheet doesn't introduce easily pick up and drop off times and dates. This example can be useful in a number of cases, but it is often needed to precise dates and times to be respected for the dispatcher to know.

I don't know if the solution would be multiple example spreasheets or an evolutive one.

vova1vova:
@AdrienSICKLO I've tried to import the example sheet, and I see a different error
<img width="320" alt="Screen Shot 2020-08-26 at 14 41 11" src="https://user-images.githubusercontent.com/1525555/91304830-67ca7700-e7aa-11ea-89ad-68b0e09267b5.png"> 

Do you know if it's related to the bug you mentioned or is there multiple issues there?

PasLoin:
@vova1vova the exemple sheet is not realy well explained ; if you want mixt between street adress and lat long ; you must put latlongdata in adress colomn ; and that's work. 

@AdrienSICKLO for me it's not realy a "bug" ; because you can also use this format of import for recuring task and importing in the date of the delivery in the dashboard. Maybe it's better to have 2 or 3 example sheet depending of the usage. (date and time ; only time , only date range, no time and no date etc ...)

AdrienSICKLO:
@PasLoin
1- do you know if LatLong is used by people? 
2- Ok, let me change the title

PasLoin:
@AdrienSICKLO LatLong **in production No** ; i see in our import data that we use it for one adress (it's a place in a park) but this make not realy sence because in https://github.com/coopcycle/coopcycle-web/blob/master/src/Service/Geocoder.php it make a reverse and return the adress he found. But in our **Sandbox** yes we used LatLong sometimes to make simulation for public or private procurement and import some open data to make it. (because some data have not adress but always LatLong)

alexsegura:
The data of the example sheet comes from [TaskSpreadsheetParser](https://github.com/coopcycle/coopcycle-web/blob/0f8c4e4f459e00a6df2569d09157e855d479129e/src/Spreadsheet/TaskSpreadsheetParser.php#L50-L76)

ERAMANrepartos:
The specific feature requested has been implemented

PR #2064, opened 2021-01-25
Harmonizing task export and import

TheoPatros:
This PR is the follow up of  #681. It allows use of both "technical" and "meaningful" label in header of task import spreadsheet (i.e. "address.streetAddress", "address", "adress.latlng" and "latlong"). 
One must still use either address or geocoordinates, but not both at the same time, though.


alexsegura:
Thank you @TheoPatros 💯 

There was a BC break, we allow setting both the columns for address as text and lat/lng. 
Maybe it's a problem, see #1662

PR #2082, opened 2021-02-02
Refactor redux store structure (this time it's ok)

alexsegura:
Continues #1968

**TODO**

- [x] Handle `DELETE_GROUP_SUCCESS` action (87a55780195eeb22406c3d716d375d86ade26a94)


js/app/dashboard/redux/taskListEntityReducers.js
```diff
@@ -0,0 +1,118 @@
...
+
+      if (matching.length === 0) {
+
+        return state
+      }
+
+      return {
+        ...state,
+        byId: _.mapValues(state.byId, current => {
+          const newTaskList = _.find(matching, o => o['@id'] === current['@id'])
```
vladimir-8, line 85: may be `matchingLists` instead of `matching` and `matchingList` instead of `newTaskList` would make this part a bit easier to understand?

js/app/dashboard/redux/actions.js
```diff
@@ -458,6 +458,16 @@ function closeImportModal() {
   return { type: CLOSE_IMPORT_MODAL }
 }
 
+function updateTask(task) {
+  return function(dispatch, getState) {
+    let date = selectSelectedDate(getState())
+
+    if (isInDateRange(task, date)) {
```
alexsegura, line 465: This means that if the user changes the task date, changes are ignored. 

(see 7ee119308b301a172cdf9fb280bf8cd0240bb99e)

PR #2238, opened 2021-03-27
Farewell FOSUser

alexsegura:


app/config/security.yml
```diff
@@ -16,7 +16,7 @@ security:
     # http://symfony.com/doc/current/book/security.html#where-do-users-come-from-user-providers
     providers:
         fos_userbundle:
```
sdonner-mobiletrend, line 18: doesn't this have to be changed to nucleos as well?

alexsegura: AFAIK, the key here is arbitrary, and doesn't have any meaning. 

https://github.com/symfony/security-bundle/blob/5.3/DependencyInjection/MainConfiguration.php#L341-L352

PR #2298, opened 2021-04-18
Final refactoring of vendor API

### LINKED ISSUE #3783: Redesign relationship between orders & shops
When we introduced hubs, we also introduced the `Vendor` concept. 
Basically, an order is linked with a vendor, which can be either a single shop, or a hub. 

What we want now is to remove this, and resolve the shops based on the order items. 

Take a look at #2298, which contains some code in this direction.

alexsegura:
Fixes #3783


lucasferraro (approved):

lucasferraro:
@alexsegura Why there is no migrations in the PR to drop the Vendor table and its relations? Maybe is just to keep the information and support some backward compatibility?

alexsegura:
> Why there is no migrations in the PR to drop the Vendor table and its relations? Maybe is just to keep the information and support some backward compatibility?

No, it's just that I forgot. Nice catch, let's add it.

PR #2412, opened 2021-05-15
Update wording Planning → Holidays

agichim:
In British English, the term `Holidays` conveys the right meaning instead of `Planning`.


agichim:
In British English, the term `Holidays` conveys the right meaning instead of `Planning`.

alexsegura:
Thank you @agichim

PR #2454, opened 2021-05-26
sticky footer with buttons added in new product window

AniaCuapio:


js/app/product/form.scss
```diff
@@ -0,0 +1,8 @@
+.sticky-buttons{
+  position:  fixed;
+  left: 0;
+  bottom: 0;
+  width: 100%;
+  padding: 0% 2%;
+  box-sizing: border-box;
+  }
```
alexsegura, line 8: Missing new line at end of file

js/app/product/form.scss
```diff
@@ -0,0 +1,8 @@
+.sticky-buttons{
```
alexsegura, line 1: You should actually move this class into `js/widgets/admin.scss`, this way it will be available throughout the whole admin interface.

js/app/user/form.scss
```diff
@@ -0,0 +1,8 @@
+.sticky-buttons{
```
alexsegura, line 1: This is useless, no? It's already there in `js/app/widgets/admin.scss`

alexsegura (changes requested):
Some fields are appearing above the buttons. 

buttons-zindex

alexsegura (changes requested):
On some pages the content at the bottom is not accessible. Some padding is needed. 

buttons-padding

AniaCuapio:
Ok, I can see the issues, I will work on them :)

PR #2826, opened 2021-10-05
Mercadopago gateway

lucasferraro:
Included in this PR

- [x] Changes to support connections with Mercadopago Marketplace
- [x] Store needed fields from Mercadopago responses
- [x] New endpoint to create Mercadopago Preferences


features/orders.feature
```diff
@@ -1261,3 +1261,27 @@ Feature: Orders
         "channel":"coopcycle_order_events#1"
       }
       """
+
+  @debug
```
lucasferraro, line 1265: @alexsegura should we remove this tag?
BTW. The step `And the user "bob" has ordered something at the restaurant with id "1"` is breaking in my local for all the orders tests :thinking:

src/Controller/MercadopagoController.php
```diff
@@ -108,7 +110,7 @@ public function oAuthCallbackAction(
         if (isset($res['error']) && !empty($res['error'])) {
             $this->addFlash(
                 'error',
-                $res['error_description']
+                $res['error']
```
lucasferraro, line 113: When I got some errors doing tests against Mercadopago API I saw that the field `error_description` was not present in the response. I think we should take `error_description` when it is present or `error` when it isn't but I don't know how to make that in PHP :)

src/Action/Order/MercadopagoPreference.php
```diff
@@ -0,0 +1,61 @@
...
+
+        // Create a preference object
+        $preference = new MercadoPago\Preference();
+
+        // Create items for the preference
+        // We add one item for the whole order
+        $items = [];
+        $mpItem = new MercadoPago\Item();
+        $mpItem->title = sprintf('Pedido #%s', $data->getNumber()); // este texto es el que el vendedor puede ver en el detalle de la Operación/Actividad en mercadopago.com.ar/activities/detail/
+        $mpItem->description = $data->getRestaurant()->getName(); // este texto es el que el usuario/comprador puede ver al momento de pagar en el flujo de MP
```
alexsegura, line 38: PHPStan is complaining because properties are protected, however the `MercadoPago\Entity` class indeed has a magic `__set` method. I think this needs to be ignored in `phpstan.neon` file.

src/Controller/MercadopagoController.php
alexsegura: Like this:
```php
$res['error_description'] ?? $res['error']
```

features/orders.feature
alexsegura: > The step And the user "bob" has ordered something at the restaurant with id "1" is breaking in my local

We need to investigate this, no idea why.

src/Action/Order/MercadopagoPreference.php
lucasferraro: I think this should be fixed [here](https://github.com/coopcycle/coopcycle-web/pull/2826/commits/f3e8bf5fb52d8b5474ac2f55d028dedf3e06e1ff?authenticity_token=VVTnGagoDcjeoGxyNtTve1tv70wLqkg35F09rw8tiPmOthygOp9AdW5u%2BpafOO9Q0YbZyKLxHS6Kh84xjqgYyA%3D%3D&file-filters%5B%5D=.neon&file-filters%5B%5D=.php#diff-0361f0c81f363476ddc6f44ab36fcbe66ee685d5f4c2a46b054924591544b766)

PR #2831, opened 2021-10-06
Fix 2531 opening automatically accordion

AniaCuapio:


js/app/dashboard/components/TaskList.js
```diff
@@ -73,6 +73,10 @@ class TaskList extends React.Component {
     this.props.unassignTasks(this.props.username, task)
   }
 
+  componentDidUpdate(){
+    console.log('componentDidUpdate')
```
alexsegura, line 77: Ok so here the goal is to clear `preExpanded` (i.e reset it to an empty array). 

1 - Add a new Redux action specifically for this (for example, `CLEAR_PRE_EXPANDED`)
2 - Call this Redux action in `componentDidUpdate`
3 - In the reducer function, when this action is triggered, it should clear the `preExpanded` array
```js
return {
  ...state,
  preExpanded: [],
}
```

PR #2877, opened 2021-11-02
Mercadopago fixes

lucasferraro:
For testing payment in mobile app we need to use the live access token setting, becuase for that test we are using test users with their live credentiales (instead of real user with test credentials).


src/Service/SettingsManager.php
```diff
@@ -104,6 +104,10 @@ public function get($name)
             case 'mercadopago_access_token':
                 $name = $this->isMercadopagoLivemode() ? 'mercadopago_live_access_token' : 'mercadopago_test_access_token';
                 break;
+            case 'mercadopago_access_token_for_test':
+                // to be used only for test
+                $name = !$this->isMercadopagoLivemode() ? 'mercadopago_live_access_token' : '';
```
alexsegura, line 109: Does this work? I mean trying to retrieve a setting with an empty (`''`) name will return `null`, right?

templates/_partials/restaurant/mercadopago_accounts.html.twig
```diff
@@ -7,7 +7,7 @@
   <tr>
     <td>
       {% if mercadopago_account is not empty %}
-        <span class="label label-success">Connected</span>
+        <span class="label label-success">{{ 'form.local_business.mercadopago.connected'|trans }}</span>
```
alexsegura, line 10: Nice! 🙂

src/Service/SettingsManager.php
lucasferraro: @alexsegura  It should never retrieve an empty string because the setting `mercadopago_access_token_for_test` should only be requested when Mercadopago isn't in live mode. I just added this conditional to be clear about to use this key only for test purpose. Let me know if it does not make sense for you.

alexsegura: It's fine.

PR #2890, opened 2021-11-10
Product images in restaurant menus response and quantity when it's added to cart

lucasferraro:
Two changes that we need in API
- [x] When add an item to the cart should be possible to specify the quantity of that item
- [x] On Restaurant Menus we have to return in products informations the images of them


src/Api/Dto/CartItemInput.php
```diff
@@ -14,7 +14,7 @@ final class CartItemInput
     /**
      * @Groups({"cart"})
      */
-    public $quantity = 1;
+    public $quantity;
```
alexsegura, line 17: I'm afraid this can break current versions of the app that are not sending the quantity yet.

alexsegura: We should add a functional test for this. To check if it works **both** when sending the quantity, or not (legacy).

lucasferraro: You're right, I'll check how to make it work for both.

alexsegura: I think you can just let it like this, no? With the default value I mean.

lucasferraro: Yes thanks, if I let it with the default value it works. I thought that it will always put 1 as quantity but not, it's just a default value.
Reverted :)

src/Serializer/JsonLd/RestaurantMenuNormalizer.php
```diff
@@ -109,6 +115,24 @@ private function normalizeOptionValues(ProductOptionInterface $option, $optionVa
         return $data;
     }
 
+    private function getProductImagePath($product, $ratio) {
+        try {
+            $product_image = $product->getImages()->filter(function ($image) use($ratio) {
```
alexsegura, line 120: Better written in camel case, i.e `$productImage`

src/Serializer/JsonLd/RestaurantMenuNormalizer.php
```diff
@@ -169,6 +193,15 @@ public function normalize($object, $format = null, array $context = array())
                         }, $allergens));
                     }
 
+                    $image_1x1 = $this->getProductImagePath($product, '1:1');
+                    if ( $image_1x1 ) {
+                        $item['image_1x1'] = $image_1x1;
+                    }
+                    $image_16x9 = $this->getProductImagePath($product, '16:9');
+                    if ( $image_16x9 ) {
+                        $item['image_16x9'] = $image_16x9;
```
alexsegura, line 202: IMHO all images should be under an `images` property, in an array

```json
{
  "images": [
    {
      "ratio": "16:9",
      "url": "https://example.com/image.jpg"
    }
  ]
}
```

src/Entity/Sylius/Order.php
```diff
@@ -1000,9 +1000,10 @@ public function getFulfillmentMethod(): string
     public function getPaymentMethod(): string
     {
         $payment = $this->getLastPayment();
+        $method = $payment->getMethod();
```
alexsegura, line 1003: I had to revert this change, because sometimes `$payment` can be null

lucasferraro: Oh ok, sorry. I made this change because some tests are failing because the methods are null and then when try to acccess getCode() method an exception is raised.

src/Serializer/JsonLd/RestaurantMenuNormalizer.php
lucasferraro: I agree, nice improvement :)

src/Entity/Sylius/Order.php
alexsegura: Ok thanks I think it's more a problem of test data not properly initialized

PR #2948, opened 2021-12-06
Guest checkout from mobile app

lucasferraro:
- [x] Allow to assign to a guest customer an order
- [x] Added new security rules to allow acces to endpoints with the temp session token for guests
- [x] New controller to get order details from API

### DIFF
--- a/features/orders.feature
+++ b/features/orders.feature
@@ -1199,6 +1199,142 @@ Feature: Orders
       }
       """
 
+  Scenario: Get cart payment methods for guest
+    Given the fixtures files are loaded:
+      | sylius_channels.yml |
+      | products.yml        |
+      | restaurants.yml     |
+    And the restaurant with id "1" has products:
+      | code      |
+      | PIZZA     |
+      | HAMBURGER |
+    And the setting "brand_name" has value "CoopCycle"
+    And the setting "default_tax_category" has value "tva_livraison"
+    And the setting "subject_to_vat" has value "1"
+    And the setting "stripe_test_publishable_key" has value "pk_1234567890"
+    And the setting "stripe_test_secret_key" has value "sk_1234567890"
+    And the setting "stripe_test_connect_client_id" has value "ca_1234567890"
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I add "Accept" header equal to "application/ld+json"
+    And I send a "POST" request to "/api/carts/session" with body:
+      """
+      {
+        "restaurant": "/api/restaurants/1"
+      }
+      """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the JSON should match:
+      """
+      {
+        "token":@string@,
+        "cart":{
+          "@context":"/api/contexts/Order",
+          "@id":"/api/orders/1",
+          "@type":"http://schema.org/Order",
+          "customer":null,
+          "restaurant":"/api/restaurants/1",
+          "shippingAddress":null,
+          "shippedAt":null,
+          "shippingTimeRange": null,
+          "reusablePackagingEnabled":false,
+          "reusablePackagingPledgeReturn": 0,
+          "notes":null,
+          "items":[],
+          "itemsTotal":0,
+          "total":0,
+          "adjustments":@...@,
+          "fulfillmentMethod":"delivery"
+        }
+      }
+      """
+    Given the client is authenticated with last response token
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I add "Accept" header equal to "application/ld+json"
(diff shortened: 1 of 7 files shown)


src/Action/Order/Assign.php
```diff
@@ -35,6 +47,42 @@ public function __invoke($data)
             $data->setCustomer($user->getCustomer());
         }
 
+        $body = [];
+        $content = $request->getContent();
+        if (!empty($content)) {
+            $body = json_decode($content, true);
+        }
+
+        if (isset($body['guest']) && true === $body['guest']) {
```
alexsegura, line 56: I guess this is already an issue, but maybe we should check if the order is in an expected state here? The state should be `cart`.

src/Controller/OrderController.php
```diff
@@ -576,4 +577,34 @@ public function continueAction(Request $request,
 
         return $this->redirectToRoute('restaurant', ['id' => $restaurants->first()->getId()]);
     }
+
+    /**
+     * @Route("/order/{hashid}/preview", name="order_preview")
+     */
+    public function dataPreviewAction($hashid, OrderRepository $orderRepository)
```
alexsegura, line 584: Maybe this could be moved to being managed by API Platform? 
This would need a [Custom Item Data Provider](https://api-platform.com/docs/core/data-providers/#custom-item-data-provider) to load an order via its hashid.

features/orders.feature
```diff
@@ -1272,25 +1408,3 @@ Feature: Orders
       }
       """
 
-    Scenario: Get order timing
```
alexsegura, line 1275: Why is it removed?

src/Controller/OrderController.php
lucasferraro: Thanks for suggestion. Yes I agree about to move this and be managed by API Platform.
But I don't fully understand how Customer Item Data Provider works.
I've created a data provider but it tries to get an order item  if a do a GET to `api/orders/LnOpkxEN6xE5M7XR` or `api/orders/51` (and I guess we just want the data provider to manage the hash id case, right?).
BTW, when I send a hash id like here  `api/orders/LnOpkxEN6xE5M7XR` in the data provider the param value is 0 instead of the hash. But if I send an ID like '51' the param is 51.

This is what I have:
```
   final class OrderDataProvider implements ItemDataProviderInterface, RestrictedDataProviderInterface
   {
      public function __construct(string $secret)
      {
          $this->hashids = new Hashids($secret, 16);
      }
  
      public function supports(string $resourceClass, string $operationName = null, array $context = []): bool
      {
          return Order::class === $resourceClass;
      }
  
      public function getItem(string $resourceClass, $hashid, string $operationName = null, array $context = []): ?Order
      {
          $decoded = $this->hashids->decode($hashid);
  
          if (count($decoded) !== 1) {
              throw new BadRequestHttpException(sprintf('Hashid "%s" could not be decoded', $hashid));
          }
  
          $id = current($decoded);
          return new Order($id);
      }
  }
```

At `services.yml`
```
AppBundle\Api\DataProvider\OrderDataProvider:
    arguments:
      $secret: '%secret%'
    tags:
      - { name: 'api_platform.item_data_provider', priority: 2 }
    autoconfigure: false
```

features/orders.feature
lucasferraro: This test is not actually testing anything. We've added this test at the begining of work with mercadopago to get the response from preferences endpoint. But now id doesn't work because we would need to set an Access Token to get a preference and we don't have that for now.

alexsegura: Ok 👍

src/Controller/OrderController.php
alexsegura: No problem, I never implemented a Custom Item Data Provider in this use case.

alexsegura (approved):

src/Action/Order/Assign.php
```diff
@@ -6,15 +6,27 @@
...
+        )
     {
         $this->tokenStorage = $tokenStorage;
+        $this->customerRepository = $customerRepository;
+        $this->canonicalizer = $canonicalizer;
+        $this->customerFactory = $customerFactory;
     }
 
-    public function __invoke($data)
+    public function __invoke($data, Request $request)
```
alexsegura, line 29: Something I forgot here, is that it should check if the instance has enabled guest checkout 😕

alexsegura: Fixed in 28bde80c2d44904fa527aca869fa8747390d6e4e

lucasferraro: Great, thanks!

PR #2968, opened 2021-12-15
Move packages at task level

alexsegura:
Now packages will be attached to **dropoff** tasks, instead of delivery. 

**TODO**

- [x] Make it work when adding multiple dropoffs (has error "This form should not contain extra fields.")
- [x] The `Delivery::getPackages()` method should compute an aggregated collection of packages when there are multiple dropoffs. For ex, if a delivery contains 3 tasks, and 2 of them have 1 package of type "XL", `Delivery::getPackages()` should return 2 XL packages. 

---

The `Delivery` class still exposes the same interface, but the implementation will now attach packages to the first dropoff task. 

```php
// Both are equivalent
$delivery->addPackageWithQuantity($small, 1);
$delivery->getDropoff()->addPackageWithQuantity($small, 1);
```

Same for the API, clients can still send `packages` at the delivery level, but they will be attached to the first dropoff task. 

```
POST /api/deliveries
{
  "packages": [
    {"type": "XL", "quantity": 2}
  ]
}
```

```
POST /api/deliveries
{
  "dropoff": {
    "packages": [
      {"type": "XL", "quantity": 2}
    }
  ]
}
```

--- 

task_package

### DIFF
--- a/app/DoctrineMigrations/Version20211215104949.php
+++ b/app/DoctrineMigrations/Version20211215104949.php
@@ -1,0 +1,96 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20211215104949 extends AbstractMigration
+{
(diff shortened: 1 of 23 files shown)


AdrienSICKLO:
Looks good! Same thing to do with weight and we'll be ok on this imho

PR #2988, opened 2021-12-23
Closed restaurants improvements

lucasferraro:
Changes in WEB and API to improve how we show and handle the different posibilites of a closed restaurant.

UI changes for WEB:

- **(1)** List of restaurants, when a restaurant is closed we show an icon, the label 'closed now' and the background has less opacity. In this case we have these restaurants:
  - La isla oriental: opened
  - La berenjena: closed but opens today later
  - La silla caliente: closed but opens tomorrow
  - La sarten que fuma: closed and opens another day that is not today or tomorrow
  - La patata caliente: closed with a closing rule, not available

- **(2)** Restaurant page, when it's open the button label is "Order" because it's available to receive orders:

- **(3)** Restaurant page, when it's closed but opens today:

- **(4)** Restaurant page, when it's closed but opens tomorrow and is available "pre order" option:

- **(5)** Restaurant page, when it's closed but opens tomorrow and is not available "pre order" option:

- **(6)** Restaurant page, when it's closed but opens other day which is not today or tomorrow  and is available "pre order" option:

- **(7)** Restaurant page, when it's closed but opens other day which is not today or tomorrow  and is not available "pre order" option:

- **(8)** Restaurant page, when it's closed and won't be open until finish a configured closing rule:

### DIFF
--- a/features/foodtech.feature
+++ b/features/foodtech.feature
@@ -586,6 +586,8 @@ Feature: Food Tech
             "validThrough":"2020-10-03"
           }
         ],
-        "image":@string@
+        "image":@string@,
+        "isOpen":false,
+        "nextOpeningDate":@string@
       }
       """
--- a/features/orders.feature
+++ b/features/orders.feature
@@ -566,7 +566,7 @@ Feature: Orders
     """
 
   Scenario: Create order without shipping date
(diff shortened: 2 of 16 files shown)


alexsegura (approved):
Looks Good To Me 🙂

PR #3015, opened 2022-01-05
set restaurant public key

lucasferraro:
We are sending the public key of the market place owner and the test in production has failed. 
I guess because we have to send the public key of the restaurant instead.

### DIFF
--- a/js/app/payment/mercadopago.js
+++ b/js/app/payment/mercadopago.js
@@ -121,17 +121,15 @@ const MercadoPagoForm = ({ amount, onChange }) => {
 export default ({ onChange }) => ({
   init(form) {
     this.form = form
-    const { country, restaurantPublicKey } = this.config.gatewayConfig
+    const { country, publishableKey } = this.config.gatewayConfig
 
-    if (!restaurantPublicKey) {
-      throw new Error('Current restaurant has not configured the Public Key for Mercadopago')
+    if (!publishableKey) {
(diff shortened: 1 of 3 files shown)


lucasferraro:
I convert this PR to draft beacuse is better if we first merge #3027.
After merge #3027 I need to make some extra changes on this PR to support MercadoPago in forms.

alexsegura:
@lucasferraro #3027 is now merged 😉

lucasferraro:
@alexsegura Now it's ready :)
When this PR is merged we need it deployed on TRU instance so we can test Mercadopago in prod.
Thanks!

js/app/payment/mercadopago.js
```diff
@@ -123,11 +123,10 @@ export default ({ onChange }) => ({
     this.form = form
     const { country, publishableKey } = this.config.gatewayConfig
 
-    /*
-     * When this flow is tested in a production environment we have to check if the Public Key to send
-     * has to be from the Marketplace owner or from the Seller/Restaurant.
-     * Now we are using the Public Key of the Marketplace owner (Coop).
-     */
+    if (!publishableKey) {
+      throw new Error('You need a Public Key for Mercadopago')
```
alexsegura, line 127: This means it won't work for delivery form, right?

alexsegura: Ok my bad. Perfect 🙂

PR #3091, opened 2022-02-01
Paygreen

### LINKED ISSUE #3042: Implement PayGreen - WEB
What we want is to be able to use Stripe **OR** PayGreen. 
For each restaurant, we need to add a property that allows to choose between Stripe and PayGreen. 

- [x] Allow storing PayGreen app id & private key in platform settings
- [x] Allow connecting each restaurant with PayGreen
- [ ] Change `GatewayResolver::resolve` to pass the order as parameter, and decide which gateway to use based on the order context (if the restaurant of the order should use PayGreen, use it)
- [ ] Allow choosing between Stripe & Paygreen at restaurant level

See libs 
- https://paygreen.github.io/paygreen-ui-doc/docs/
- https://github.com/PayGreen/paygreen-php

See examples 
- https://github.com/RenaudGreen/PayGreenPrestashop
- https://github.com/PayGreen/sylius-paygreen-plugin

See docs 
- https://github.com/PayGreen/paygreen-php/tree/master/docs/v2
- https://paygreen.fr/documentation/api-documentation-categorie

r0xsh:
Fixes #3042

### DIFF
--- a/src/Controller/AdminController.php
+++ b/src/Controller/AdminController.php
@@ -53,7 +53,7 @@ use AppBundle\Form\PackageSetType;
 use AppBundle\Form\PricingRuleSetType;
 use AppBundle\Form\RestaurantAdminType;
 use AppBundle\Form\SettingsType;
-use AppBundle\Form\StripeLivemodeType;
+use AppBundle\Form\PaygreenLivemodeType;
 use AppBundle\Form\Type\TimeSlotChoiceType;
 use AppBundle\Form\Sylius\Promotion\CreditNoteType;
 use AppBundle\Form\TimeSlotType;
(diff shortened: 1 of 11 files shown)


src/Controller/AdminController.php
```diff
@@ -53,7 +53,7 @@
 use AppBundle\Form\PricingRuleSetType;
 use AppBundle\Form\RestaurantAdminType;
 use AppBundle\Form\SettingsType;
-use AppBundle\Form\StripeLivemodeType;
```
europrimus, line 56: in the issue #3042 it's say
> use Stripe **OR** PayGreen

if replacing Stripe by PayGreen, Stripe no more available.

PR #3112, opened 2022-02-10
Packages and weight info in pickup task

lucasferraro:
With these changes we are sending in the API response (tasks and delivery endpoints) packages and weight information with the following structure:

**Pickup Task**

```javascript
{
  type: "PICKUP",
  ...
  packages: [
    {
      quantity: 2,
      packages: {name: "XL", volumeUnits: 2}
    },
    {
      quantity: 1,
      packages: {name: "M", volumeUnits: 1}
    }
  ],
  weight: 5000
}
```

**Dropoff tasks**

```javascript
{
  type: "DROPOFF",
  ...
  packages: [
    {
      quantity: 1,
      packages: {name: "M", volumeUnits: 1}
    }
  ],
  weight: 1000
},
{
  type: "DROPOFF",
  ...
  packages: [
    {
      quantity: 2,
      packages: {name: "XL", volumeUnits: 2}
    }
  ],
  weight: 4000
}
```

I'm not sure if we should remove from pickup task comments the packages and weight information. What do you think @alexsegura ?
Because if some user uses this information in the app, maybe is better to keep it so we can ensure backward compatibility. But if it's something that only is used in the web side I think is better to remove it. Let me know.

### DIFF
--- a/features/deliveries.feature
+++ b/features/deliveries.feature
@@ -230,7 +230,7 @@ Feature: Deliveries
           "after":"@string@.isDateTime()",
           "doneBefore":"@string@.isDateTime()",
           "before":"@string@.isDateTime()",
-          "comments": "2.00 kg",
+          "comments": "",
           "weight": 2000,
           "packages": []
         },
@@ -335,7 +335,7 @@ Feature: Deliveries
           "after":"@string@.isDateTime()",
           "doneBefore":"@string@.isDateTime()",
           "before":"@string@.isDateTime()",
-          "comments": "2 × XL\n6.00 kg",
+          "comments": "",
           "weight": 6000,
           "packages": [
             {
--- a/features/deliveries.feature
+++ b/features/deliveries.feature
@@ -128,7 +128,9 @@ Feature: Deliveries
           "after":"@string@.isDateTime()",
           "doneBefore":"@string@.isDateTime()",
           "before":"@string@.isDateTime()",
-          "comments": ""
+          "comments": "",
+          "weight": null,
+          "packages": []
         },
         "dropoff":{
           "@id":"@string@.startsWith('/api/tasks')",
@@ -153,7 +155,8 @@ Feature: Deliveries
           "doneBefore":"@string@.isDateTime()",
           "before":"@string@.isDateTime()",
           "comments": "Beware of the dog\nShe bites",
-          "weight":null
+          "weight":null,
+          "packages": []
         }
       }
       """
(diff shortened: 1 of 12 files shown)


src/Entity/Package.php
```diff
@@ -3,13 +3,20 @@
...
 {
     use Timestampable;
 
     protected $id;
+    /**
+     * @Groups({"task", "delivery"})
+     */
     protected $name;
+    /**
+     * @Groups({"task", "delivery"})
```
alexsegura, line 18: At the moment it's not useful to serialize the `volumeUnits` property. It's something that is yet pretty unused (except very recently in #3118), it was introduced to be used in a near future for [route optimization using VROOM](https://github.com/VROOM-Project/vroom/blob/master/docs/API.md#shipments). 

But in any case even if it was used for route optimization, API consumers wouldn't care that much about the volume units of each packages, it's more for internal purpose.

features/deliveries.feature
```diff
@@ -252,7 +257,137 @@ Feature: Deliveries
...
+          "after":"@string@.isDateTime()",
+          "doneBefore":"@string@.isDateTime()",
+          "before":"@string@.isDateTime()",
+          "comments": "Beware of the dog\nShe bites",
+          "weight": 6000,
+          "packages": [
+            {
+              "@type": "Package",
+              "@id": @string@,
+              "package": {
```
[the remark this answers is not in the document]
lucasferraro: @alexsegura Fixed.
Now we are returning this structure for pickup and dropoff packages:
```json
"packages": [
  {
    "type": "XL",
    "name": "XL",
    "quantity": 2
  }
]
```
For now, `type` field is the same as `name`. Is that ok?

src/Entity/Package.php
lucasferraro: Ok, I just removed `volumeUnits` from the response.

features/deliveries.feature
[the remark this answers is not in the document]
alexsegura: Perfect

src/Entity/Package.php
```diff
@@ -3,12 +3,16 @@
...
 use Gedmo\Timestampable\Traits\Timestampable;
+use Symfony\Component\Serializer\Annotation\Groups;
 
 class Package
 {
     use Timestampable;
 
     protected $id;
+    /**
+     * @Groups({"task", "delivery"})
```
alexsegura, line 14: The annotations can be removed as its always overridden in `TaskNormalizer`

alexsegura (approved):

PR #3127, opened 2022-02-16
Improve management of deliveries with multiple dropoffs.

alexsegura:
Following #3112, I realised that we didn't have a functional test that checks that the pickup task contains an aggregation of the packages and weight of the subsequent dropoffs. 

This pull request adds a test for that, and also allows to send a payload with multiple dropoffs.

### DIFF
--- a/features/deliveries_multi.feature
+++ b/features/deliveries_multi.feature
@@ -90,3 +90,118 @@ Feature: Multi-step deliveries
         }
       }
       """
+
+  Scenario: Create delivery with pickup & dropoff + packages with OAuth
+    Given the fixtures files are loaded:
+      | sylius_channels.yml |
+      | stores.yml          |
+    Given the setting "latlng" has value "48.856613,2.352222"
+    And the store with name "Acme" has an OAuth client named "Acme"
+    And the OAuth client with name "Acme" has an access token
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I add "Accept" header equal to "application/ld+json"
+    And the OAuth client "Acme" sends a "POST" request to "/api/deliveries" with body:
+      """
+      {
+        "tasks": [
+          {
+            "type": "pickup",
+            "address": "24, Rue de la Paix",
+            "before": "tomorrow 13:00"
+          },
+          {
+            "type": "dropoff",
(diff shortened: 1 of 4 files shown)


src/Entity/Delivery.php
```diff
@@ -298,6 +288,25 @@ public static function createWithTasks(Task ...$tasks)
...
+
+        $pickup = array_shift($tasks);
+
+        // Make sure the first task is a pickup
+        $pickup->setType(Task::TYPE_PICKUP);
+
+        $this->addTask($pickup);
+
+        foreach ($tasks as $dropoff) {
+            $dropoff->setPrevious($pickup);
```
lucasferraro, line 303: How does it work this `previous` concept? 
If I have 3 tasks (1 pickup and 2 dropoffs) in my mind I think that:
- first task is pickup, doesn't have a previous
- second task, its previous is the pickup task and the next the other dropoff
- third task, its previous is the second dropoff and doesn't have a next

but here for both dropdoff tasks the previous it's always the pickup :thinking:

lucasferraro (approved):

features/deliveries_multi.feature
```diff
@@ -90,3 +90,118 @@ Feature: Multi-step deliveries
...
+          "packages":[
+            {
+              "type":"XL",
+              "name":"XL",
+              "quantity":2
+            }
+          ]
+        }
+      }
+      """
```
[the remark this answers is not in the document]
alexsegura: It's because a fix was needed in `DeliveryNormalizer`. It only supported 2 tasks.

src/Entity/Delivery.php
alexsegura: In the case of a delivery with multiple tasks (a pickup + 2 dropoffs like you say), we don't necessarily know the order of dropofs. It's because we can't trust clients/users to be sending data that is already optimised. 

When completing a task, there is a piece code that checks if [the previous task has been completed](https://github.com/coopcycle/coopcycle-web/blob/315f848a4607a091d237390e0cc87c2340f1f2c6/src/Domain/Task/Handler/MarkAsDoneHandler.php#L36-L38). The pickup will always need to be completed first, but the dropoffs can be completed in any order.

Issue #3010, opened 2022-01-05
Implement filtering (tags, categories) on restaurant list


alexsegura:
Basically, we should add a sidebar that allows filtering between cuisines, and/or tags.

alexsegura:
### Long term vision

What we want to implement is "faceted search" (https://en.wikipedia.org/wiki/Faceted_search)

Right now, filtering will be implemented by performing SQL queries in the database.
This may lead to bad performance, and put some pressure on the database.  
We can mitigate the performance impact by caching results. 

To reach a higher throughput, we will need to use a search/indexing engine. 
So, filtering system (mainly URLs) needs to be designed so that we can migrate later. 

This will be implemented on the web first, but as we will want to implement it on the app, the filtering will probably need to be moved behind the API, to make it reusable out of the box on the app. 

#### 1/ Implement horizontal scroll for "collections" (homepage)

This will be implemented directly on the (current) homepage, to show more shops (less vertical space). It is basically what already exists on the homepage: we have a collection per type of business.  

At the moment, collections will be built automatically on what we have. 
We will have a collection per type of business like now, and we can also have: 
- Featured shops
- Exclusive shops
- New shops
- Some collections per cuisine (the ones with the most restaurants)

Later, we can allow "editing" collections by hand, creating custom collections with arbitrary shops. 

#### 2/ Implement lazy-loading for images

If we are going to show more shops, at least we must lazy load images for performance. 

#### 3/ Implement infinite vertical scroll

It shouldn't be necessary to click on a button to show more shops. 
As most visitors are coming on mobile, this improves UX a lot when scrolling with fingers. 

#### 4/ Implement sidebar on shops list page

At the moment, we will have 2 sections (= facets) on the sidebar: type of business, and cuisine. 

:warning: Existing URLs should continue working, and redirect (?) to new URLs. 

- /fr/restaurants
- /fr/shops?type=fast-food-restaurant
- /fr/restaurants/cuisines/noodle_bar

#### 5/ Start indexing shops using Typesense

We are already using Typesense, and it implements faceted search. 
For some time, we won't be using it, just monitoring the RAM usage across all instances. 

https://typesense.org/docs/0.22.2/api/documents.html#facet-results

#### 6/ Implement search box + use Typesense to return filtered results

Once everything is indexed properly in Typesense, we can use the search engine, instead of performing SQL queries.  

#### 7/ Allow adding custom facets to shops

[...]

PR #3176, opened 2022-03-18
New side bar in shops list page

lucasferraro:
### Fourth step for issue https://github.com/coopcycle/coopcycle-web/issues/3010

**Implement sidebar on shops list page**
At the moment, we will have 2 sections (= facets) on the sidebar: type of business, and cuisine.

:warning: Existing URLs should continue working, and redirect (?) to new URLs.

/fr/restaurants
/fr/shops?type=fast-food-restaurant
/fr/restaurants/cuisines/noodle_bar

### How it looks with this PR

**Web**
https://user-images.githubusercontent.com/4819244/160000457-5bb92116-da85-4841-869d-ba7a6ddeca1b.mp4

**Mobile**
https://user-images.githubusercontent.com/4819244/159996834-20520b11-9b4a-4a77-98d0-8affeb23f46e.mp4

**Filtered shops when click 'View all' in homepage**
https://user-images.githubusercontent.com/4819244/159996909-d3c86e7c-f3f5-432b-81f1-ee3d9ea7f32a.mp4


alexsegura:
Actually, we have another facet for filtering that is already there: zero waste restaurants

ERAMANrepartos:
First impressions, fucking awesome! I love it!

Things that occur to me:
1. in mobile, that symbol for filters isnt well known here where I am. (see photo)
image
Potentially better would be horizontal rectangular buttons with text in them, both making it clearer what they are, more obvious that they exist, and more customizable because of translation of text rather than symbols. such as...
image

alexsegura:
It's not directly related with this issue but honestly, I think we can get rid of the map view. It was there since the beginning when most instance didn't have that much restaurants/shops to show, but now we can remove it, I don't think much people are using it.

src/Controller/RestaurantController.php
```diff
@@ -229,16 +262,30 @@ public function listAction(Request $request,
...
         } else {
 
-            $cacheKey = sprintf('restaurant.list.ids|%s', LocalBusiness::getKeyForType($type));
+            $cacheKey = $this->getShopsListCacheKey($request, $type);
 
-            $restaurantsIds = $projectCache->get($cacheKey, function (ItemInterface $item) use ($repository, $type) {
+            // for filtering we need in query param the full type instead of the key
+            $request->query->set('type', $type);
+
+            if ($request->query->has('cuisine')) {
```
alexsegura, line 270: Instead of using comma-separated values in the params, we should make usage of native PHP management of arrays. The URLs would look like `?cuisine[]=indian&cuisine[]=asian`. Then `$request->query->get('cuisine')` would directly return an array.

src/Entity/LocalBusinessRepository.php
```diff
@@ -264,6 +266,73 @@ public function findByCuisine($cuisine) {
...
+                            ->setParameter('cuisineIds', $value);
+                        break;
+                    case 'category':
+                        switch($value) {
+                            case 'featured':
+                                $qb
+                                    ->andWhere('r.featured = :featured')
+                                    ->setParameter('featured', true);
+                                break;
+                            case 'exclusives':
```
alexsegura, line 305: Should be `exclusive` (singular)

src/Entity/LocalBusinessRepository.php
```diff
@@ -264,6 +266,73 @@ public function findByCuisine($cuisine) {
...
+                                $qb
+                                    ->andWhere('r.exclusive = :exclusive')
+                                    ->setParameter('exclusive', true);
+                                break;
+                            case 'news':
```
alexsegura, line 310: Should be `new` (singular)

ERAMANrepartos:
I agree with mex about map view. It´s a cool feature to have generally speaking (when I was in Madrid I used it to find a place to eat based on La Pajara´s restaurants) but its not so key it needs to be everywhere.

src/Controller/IndexController.php
```diff
@@ -128,7 +130,9 @@ public function indexAction(LocalBusinessRepository $repository, CacheInterface
             $sections[] = [
                 'title' => $translator->trans('homepage.exclusives'),
                 'shops' => iterator_to_array($exclusivesIterator),
-                'show_more' => false,
+                'view_all_path' => $urlGenerator->generate('shops', [
+                    'category' => 'exclusives',
```
alexsegura, line 134: Should be `exclusive` (singular)

templates/restaurant/_partials/mobile_filters.html.twig
```diff
@@ -0,0 +1,84 @@
...
+        <button class="filter-section-btn btn btn-sm shadow-none bg-transparent w-100"
+            data-toggle="collapse" href="#collapseCategoryMobile" aria-expanded="true" aria-controls="collapseCategoryMobile">
+            <div class="d-flex flex-row align-items-center justify-content-between">
+                <h4><strong>{{ 'shops.filters.title.categories'|trans }}</strong></h4>
+                <strong><span class="filter-caret caret"></span></i></strong>
+            </div>
+        </button>
+        {% set categories = [
+            {key: 'featured', transKey: 'homepage.featured'},
+            {key: 'exclusives', transKey: 'homepage.exclusives'},
```
alexsegura, line 66: Should be `exclusive` (singular)

templates/restaurant/list.html.twig
```diff
@@ -2,26 +2,116 @@
...
+        <button class="filter-section-btn btn btn-sm shadow-none bg-transparent w-100"
+          data-toggle="collapse" href="#collapseCategory" aria-expanded="true" aria-controls="collapseCategory">
+          <div class="d-flex flex-row align-items-center justify-content-between">
+            <h4><strong>{{ 'shops.filters.title.categories'|trans }}</strong></h4>
+            <strong><span class="filter-caret caret"></span></i></strong>
+          </div>
+        </button>
+        {% set categories = [
+          {key: 'featured', transKey: 'homepage.featured'},
+          {key: 'exclusives', transKey: 'homepage.exclusives'},
```
alexsegura, line 72: Should be `exclusive` (singular)

alexsegura:
@lucasferraro sorry, I changed `exclusives` to `exclusive` in translations to simplify (7ac3b80da3654fc046d3205a9dca175a89c71ce3), you may need to rebase/merge

src/Controller/RestaurantController.php
lucasferraro: I've changed to use arrays instead of comma-separated string ([commit](https://github.com/coopcycle/coopcycle-web/pull/3176/commits/0da0eff26896155ae397cd1dc18e4fa4170084ee)) 
Something that I can't figure out how to change is that the array has the index for each value and the URL looks like this:
`/shops?cuisine[0]=asian&cuisine[1]=indian`

lucasferraro:
> We can't keep this in `RestaurantController::listAction`
> 
> ```
> if ($request->query->has('geohash') && strlen($request->query->get('geohash')) > 0) {
>    // Filter restaurants by address
> } else {
>   // Filter restaurants by facets
> }
> ```
> 
> Because it won't work for the flow where the visitor enters the address first on the homepage (see GIF below): **it will never apply the filters**.
> 
> The algorithm should be like this:
> 
> * Apply the facets filters, if any
> * Apply the lat/lng filter, if the `geohash` parameter is present
> 
> You can use `AppBundle\Utils\RestaurantFilter` to achieve this:
> 
> ```
> $matches = ... // Retrieve restaurants matching filters like you are doing
> 
> if ($request->query->has('geohash') && strlen($request->query->get('geohash')) > 0) {
>     $geotools = new Geotools();
>     $geohash = $request->query->get('geohash');
> 
>     $decoded = $geotools->geohash()->decode($geohash);
> 
>     $latitude = $decoded->getCoordinate()->getLatitude();
>     $longitude = $decoded->getCoordinate()->getLongitude();
> 
>     $matches = $restaurantFilter->matchingLatLng($matches, $latitude, $longitude);
> }
> ```
> 
> search_addr_filters

Thanks for catching this! :raised_hands: 
Fixed here -> https://github.com/coopcycle/coopcycle-web/pull/3176/commits/3e0cb8fdd6da687a2341ef4c678fa31bc4dbba07

lucasferraro:
> It's not directly related with this issue but honestly, I think we can get rid of the map view. It was there since the beginning when most instance didn't have that much restaurants/shops to show, but now we can remove it, I don't think much people are using it.

@alexsegura So, should I remove map view? :)

lucasferraro:
> First impressions, fucking awesome! I love it!
> 
> Things that occur to me:
> 
> 1. in mobile, that symbol for filters isnt well known here where I am. (see photo)
>    image
>    Potentially better would be horizontal rectangular buttons with text in them, both making it clearer what they are, more obvious that they exist, and more customizable because of translation of text rather than symbols. such as...
>    image

@ERAMANrepartos @alexsegura If you think is better I can change buttons to a text button as you are suggesting. And if we remove map view, then only a "Filters" text button should exists.

alexsegura:
> So, should I remove map view? :)

@lucasferraro yes, you can remove the button. We keep the controller method to avoid 404s.

templates/restaurant/list.html.twig
```diff
@@ -32,4 +122,26 @@
 {% block scripts %}
   {{ encore_entry_script_tags('search-address') }}
   {{ encore_entry_script_tags('restaurant-list') }}
+
+  <script>
+    $('input[type=radio]').on('click', function (e) {
```
alexsegura, line 127: This selector is too wide, at least the filters should be wrapped in another top-level element to narrow the selector.

templates/restaurant/list.html.twig
```diff
@@ -32,4 +122,26 @@
 {% block scripts %}
   {{ encore_entry_script_tags('search-address') }}
   {{ encore_entry_script_tags('restaurant-list') }}
+
+  <script>
```
alexsegura, line 126: Instead of changing `window.location.href`, we could wrap the filters in a `<form method="get">`. 
Then the JavaScript code would just have to submit the form when an input is toggled.

templates/restaurant/list.html.twig
```diff
@@ -32,4 +122,26 @@
...
+      // avoid clicking another radio before navigate
+      document.querySelectorAll("input[type=radio]:not(input[type=radio]:checked)").forEach((radio) => radio.disabled=true)
+      try {
+        window.location.href = decodeURIComponent(e.target.value)
+      } catch(e) {
+        window.location.href = e.target.value
+      }
+    });
+
+    $('input[type=checkbox]').on('click', function (e) {
```
alexsegura, line 137: This selector is too wide, at least the filters should be wrapped in another top-level element to narrow the selector.

src/Controller/RestaurantController.php
```diff
@@ -214,52 +236,71 @@ public function listAction(Request $request,
...
+        $cuisines = $repository->findExistingCuisines();
 
-        if ($request->query->has('geohash') && strlen($request->query->get('geohash')) > 0) {
-            $geotools = new Geotools();
-            $geohash = $request->query->get('geohash');
+        $cacheKey = $this->getShopsListCacheKey($request, $type);
 
-            $decoded = $geotools->geohash()->decode($geohash);
+        // for filtering we need in query param the full type instead of the key
+        $request->query->set('type', $type);
```
alexsegura, line 245: Why do we need to always have a `type`? The following URL is valid `/shops?category=exclusive`, and should return all exclusive shops, no matter the type.

alexsegura (approved):

Issue #3188, opened 2022-03-30
Add a new API endpoint to retrieve filtered shops & facets


alexsegura:
Linked and needed by #3010, and also by @r0xsh to implement it on the app. 

This needs to be investigated a little bit, but we will probably need to output some kind of "facets" in the response (for example, we will need the available cuisines). Even maybe there should be 2 different endpoints, one to obtain the facets (i.e to build the filters UI), and another one to obtain the filtered results.

ERAMANrepartos:
related to search

PR #3193, opened 2022-04-05
Return facets property in restaurant list API response

alexsegura:
Fist step for #3188, this adds a `facets` property to every restaurant entry, so that search filters can be implemented client-side. To ease client-side usage, it returns translated values out of the box. 

### TODO
- [x] Return translated values for the `category` facet
- [ ] Use the `Accept-Language` HTTP header to resolve the locale

### DIFF
--- a/features/fixtures/ORM/restaurants.yml
+++ b/features/fixtures/ORM/restaurants.yml
@@ -23,6 +23,12 @@ AppBundle\Entity\Address:
     streetAddress: '17, rue Milton 75009 Paris 9ème'
     geo: "@geo_3"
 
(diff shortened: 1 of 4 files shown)


alexsegura:
@r0xsh because I'm lazy and this needs to be investigated a little bit more, I propose that we DON'T use the `Accept-Language` headers atm. The values will be translated using the default locale of the instance, and I guess the vast majority of customers will be happy with this.

lucasferraro (approved):
LGTM

r0xsh (approved):
:+1:

PR #3208, opened 2022-04-20
Shops infinite scroll

lucasferraro:
With this PR we are including an infinite scroll behaviour in `/shops` screen.

In this implementation when user navigates to a restaurant clicking on it and then wants to return with the back button the `/shops` screen will automatically scroll to top and reset the "paginator" to the begining (at the end of the video you can see this behaviour).

https://user-images.githubusercontent.com/4819244/164294704-a7cedac2-1c15-4490-8c78-9ce2bab37d60.mp4


alexsegura (commented):
When loading other pages, the time slots are not loaded. 
Basically, it's about calling the code above again after the elements are loaded:

https://github.com/coopcycle/coopcycle-web/blob/37743eb25c17678a30abfdabde166f4381f5bd94/js/app/restaurant/list.js#L13-L42

Capture d’écran 2022-04-22 à 16 25 24

js/app/restaurant/list.js
```diff
@@ -1,15 +1,19 @@
...
 import { asText } from '../components/ShippingTimeRange'
+import { useIntersection } from '../hooks/useIntersection'
 
 import 'swiper/css';
 import 'swiper/css/navigation'
 
 import './list.scss'
 
+// turn off automatic browser handle of scroll
+window.history.scrollRestoration = 'manual'
```
alexsegura, line 15: This is super important. Maybe add a more explicit comment to explain even more why we do this?

alexsegura (approved):

PR #3210, opened 2022-04-20
Use ajax for filtering

lucasferraro:
With this PR we using AJAX for make requests when filters are applied.

https://user-images.githubusercontent.com/4819244/164298702-36db2a24-47fc-4e03-bdd7-93172c4b7933.mp4

### DIFF
--- a/js/app/restaurant/list.js
+++ b/js/app/restaurant/list.js
@@ -6,6 +6,8 @@ import Swiper, { Navigation } from 'swiper'
 import { asText } from '../components/ShippingTimeRange'
 import { useIntersection } from '../hooks/useIntersection'
 
+require('gasparesganga-jquery-loading-overlay')
+
 import 'swiper/css';
 import 'swiper/css/navigation'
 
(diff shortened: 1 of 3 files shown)


lucasferraro:
> IMHO it should be more obvious that the page is loading, right now we just disable the inputs in the sidebar, but the shop items should be greyed out as well, or we should show a loading overlay. In other parts we are using `gasparesganga-jquery-loading-overlay`, or we could use [js-loading-overlay](https://github.com/muhdfaiz/js-loading-overlay).

Hmm yes, I agree with add some feedback to the user while we are loading/filtering the data. I'll check that loading component.

lucasferraro:
> IMHO it should be more obvious that the page is loading, right now we just disable the inputs in the sidebar, but the shop items should be greyed out as well, or we should show a loading overlay. In other parts we are using `gasparesganga-jquery-loading-overlay`, or we could use [js-loading-overlay](https://github.com/muhdfaiz/js-loading-overlay).

Added a loading overlay https://github.com/coopcycle/coopcycle-web/pull/3210/commits/b94542d48e011cd58ad0b818b3660d1fa7d2238d

alexsegura (approved):
LGTM

Issue #3223, opened 2022-04-27
Possibility to change groups


AdrienSICKLO:
4 different needs:
- Possibility to change the name of an Import/Group
- Add task to a group with a right click
- Ungroup a task from an existing group with right click
- Creating a group with just one task in it.

lucasferraro:
I'm creating different PRs for each request so it's easier to read and check changes

- [x] [Possibility to change the name of an Import/Group](https://github.com/coopcycle/coopcycle-web/pull/3228)
- [x] [Add task to a group with a right click](https://github.com/coopcycle/coopcycle-web/pull/3227)
- [x] [Ungroup a task from an existing group with right click](https://github.com/coopcycle/coopcycle-web/pull/3229)
- [ ] Creating a group with just one task in it.

lucasferraro:
Closed by
#3229 
#3228 
#3227

PR #3227, opened 2022-05-05
Possibility to add task/tasks to group

lucasferraro:
This PR adds the possbility to add a task/tasks to a group (https://github.com/coopcycle/coopcycle-web/issues/3223)

How it works:
https://user-images.githubusercontent.com/4819244/166986316-137ed054-e5ad-429f-8ef4-db1fe08cd487.mp4

Algo, adds the possibility of create a group with just one task in it:

https://user-images.githubusercontent.com/4819244/167160012-d9bbea81-6655-4e50-a03b-02065fa3f560.mp4


features/tasks.feature
```diff
@@ -261,6 +261,54 @@ Feature: Tasks
...
+    Given the fixtures files are loaded:
+      | sylius_channels.yml |
+      | tasks.yml           |
+      | users.yml           |
+    And the user "bob" has role "ROLE_ADMIN"
+    And the user "bob" is authenticated
+    And the tasks with comments matching "#bob" are assigned to "bob"
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I add "Accept" header equal to "application/ld+json"
+    And the user "bob" sends a "PUT" request to "/api/tasks/1/add_to_group/1" with body:
```
alexsegura, line 274: IMHO it would be more restful to do it like this:

```
POST /api/task_groups/1/tasks

{
  "tasks":["/api/tasks/1"]
}
```

src/Domain/Task/Handler/AddToGroupHandler.php
```diff
@@ -0,0 +1,21 @@
...
+
+class AddToGroupHandler
+{
+    public function __invoke(AddToGroup $command)
+    {
+        $task = $command->getTask();
+        $group = $command->getTaskGroup();
+
+        if (null !== $task->getGroup()) {
+            throw new TaskAlreadyGrouppedException(sprintf('Task #%d is already in group #%d', $task->getId(), $task->getGroup()->getId()));
```
alexsegura, line 16: `TaskAlreadyGrouppedException` -> `TaskAlreadyBelongsToGroupException`

js/app/dashboard/components/ContextMenu.js
```diff
@@ -41,6 +42,8 @@ const DynamicMenu = ({
 
   const containsOnlyUnassignedTasks = !_.find(selectedTasks, t => t.isAssigned)
 
+  const containsOnlyGrouppedTasks = selectedTasks.every(task => Object.prototype.hasOwnProperty.call(task, 'group') && task.group)
```
alexsegura, line 45: `containsOnlyGrouppedTasks` -> `containsOnlyGroupedTasks` (only one `p`)

src/Entity/Task.php
```diff
@@ -128,6 +129,16 @@
  *         }
  *       }
  *     },
+ *     "task_add_to_group"={
+ *       "method"="PUT",
+ *       "path"="/tasks/{id}/add_to_group/{group_id}",
+ *       "controller"=AddToGroup::class,
+ *       "denormalization_context"={"groups"={"task_operation"}},
+ *       "access_control"="is_granted('ROLE_ADMIN') or is_granted('edit', object)",
```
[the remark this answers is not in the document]
lucasferraro: How does that work? Because if I remove `is_granted('ROLE_ADMIN')` and in the test replace `And the user "bob" has role "ROLE_ADMIN"` with `And the user "bob" has role "ROLE_USER"` the operation is executed without errors.

Also seems like there was an error in what I did here `is_granted('edit', object)` because if I put only that constraint doesn't matter what role user has, the operation always returns a 403.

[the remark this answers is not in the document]
alexsegura: > How does that work?


When there is a `security` or `access_control` property, the [services tagged with `security.voter`](https://github.com/coopcycle/coopcycle-web/blob/master/app/config/services.yml#L1007-L1038) are invoked, and their `supports` method is called, to check if they are able to perform the necessary checks. 

Your problem here is that [`TaskOperationsVoter`](https://github.com/coopcycle/coopcycle-web/blob/9232b245c848668306dc473a359fa6f64fd64c31/src/Security/TaskOperationsVoter.php) only supports the `view` attribute. But if you change the operation to be more "restful" (see other comment), your operation definition would be on `TaskGroup`, and the associated voter supports the `edit` attribute.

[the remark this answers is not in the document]
alexsegura: So, sorry, my comment was wrong ☹️

features/tasks.feature
alexsegura: If we add the operation at the group level, we can add multiple tasks to a group with just one request, no?

```
POST /api/task_groups/1/tasks

{
  "tasks":[
    "/api/tasks/1",
    "/api/tasks/2",
    "/api/tasks/3"
  ]
}
```

lucasferraro: Yes, I agree. I'll try to make this :)

alexsegura (approved):

PR #3240, opened 2022-05-20
improve experience for creating a pricing rule condition when there a…

### LINKED ISSUE #3128: Improve selection of packages when creating a pricing rule condition
When there is a lot of packages, they are not sorted, and it's difficult to find what you are looking for. 

They should be sorted at least. Even better, we could use [react-select](https://react-select.com/home) to make the dropdown searchable. 

Capture d’écran 2022-02-16 à 10 33 18

sfishel18:
…re many packages

- sort the packages alphabetically in the selection list
- switch to `react-select` to support searching

fixes #3128

### DIFF
--- a/js/app/components/PricePerPackageEditor.js
+++ b/js/app/components/PricePerPackageEditor.js
@@ -1,6 +1,41 @@
 import React, { useState } from 'react'
 import { getCurrencySymbol } from '../i18n'
 import { useTranslation } from 'react-i18next'
+import Select from 'react-select'
+import _ from 'lodash'
+
+/**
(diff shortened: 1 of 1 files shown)


sfishel18:
i had to wrestle with the layout a little. when i tried to use any kind of float or inline/flex display for the react-select control it would collapse down to a tiny width, so i opted for a fixed width instead.

another option would be to upgrade to version 5 of react-select, which fixes the layout issue but uses css grid so doesn't work in ie 11

Screen Shot 2022-05-19 at 9 07 24 PM

lucasferraro (approved):
LGTM :)

PR #3261, opened 2022-06-04
refund canceled orders when made with giropay

### LINKED ISSUE #3194: Order paid with Giropay should be refunded when cancelled
Atm it's only refunded when the order is refused. It should be refunded when cancelled as well. 

The code below should also be executed in [CancelOrderHandler](https://github.com/coopcycle/coopcycle-web/blob/285550631112d4e83d9ee232e09d668f03542536/src/Domain/Order/Handler/CancelOrderHandler.php)

https://github.com/coopcycle/coopcycle-web/blob/6e5639f4faca5d1abf8a167f7e8143e6a91966ab/src/Domain/Order/Handler/RefuseOrderHandler.php#L28-L30

sfishel18:
fixes #3194

### DIFF
--- a/app/config/message_bus/order.yml
+++ b/app/config/message_bus/order.yml
@@ -63,6 +63,7 @@ services:
   coopcycle.domain.order.handler.cancel_order:
     class: AppBundle\Domain\Order\Handler\CancelOrderHandler
     arguments:
+      - '@coopcycle.stripe_manager'
       - '@event_recorder'
     tags:
       - name: command_handler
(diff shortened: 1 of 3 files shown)


sfishel18:
hi @alexsegura, no rush of course but wanted to make sure you saw this. it looks like i don't have permissions to add reviewers to the PR.

sfishel18:
> Thank you @sfishel18
> 
> Sorry it took so long to merge ☹️

@alexsegura no problem at all! let me know if you can think of any other beginner tasks i could look into, otherwise i'll just check out the open issues and find something that looks interesting

PR #3263, opened 2022-06-06
add a filter by tag to some tasks' related graphs

vladimir-8:
Before:
Screenshot 2023-04-26 at 15 39 39

After:
Screenshot 2023-04-26 at 15 40 01


r0xsh (commented):
The addition of the filter is a good idea, the code seems clean. My only concern is that some calculations are now being performed on the client-side. I'm afraid that on cooperatives with a large number of orders, it might slow down the opening of the metrics page.

r0xsh:
> @r0xsh, thanks for the review! Which calculations do you think might be an issue? I can try moving some of them to the cube.
> 
> These graphs currently are a sort of "proof-of-concept". They might need adjustments, so I'll refrain from optimizing them too much too early 😉

Well, know that i've a bit more about our stack, i feel like my initial code review was wrong.
Things seems nice, and yeah your schemas will work with CubeJS even since our migration to ClickHouse back-end

r0xsh (approved):

PR #3291, opened 2022-06-22
Implement Woopit

### LINKED ISSUE #3289: Implement Woopit
https://woop.slab.com/public/posts/i4y8hn45
https://woop.stoplight.io/docs/carrier

**TODO**

- [x] Add a database table to know which deliveries are for Woopit (for example `woopit_delivery`). A new line on this table would be added when `AppBundle\Action\Woopit\DeliveryRequest` is executed. 
- [x] Create a new message/message handler to send the [status updates](https://woop.stoplight.io/docs/carrier/d6d92c6eac0a4-status-update-of-a-delivery-or-a-package) (like `AppBundle\Message\Webhook` & `AppBundle\MessageHandler\WebhookHandler`). This is where the `woopit_delivery` table will be helpful, to know if we need to trigger a status update for the delivery being updated. The list of statuses expected by Woopit can be found [here](https://woop.stoplight.io/docs/carrier/29a6c7af2fbda-delivery-status). 
- [x] Add a new endpoint to implement [the delete operation](https://woop.stoplight.io/docs/carrier/49b91925f46d4-delete-a-delivery). The implementation is the same as the [delete operation](https://github.com/coopcycle/coopcycle-web/blob/7a9d90c5b83b2d80905dfcce8fc942431434d4a6/src/Action/Delivery/Cancel.php) on our side. 
- [x] Implement the [subscription](https://woop.stoplight.io/docs/carrier/ZG9jOjUwNzc3OTA4-subscriptions) to subscribe to requests and events related to orders placed on the Woop platform.
- [x] Add a new endpoint to implement the [update operation](https://woop.stoplight.io/docs/carrier/d3532629fd722-update-delivery-status). Make sure that API Platform will accept the [PATCH operation](https://api-platform.com/docs/core/content-negotiation/#configuring-patch-formats)
- [ ] Implement the [upload of documents](https://woop.stoplight.io/docs/carrier/665af14732ad0-adding-a-delivery-document) to Woopit. A possible implementation, to react when images are added to a task, is to implement a Doctrine subscriber, that checks updates on a Task entity, using the [preUpdate](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/events.html#preupdate) event, and checking if the `images` field has changed. Then, an async message should be added to the queue, to actually implement the request to Woopit.

alexsegura:
Fixes #3289


features/woopit.feature
```diff
@@ -522,3 +522,300 @@ Feature: Woopit
...
+      """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the JSON should match:
+      """
+      {
+        "deliveryId":"yXJPEP30xgja"
+      }
+      """
+    When I add "Content-Type" header equal to "application/merge-patch+json"
```
alexsegura, line 734: Does it work without this header? Not sure it will be sent by Woop, I think they will send `Content-Type: application/json` as per the documentation

alexsegura: Maybe you will need to use this? https://api-platform.com/docs/core/content-negotiation/#configuring-formats-for-a-specific-resource-or-operation

PR #3358, opened 2022-08-05
Allow importing packages via spreadsheets

### LINKED ISSUE #3337: Allow importing packages via spreadsheets
We want to be able to import packages when importing deliveries and tasks. To implement this, we need to add a new column, that allows to specify the packages as a string. 

A (relatively) "human readable" format could be:
```
bouquet-xl=2 moyen-bouquet=3
```

Any ideas about the best format @biologeek?

--- 

But before this, we need to slugify the package names in database, so that: 
- they are unique
- they are not linked with the name of the package. 

In the example above, we would end up having: 
- bouquet-xl
- moyen-bouquet
- petit-bouquet

Capture d’écran 2022-07-22 à 14 45 41

It should be easy to implement using the [Sluggable behavior of Doctrine](https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/doc/sluggable.md). 

The slugs to use would need to be displayed somewhere in the packages screens, so that users know which ones need to be used. 

--- 

This needs to be implemented both in `DeliverySpreadsheetParser` and `TaskSpreadsheetParser`. 
When importing deliveries, the packages would be in a `dropoff.packages` column.

lucasferraro:
Fixes #3337 

- [x] slugify the package names in database
- [x] allow import packages when importing deliveries and tasks

### DIFF
--- a/app/DoctrineMigrations/Version20220805151009.php
+++ b/app/DoctrineMigrations/Version20220805151009.php
@@ -1,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use AppBundle\Entity\Package;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220805151009 extends AbstractMigration implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE package ADD slug VARCHAR(255)');
+        $this->addSql('CREATE UNIQUE INDEX UNIQ_DE686795989D9B62 ON package (slug)');
+    }
+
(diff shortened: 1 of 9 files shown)


app/DoctrineMigrations/Version20220805151009.php
```diff
@@ -0,0 +1,52 @@
...
+    public function postUp(Schema $schema): void
+    {
+        $em = $this->container->get('doctrine.orm.entity_manager');
+
+        $packages = $this->container->get('doctrine')
+            ->getRepository(Package::class)
+            ->findAll();
+
+        foreach ($packages as $package) {
+            $package->setSlug($package->getName());
```
alexsegura, line 40: The name should be slugified

alexsegura (changes requested):

app/DoctrineMigrations/Version20220805151009.php
alexsegura, line 40: Just to be sure: because you are using `postUp`, the name will be slugified automatically using the Doctrine behavior?

lucasferraro: Yes. There's a way to 'urilize' strings with this `Gedmo\Sluggable\Util\Urlizer::urlize(...)` but it doesn't take into account existing slugs in DB. So if you already have a slug 'small-box' in DB using Doctrine behaviour  for the same string it will generate the slug 'small-box-1', and etc.

alexsegura:
@lucasferraro it's fine to avoid dying if a package does not exist

PR #3372, opened 2022-08-24
New integration admin form to validate requests based on dimensions, weight, zones and products

### LINKED ISSUE #3362: Reject quote request based on dimensions, weight, and area/zone
We need to make sure that quote requests that are not feasible are rejected. 

We need to be able to configure, at an instance level, the constraints in terms of dimensions for Woop. 

To make it easier for everyone, and avoid using yet another env variable, let's create a new page under `/admin/integrations/woopit`, that will allow configuring the constraints. 
We can store the configuration via `SettingsManager`, or into a dedicated table. 
For future integrations, we can also use the same pattern. 

#### Dimensions 

We still need to determine if it's expressed in terms of max length/height/width, or in m³. 
If it's expressed in m³, we need to sum up the volume of all products. 

#### Weight 

We still need to determine if we should reject the request when ONE of the products is above the limit, or if the sum of products is above the limit. 

#### Zone

Quote requests can originate from multiple store at different locations. Ideally, we should be able to map a zone to a store sent by Woop (under `retailer.store.id`). But we would need to know by advance all the store ids (seems possible to ask them to Woop each time). 

💡 Another implementation would be to allow mapping a CoopCycle store, to a Woop store, and define the zone constraint at the CoopCycle store level. There is already a "hidden feature" for this (see [CheckDeliveryValidator](https://github.com/coopcycle/coopcycle-web/blob/031901e8a71f00d07ee776527e79f03ecb43e8a6/src/Validator/Constraints/CheckDeliveryValidator.php)). This would require more configuration (setup the Woop store, and the other stores in CoopCycle, and do the mapping). Then, the Woop store id could be associated to the CoopCycle store.

lucasferraro:
Closes #3362


src/Controller/AdminController.php
```diff
@@ -1490,6 +1492,98 @@ public function apiAppAction($id, Request $request)
         ]);
     }
 
+    /**
+     * @Route("/admin/integrations", name="admin_integrations")
```
alexsegura, line 1496: Should be `/admin/integrations/woopit` + `admin_integrations_woopit`. 
`/admin/integrations` should exist, but just show "Woopit".

alexsegura (approved):

src/Form/WoopitIntegrationType.php
```diff
@@ -0,0 +1,109 @@
...
+                },
+                'expanded' => true,
+                'multiple' => true
+            ]);
+
+        $builder->addEventListener(FormEvents::SUBMIT, function (FormEvent $event) {
+            $woopitIntegration = $event->getData();
+
+            if (null !== $woopitIntegration->getZone()) {
+                $woopitIntegration->getStore()->setCheckExpression(
```
alexsegura, line 96: Smart 💡

PR #3379, opened 2022-08-30
split terms and conditions and privacy policy

lucasferraro:
Split the checkboxes "terms & conditions" & "privacy policy".
Make it configurable via an environment variable, so that we can keep on checkbox for other countries.


.env.dist
```diff
@@ -109,3 +109,5 @@ DABBA_ENABLED=0
 DABBA_BASE_URL=https://api.dabba-consigne.fr
 DABBA_CLIENT_ID=
 DABBA_CLIENT_SECRET=
+DABBA_ENABLED=0
```
alexsegura, line 112: Duplicate env var

lucasferraro:
Included the expose of new ENV var through Settings endpoint (to be used in the app).

alexsegura (approved):

PR #3441, opened 2022-10-17
Changed pickup done after to before delivery dropoff

biologeek:
Pickup should occur between "dropoff beginning - X minutes" and "dropoff end - X minutes" where X corresponds to ride duration

****** THIS WAS NOT TESTED ******

### DIFF
--- a/src/Form/DeliveryImportType.php
+++ b/src/Form/DeliveryImportType.php
@@ -63,7 +63,7 @@ class DeliveryImportType extends AbstractType
                 foreach ($deliveries as $delivery) {
                     $violations = $this->validator->validate($delivery);
                     if (count($violations) > 0) {
-                        throw new \Exception($violations->get(0)->getMessage());
+                        throw new \Exception((string) $violations->get(0));
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


src/Service/DeliveryManager.php
```diff
@@ -164,9 +164,12 @@ public function setDefaults(Delivery $delivery)
                     $duration = $this->routing->getDuration(...$coords);
 
                     $pickupDoneBefore = clone $dropoff->getDoneBefore();
+                    $pickupDoneAfter = clone $dropoff->getDoneAfter();
```
alexsegura, line 167: The problem is that at this step, we may not have the beginning of the dropoff slot. 
On the API or on the form to create a delivery, the minimal data we ask is the end of the dropoff timeslot. 
The beginning of the slot, if missing, is then "calculated" with a default of 15 minutes in [TaskListener](https://github.com/coopcycle/coopcycle-web/blob/e247f64de5a5dccd2841372633d0ca6882162e2e/src/Entity/Listener/TaskListener.php#L31-L38)

biologeek: Got it, I'll check if it can be done another way

PR #3471, opened 2022-11-09
Allow admins to select a date in the past when creating a delivery

### LINKED ISSUE #3465: Allow selecting a date in the past when creating a delivery
This should be possible only for administrators. 

When the user is connected as an administrator, show a `DatePicker.RangePicker` (like in the dispatch dashboard) that allows to select any date, including in the past. 

This should be decided in `createDatePickerWidget`

https://github.com/coopcycle/coopcycle-web/blob/d6ed25ef9ed843592400cfadb1204e86cb2f4ebe/js/app/forms/delivery.js#L127-L161

lucasferraro:
Closes #3465 

**Store with timeslot configured. Admin user is able to select a free range**
Storewithtimeslotconfigured Adminuserisabletoselectafreerange

**Store with timeslot configured. Store user is only able to select a timeslot, not a free range**
not_admin_timeslot_now_allowed_free_range

**Store without a timeslot configured. Admin user is able to select a free range**
Storewithoutatimeslotconfigured Adminuserisabletoselectarange

**Store without a timeslot configured. Store user is able to select a range only in the future**
not_admin_not_allowed_past_dates

### DIFF
--- a/js/app/forms/delivery.js
+++ b/js/app/forms/delivery.js
@@ -124,8 +124,7 @@ function getTaskType(name, type) {
   return document.querySelector(`#${name}_${type}_type`).value.toUpperCase()
 }
 
-function createDatePickerWidget(name, type) {
-
+function createDatePickerWidget(name, type, userAdmin) {
   const datePickerEl = document.querySelector(`#${name}_${type}_doneBefore`)
   const timeSlotEl = document.querySelector(`#${name}_${type}_timeSlot`)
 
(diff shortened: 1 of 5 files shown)


js/app/forms/delivery.js
```diff
@@ -124,8 +124,7 @@ function getTaskType(name, type) {
   return document.querySelector(`#${name}_${type}_type`).value.toUpperCase()
 }
 
-function createDatePickerWidget(name, type) {
-
+function createDatePickerWidget(name, type, userAdmin) {
```
alexsegura, line 127: When it's an admin, we should show a `DatePicker.RangePicker` instead of a `DateTimePicker`.

lucasferraro: We're already showing a `DateTimePicker` when store has not a timeslot configured. 
Why should we introduce a new component here instead of use that one?

alexsegura (approved):

PR #3481, opened 2022-11-18
New address parameter

### LINKED ISSUE #3476: Do not trust session storage for address search
When executing the following sequence:
- On homepage, search for an address
- Click on a restaurant
- Click on "View all restaurants"

We end up showing the restaurant list page with an address in the search bar, but the list shows all the restaurants, including those who are too far for this address. 

search_address

This can lead to confusion. We should make sure that when we end up on the list page, we show a list of restaurants that corresponds to what is showing in the search bar. 

We should always have an `address` parameter, like when user is selecting an address that already exists. 

https://github.com/coopcycle/coopcycle-web/blob/20ea14611e43f9774f9083fd50482990e9f2f25d/js/app/search/address.js#L43-L80

The `address` parameter could be a base64-encoded JSON object, no matter if is a new address, or a saved address. 

This means the following: 
- URL is like `/fr/shops?geohash=XXX` --> Do not show any address as text in the search bar
- URL is like `/fr/shops?geohash=XXX&address=XXX` --> Show the address corresponding to the `address` parameter

Also, when navigating back to the restaurants list page, the `address` parameter should be transmitted. 

--- 

To introduce "always have an address" concept, let's also ask for the address on the restaurant list page. 
If there is no address in the search bar, when the user clicks on a restaurant, scroll into the search bar, and ask for an address.

lucasferraro:
Closes #3476 

This is how it looks now
https://user-images.githubusercontent.com/4819244/202785390-9b44b8cb-c090-4162-8ce1-9cd2c97504f2.mp4

### DIFF
--- a/src/Controller/ArgumentResolver/AddressValueResolver.php
+++ b/src/Controller/ArgumentResolver/AddressValueResolver.php
@@ -6,6 +6,7 @@ use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\ItemNotFoundException;
 use AppBundle\Entity\Address;
+use AppBundle\Entity\Base\GeoCoordinates;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
 use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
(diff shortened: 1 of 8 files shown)


lucasferraro:
@alexsegura a weird behaviour that happens with this implementation is the following:

- In homepage now the user is not able to enter to a restaurant page because if there isn't an address selected we'll ask for an address preventing the navigation to restaurant page. And when user enters an address, we automatically navigate to `shops` page, only in this page the user is able to click on a restaurant and navigate to its page.

https://user-images.githubusercontent.com/4819244/202785998-834c87f6-e9ae-4607-99ef-06585909722d.mp4

alexsegura:
@lucasferraro I had an exception when clicking on a restaurant. 

Capture d’écran 2022-11-23 à 15 39 19

The "real" exception is this one: 

```
Uncaught PHP Exception TypeError: "get_object_vars(): Argument #1 ($object) must be of type object, null given" at /var/www/html/src/Controller/ArgumentResolver/AddressValueResolver.php line 51
```

src/Controller/ArgumentResolver/AddressValueResolver.php
```diff
@@ -46,27 +47,35 @@ public function resolve(Request $request, ArgumentMetadata $argument)
 
         $user = $this->security->getUser();
 
-        if (!is_object($user)) {
-            // e.g. anonymous authentication
-            return yield null;
-        }
+        $data = base64_decode($value);
+        $data = get_object_vars(json_decode($data));
```
alexsegura, line 51: I think `get_object_vars(json_decode($data))` can be simplified to `json_decode($data, true)` ([2nd parameter ](https://www.php.net/manual/en/function.json-decode.php)allows to return an associative array)

alexsegura:
@lucasferraro 

> In homepage now the user is not able to enter to a restaurant page because if there isn't an address selected we'll ask for an address preventing the navigation to restaurant page. And when user enters an address, we automatically navigate to shops page, only in this page the user is able to click on a restaurant and navigate to its page.

Yes, it's quite confusing. Can we use `applyClickListenerForRestaurantItem` only on the list page?

alexsegura:
It don't seem to be able to change the address after one search

address_search_bug

src/Controller/ArgumentResolver/AddressValueResolver.php
lucasferraro: Great, thanks

alexsegura (approved):

Issue #1456, opened 2020-06-02
In order to [enable grouped orders] As a [client and restaurant and courier] I want [a new feature for clients to create a groupe and then order with this group]


Doyoubouilloux:
**In order to**  `enable grouped orders`
Enable clients working in companies to create a group on the platform resto.paris and then order with this group. No one has to pay for the group, each person within the group pay for its own order witch is then added on the group order.

**As a** `client`
The client won't have to pay for the whole group and can benefit from the grouped order to have free delivery (because the group order will be above a certain amount of money)

**As a** `restaurant and courier`
The restaurant and courier will have regular orders from companies and will be able to organize its work better in advanced. It is secure income as well.

**I want** `a feature enablelling grouped orders`
One client must be able to create a group and invite its colleagues
Each member of the group must be able to créate a group order and then pay for their own order
The order must be validated once the order reaches a certain amount of money. Then the delivery is organised.
Each group = one adress = a perimeter with restaurants where they can order

**Whereas currently** `one client has to take responsability for the group order`
Now, someone in the company has to pay the order for the whole groupe with its personal account...

**Additional context**
Feature required for resto.paris in order for the plateform to make a difference and reach companies need
[Resto.paris_spec commande groupée.pdf](https://github.com/coopcycle/coopcycle-web/files/4718093/Resto.paris_spec.commande.groupee.pdf)

alexsegura:
Basically, a customer may start an order, and invite friends to add items to this same order. 

### Implementing the shareable link

An order can be shared via a JSON Web Token. The JWT contains custom claims with: 
- the order id
- the email of the guest
- (maybe also the id of the "host" customer)

Use the example below on [jwt.io](https://jwt.io/)

```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJvcmRlciI6MSwiZ3Vlc3QiOiJtZXguemt0a0BnbWFpbC5jb20ifQ.uMwUEU__oT4MsL1B4q6y-iHko45E3APQS92CW6EN5IE
```

This allows us to restore the order from database easily. 
The shareable link will be like `/order/share?token=${jwt}`

The shareable link does not require to have an account on the website, or to be signed in, anybody having the link can use it. 

### Changes to the order model

[...]

PR #3535, opened 2023-02-01
Group orders

lucasferraro:
WIP for #1456 .

What has been done so far:

- New API endpoint `/api/orders/{id}/invite` which creates a JWT token with the data and send emails with a shareable link to guests.
- Controller created to be called when the shareable link `/order/share?jwt={jwt}` is invoked. After some checks, in progress too, if JWT is valid it will render the restaurant page.
- New button in restaurant page (cart section) to invite people to add products to the order. When it is pressed a modal will be opened asking for people emails. (submit of this modal is making the API call to `/invite` but is failing because of auth issues).


r0xsh:
Fixes #1456

src/Controller/RestaurantController.php
```diff
@@ -567,6 +570,7 @@ public function cartAction($id, Request $request,
             $this->orderManager->flush();
         }
 
+        $this->eventBus->handle(new OrderUpdated($cart));
```
alexsegura, line 573: Would have been cool to avoid duplicating this piece of code, and have it in a central place. Maybe in `OrderModifier`?

js/app/i18n/locales/es.json
```diff
@@ -305,6 +305,13 @@
         "PAY_WITH_SAVED_CREDIT_CARD": "Pagar con una tarjeta de crédito guardada",
         "ADD_NEW_CREDIT_CARD": "Ingresar nueva tarjeta de crédito",
         "ADD_OR_SELECT_A_CARD_TO_PAY": "Por favor seleccione o ingrese una tarjeta de crédito para realizar el pago",
+        "INVITE_PEOPLE_TO_ADD_ITEMS": "Invita amigos para que agreguen productos",
+        "ADD_GUESTS_EMAILS_TITLE": "Agrega los emails de las personas que quieres invitar",
+        "ENTER_EMAILS_SEPARATED_BY_COMMA": "Ingresa los emails separados por coma",
+        "SEND_INVITATIONS": "Enviar invitaciones",
+        "ADD_AT_LEAST_ONE_EMAIL": "Agregar al menos un email",
+        "ONE_OR_MORE_EMAIL_IS_INVALID": "Uno o más emails son inválidos. Por favor chequea las direcciones ingresadas.",
+        "SEND_INVITATIONS_FAILURE": "Ha ocurrido un error al intentar enviar las invitaciones.",
```
alexsegura, line 314: Unused

alexsegura:
@r0xsh I have a small doubt: the `sylius_order_item.customer_id` is always set, even for the "owner" of the order. In my mind, it was supposed to stay `NULL` for the "owner", so that it's easy to distinguish via a SQL query. Also, orders made before the feature was introduced will have `NULL` in this column forever. 

My doubt is that it complicates things a little bit, because this column can be NULL or equal to the `sylius_order.customer_id` column

r0xsh:
> I have a small doubt: the `sylius_order_item.customer_id` is always set, even for the "owner" of the order. In my mind, it was supposed to stay `NULL` for the "owner", so that it's easy to distinguish via a SQL query. Also, orders made before the feature was introduced will have `NULL` in this column forever.
> 
> My doubt is that it complicates things a little bit, because this column can be NULL or equal to the `sylius_order.customer_id` column

Hum :thinking:
You're right, always set `sylius_order_item.customer_id` is not the right approach. The idea behind this change was to always have a customer that is returned in order to have the least conditional logic possible between all the actors of a shared cart (including the owner of the cart), it simplifies the code and reduces the error surface.

Maybe doing a conditional rendering in the `OrderItemNormalizer` like this would be more intelligent
```php
if (is_null($object->getCustomer()) {
   $player = $object->getCart()->getOwner()->getCustomer();
} else {
   $player = $object->getCustomer();
}
```



https://github.com/coopcycle/coopcycle-web/blob/79367d1b7e9f744ed1dac32ff19a78c723925abc/src/Serializer/Json/OrderItemNormalizer.php#L74

alexsegura (approved):
There is still some room for improvements, but let's merge this and move forward. Amazing work, thanks ❤️

PR #3576, opened 2023-03-14
Dispatcher role

### LINKED ISSUE #2484: [USERS] Create new user role → Dispatcher

r0xsh:
Fixes #2484

- [x] Menu area	Action	dispatcher allowed?
- [x] Dashboard	Add task	yes
- [x] 		Edit tak	yes
- [x] 		(Re)assign task	yes
- [x] 		Delete task	yes
- [x] 		Import tasks from file	yes
- [x] 		Export tasks to file	yes, with limits (see below)
- [x] Orders		yes
- [x] Deliveries – Deliveries tab	Create delivery	yes
- [x] 				Import deliveries	yes
- [x] 				Export deliveries	yes, with limits (see below)
- [x] 				View past deliveries	yes, with limits (see below)
- [x] Deliveries – Stores		no
- [x] Deliveries – Pricing		no
- [x] Deliveries – Forms		no
- [x] Deliveries – Time slots		no
- [x] Deliveries – Packages		no
- [x] Deliveries – Zones		no
- [x] Deliveries – Tags		Create tag	yes
- [x] 				Edit tag	no
- [x] 				Delete tag	no
- [x] Deliveries – API		no
- [x] Deliveries – Vehicles		no
- [x] Deliveries – Integrations	no
- [x] Shops				don‘t know, we don‘t use this
- [x] Users		View users	no
- [x] 		Edit user	no
- [x] 		Invite user	yes, but only couriers
- [x] 		Export	no
- [x] Statistics		probably not
- [x] Settings		no

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -289,7 +289,7 @@ security:
         - { path: ^/profile/notifications, roles: IS_AUTHENTICATED_REMEMBERED }
         - { path: ^/profile, roles: IS_AUTHENTICATED_REMEMBERED }
         - { path: ^/admin/settings, roles: ROLE_ADMIN }
-        - { path: ^/admin/.*$, roles: ROLE_DISPATCHER }
+        - { path: ^/admin, roles: ROLE_DISPATCHER }
         - { path: ^/_uploader/restaurant, roles: [ ROLE_ADMIN, ROLE_RESTAURANT ] }
         - { path: ^/api/login, roles: PUBLIC_ACCESS }
         - { path: ^/api/facebook/login$, roles: PUBLIC_ACCESS }
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -288,7 +288,8 @@ security:
         - { path: ^/profile/jwt, roles: IS_AUTHENTICATED_REMEMBERED }
         - { path: ^/profile/notifications, roles: IS_AUTHENTICATED_REMEMBERED }
         - { path: ^/profile, roles: IS_AUTHENTICATED_REMEMBERED }
-        - { path: ^/admin, roles: ROLE_ADMIN }
+        - { path: ^/admin/settings, roles: ROLE_ADMIN }
+        - { path: ^/admin/.*$, roles: ROLE_DISPATCHER }
         - { path: ^/_uploader/restaurant, roles: [ ROLE_ADMIN, ROLE_RESTAURANT ] }
         - { path: ^/api/login, roles: PUBLIC_ACCESS }
         - { path: ^/api/facebook/login$, roles: PUBLIC_ACCESS }
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -14,7 +14,8 @@ security:
         ROLE_RESTAURANT: ROLE_USER
         ROLE_STORE: ROLE_USER
         ROLE_LOOPEAT: ROLE_USER
-        ROLE_ADMIN: [ ROLE_COURIER, ROLE_LOOPEAT ]
+        ROLE_DISPATCHER: [ ROLE_COURIER, ROLE_LOOPEAT ]
+        ROLE_ADMIN: ROLE_DISPATCHER
 
     # http://symfony.com/doc/current/book/security.html#where-do-users-come-from-user-providers
     providers:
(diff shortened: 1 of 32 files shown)


src/Controller/TagController.php
```diff
@@ -33,6 +34,9 @@ public function newTagAction(Request $request)
 
     public function tagAction($slug, Request $request)
     {
+        if (!$this->isGranted("ROLE_ADMIN")) {
+            throw new AccessDeniedException();
+        }
```
alexsegura, line 39: Can be simplified to `$this->denyAccessUnlessGranted('ROLE_ADMIN');` (see [docs](https://symfony.com/doc/5.4/security/voters.html))

src/Controller/AdminController.php
```diff
@@ -892,6 +893,9 @@ public function taxationSettingsAction(
     public function tagsAction(Request $request, TagManager $tagManager)
     {
         if ($request->isMethod('POST') && $request->request->has('delete')) {
+            if (!$this->isGranted('ROLE_ADMIN')) {
+                throw new AccessDeniedException();
+            }
```
alexsegura, line 898: Can be simplified to `$this->denyAccessUnlessGranted('ROLE_ADMIN');` (see [docs](https://symfony.com/doc/5.4/security/voters.html))

app/config/security.yml
```diff
@@ -287,7 +288,8 @@ security:
         - { path: ^/profile/jwt, roles: IS_AUTHENTICATED_REMEMBERED }
         - { path: ^/profile/notifications, roles: IS_AUTHENTICATED_REMEMBERED }
         - { path: ^/profile, roles: IS_AUTHENTICATED_REMEMBERED }
-        - { path: ^/admin, roles: ROLE_ADMIN }
+        - { path: ^/admin/settings, roles: ROLE_ADMIN }
+        - { path: ^/admin/.*$, roles: ROLE_DISPATCHER }
```
alexsegura, line 292: Actually `^/admin` should be accessible to `ROLE_DISPATCHER`, because the button to access admin interface points to this URL, and it needs to be maintained even if it redirects to a default page. 

It's convenient to just type `/admin` (this is what I do 🙂 )

Capture d’écran 2023-03-20 à 15 26 46

src/Controller/AdminController.php
```diff
@@ -920,6 +942,7 @@ public function tagsAction(Request $request, TagManager $tagManager)
      */
     public function pricingRuleSetsAction()
     {
+        $this->denyAccessUnlessGranted('ROLE_ADMIN');
```
alexsegura, line 945: Are we sure it's better to add `denyAccessUnlessGranted`, than using `access_control` in `security.yml` (like you did for `/admin/settings`)? I see a lot of repetition here. 

Like written in the [docs](https://symfony.com/doc/5.4/security.html#add-code-to-deny-access), those are the only 2 ways. Maybe our URL structure does not fit this?

> You can define as many URL patterns as you need - each is a regular expression. BUT, only one will be matched per request: Symfony starts at the top of the list and stops when it finds the first match

src/Controller/AdminController.php
```diff
@@ -887,11 +908,12 @@ public function taxationSettingsAction(
     }
 
     /**
-     * @Route("/admin/settings/tags", name="admin_tags")
+     * @Route("/admin/tags", name="admin_tags")
```
alexsegura, line 911: URLs for time slots & packages also have the `/settings` sub-path, don't they need to be changed too? Or it's only needed for tags, as dispatchers are allowed to manipulate tags.

src/Controller/AdminController.php
r0xsh: The job of the controller is to :

1. Validate user input
2. Check if the user is allowed to perform the action
3. Do business shenanigans (transform input data for the model or view)
4. Build a response if needed

To me, like network firewalls, the Symfony firewall is a nice new layer in the security onion – but does not dispense with security checks at the lowest level of entry points

src/Controller/AdminController.php
r0xsh: Indeed, the change has been made for the `tags` because I have passed over them when creating the role dispatcher. But for the sake of consistency, it would be nice to change the url for the time slots and packages. For me the `/settings` route should contain the "technical" settings of the instance and not the CRUD of the business entities

alexsegura (approved):

PR #3584, opened 2023-03-21
X orders per Y minutes

### LINKED ISSUE #3580: Allow rate limiting orders by time window for a restaurant
I want to make it possible to limit the total amount of product sold to be picked up in a 15 minute time range (alternatively, a total number of orders)

**As a** `role`
As a restaurant owner/chef

I want to define a maximum amount of basket amount in euros that can enter in a specific time range, and any order that pushes the total over the limit, this order will be put into the next delivery range slot. 
For example: for increments of 15 minutes, my restaurant can make a total of 10 orders with an average size of 2 hamburgers with fries, which is generally a 30 euro order. I want to be able to state that in 15 minutes the limit is 300 euros of product, and if an order enters that pushes the total income from that 15 minutes over 300 euros, then that order will be put into the next 15 minute pickup or preparation time range.

Currently I can use "rush mode" to increase my preparation times, but this is a REACTIVE response and not at all dynamic. So no matter if I use this or not, I can have 10 orders enter within 5 minutes, all with the same pickup time. This is impossible to prepare, and during such a shift, I cannot stop to pick which 5 orders I want to delay, delay them one by one, etc

**Additional context**
Restaurant options in regards to normal preparation time and rush preparation time
image

Button to active Rush Mode as a restaurant
image

Dispatcher delay bar for entire cooperative delivery service
image

The "Cumplimiento" menu page on a restaurant with opening hours and other configurations, where we imagine this feature would exist, potentially as a piece of information you add to the time ranges for delivery range options of 10, 30 or 60 minutes
image

MORE CONTEXTUAL INFORMATION
In a more general sense, we want to add more tools to a restaurant in regards to how orders enter. This is because traditionally a restaurant has a relatively low level of orders and the issues could be resolved with the "rush mode" or manually by delaying individual orders. Now we are having, and hope to have, more restaurants with 50+ orders in a night. To do this effectively we need more tools that allow restaurants and the cooperative to manage their "production". Other options might include:
1. A button that closes the next available 15 minute pickup window time range
2. A delay bar for individual restaurants that only affects their orders
3. Splitting the visual organization in the application of orders into time ranges with borders between them, and allowing the possibility to drag and drop an order from one time range into another, with automatic notification to the cooperative and the client about the change in pickup time

And a photo of

r0xsh:
Fixes #3580

### DIFF
--- a/app/DoctrineMigrations/Version20230321150513.php
+++ b/app/DoctrineMigrations/Version20230321150513.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20230321150513 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE restaurant ADD max_orders_range_duration INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE restaurant ADD max_orders_amount INT DEFAULT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE restaurant DROP max_orders_range_duration');
+        $this->addSql('ALTER TABLE restaurant DROP max_orders_amount');
+    }
+}
--- a/app/DoctrineMigrations/Version20230321150513.php
+++ b/app/DoctrineMigrations/Version20230321150513.php
@@ -20,7 +20,7 @@ final class Version20230321150513 extends AbstractMigration
     public function up(Schema $schema): void
     {
         // this up() migration is auto-generated, please modify it to your needs
-        $this->addSql('ALTER TABLE restaurant ADD rate_limitr_range_duration INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE restaurant ADD rate_limit_range_duration INT DEFAULT NULL');
         $this->addSql('ALTER TABLE restaurant ADD rate_limit_amount INT DEFAULT NULL');
     }
 
--- a/app/DoctrineMigrations/Version20230321150513.php
+++ b/app/DoctrineMigrations/Version20230321150513.php
@@ -20,14 +20,14 @@ final class Version20230321150513 extends AbstractMigration
     public function up(Schema $schema): void
     {
         // this up() migration is auto-generated, please modify it to your needs
-        $this->addSql('ALTER TABLE restaurant ADD max_orders_range_duration INT DEFAULT NULL');
-        $this->addSql('ALTER TABLE restaurant ADD max_orders_amount INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE restaurant ADD rate_limitr_range_duration INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE restaurant ADD rate_limit_amount INT DEFAULT NULL');
     }
 
     public function down(Schema $schema): void
     {
         // this down() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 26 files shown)


translations/messages.fr.yml
```diff
@@ -152,6 +152,8 @@ restaurant.form.cuisines.label: Styles de cuisine
 restaurant.deliveryPerimeterExpression.title: Périmètre de livraison
 restaurant.deliveryPerimeterExpression.help: Définit le périmètre de livraison accepté.
   Vous pouvez choisir un rayon (en km) ou une zone
+restaurant.ordersRateLimiter.title: Gestion des limites de commande
+restaurant.ordersRateLimiter.help: La fonction de limite de commande vous permet de contrôler le nombre de commandes que vous pouvez accepter sur une période de temps donnée. En définissant une limite, vous évitez de recevoir plus de commandes que ce que vous pouvez gérer. Il vous suffit de spécifier le nombre maximal de commandes et la durée de la période de temps. Les commandes supplémentaires sont automatiquement reportées sur le créneau suivant. Cette fonctionnalité vous aide à mieux gérer votre capacité de production et à éviter les retards et les erreurs.
```
alexsegura, line 156: Nice 🙂

app/DoctrineMigrations/Version20230321150513.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE restaurant ADD max_orders_range_duration INT DEFAULT NULL');
```
alexsegura, line 23: Maybe we should name it `rate_limit_range_duration`/`rate_limit_amount`?

src/Utils/OrdersRateLimit.php
```diff
@@ -0,0 +1,209 @@
...
+        if (!$this->featureEnabled($order)) {
+            return false;
+        }
+
+        // Parse order and business parameters
+        $params = $this->getOrdersLimitParameters($order);
+
+        // Calculate time window
+        $start_time = $pickupTime->getTimestamp();
+        $end_time = $start_time + ( $params['max_orders_range_duration'] * 60 );
```
alexsegura, line 37: Not sure it's the good (or intuitive) way to calculate the time window. 

For example, if we have configured "10 orders per 30 minutes", I would expect to have "stable" time windows, like 10:00-10:30, 10:30:11:00... 

If we calculate the time window like this, it means that if we have a pickup at 10:26, it would generate a time window of 10:26-10:56, i.e a "sliding" time window.

app/DoctrineMigrations/Version20230321150513.php
r0xsh: Yep ! Totally agree, I changed the name of the feature in the middle of the day :laughing:

src/Form/Restaurant/FulfillmentMethodsTrait.php
```diff
@@ -25,7 +26,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'allow_add' => false,
                 'allow_delete' => false,
                 'prototype' => false,
-            ]);
+            ])
+            ->add('ordersRateLimiter', HiddenType::class);
```
alexsegura, line 30: This should be available only for users with `ROLE_ADMIN`

src/Utils/OrdersRateLimit.php
alexsegura: After discussing with @r0xsh and @ERAMANrepartos, we agreed that a _sliding_ time window works.

app/DoctrineMigrations/Version20230321150513.php
r0xsh: Fixed with [a720d93](https://github.com/coopcycle/coopcycle-web/pull/3584/commits/a720d93b3820697dc439a03b28c24b7bda62b485)

src/Form/Restaurant/FulfillmentMethodsTrait.php
r0xsh: Oh yeah thanks, good catch :)
[4e08035](https://github.com/coopcycle/coopcycle-web/pull/3584/commits/4e080352f1922625b5007ebe6c9e36d4f2536d54)

alexsegura (approved):

PR #3589, opened 2023-03-28
New tasks export using Cube

### LINKED ISSUE #3045: Improve UX for exports: one single export with all relevant information
- [ ] @Titooo94 Provide an example spreadsheet of what is expected

Right now, there are 2 places in which you can export information from the platform:

## Export of tasks from the dashboard

- Task number
- Type
- Address Name
- Address
- GPS Coordinates
- Address description
- After
- Before
- Status 
- Comments 
- Eventdone notes 
- Eventfailed notes 
- Finished at 
- Courier 
- Tag 
- Address contact name 
- Organisation

## Export of deliveries

- Organisation 
- Pick up address 
- Pick up after 
- Drop Off Address 
- Drop off before 
- Weight 
- Distance 
- Order Number 
- Order total 
- Packages

It looks like it is also missing an alias to identify deliveries. 

Having all important information in one single export would be very helpful

lucasferraro:
Fixes #3045

### DIFF
--- a/src/Form/TaskExportType.php
+++ b/src/Form/TaskExportType.php
@@ -2,26 +2,30 @@
 
 namespace AppBundle\Form;
 
-use AppBundle\Domain\Task\Event;
-use AppBundle\Entity\Task;
-use AppBundle\Entity\TaskRepository;
+use AppBundle\CubeJs\TokenFactory as CubeJsTokenFactory;
+use AppBundle\Service\Geocoder;
 use League\Csv\Writer as CsvWriter;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\DateType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormEvents;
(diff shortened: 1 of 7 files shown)


src/Form/TaskExportType.php
```diff
@@ -46,66 +50,127 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
-                if ($task->hasEvent(Event\TaskDone::messageName())) {
-                    $finishedAt = $task
-                        ->getLastEvent(Event\TaskDone::messageName())
-                        ->getCreatedAt()->format('Y-m-d H:i:s');
-                } else if ($task->hasEvent(Event\TaskFailed::messageName())) {
-                    $finishedAt = $task
-                        ->getLastEvent(Event\TaskFailed::messageName())
-                        ->getCreatedAt()->format('Y-m-d H:i:s');
-                }
+                $address = $this->geocoder->geocode($resultObject['TasksExportUnified.addressStreetAddress']);
```
alexsegura, line 149: We already have the coordinates in the `geo` field, we shouldn't need to use the geocoder for this.

PR #3593, opened 2023-03-30
Show tours as folders on TaskLists

### LINKED ISSUE #3591: Route should appear as a "folder" when assigned
See #3552

jimakker:
Fixes #3591

### DIFF
--- a/js/app/dashboard/components/TaskList.js
+++ b/js/app/dashboard/components/TaskList.js
@@ -15,10 +15,13 @@ import {
 import classNames from 'classnames'
 
 import Task from './Task'
+import Tour from './Tour'
+
 import Avatar from '../../components/Avatar'
+
 import { unassignTasks, togglePolyline, optimizeTaskList } from '../redux/actions'
 import { selectVisibleTaskIds } from '../redux/selectors'
-import { makeSelectTaskListItemsByUsername } from '../../coopcycle-frontend-js/logistics/redux'
+import { makeSelectTaskListItemsByUsername, makeSelectOrganizedTaskListItemsByUsername } from '../../coopcycle-frontend-js/logistics/redux'
 
 moment.locale($('html').attr('lang'))
 
@@ -83,7 +86,11 @@ class TaskList extends React.Component {
       isEmpty,
     } = this.props
 
-    const { tasks } = this.props
+    const { 
+      tasks,
+      standaloneTasks, 
+      tours,
+    } = this.props
 
     const uncompletedTasks = _.filter(tasks, t => t.status === 'TODO')
     const completedTasks = _.filter(tasks, t => t.status === 'DONE')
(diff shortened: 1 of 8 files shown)


js/shared/src/logistics/redux/selectors.js
```diff
@@ -72,3 +72,54 @@ export const makeSelectTaskListItemsByUsername = () => {
...
+        return []
+      }
+
+      const items = taskList.itemIds
+        .filter(id => Object.prototype.hasOwnProperty.call(tasks, id)) // a task with this id may be not loaded yet
+        .map(id => tasks[id])
+
+        const standaloneTasks = _.filter(items, task => !belongsToTour(task))
+
+        const toursMap = new Map()
```
jimakker, line 96: Can/should be refactored as this code is already used in another place

js/shared/src/logistics/redux/selectors.js
```diff
@@ -72,3 +72,54 @@ export const makeSelectTaskListItemsByUsername = () => {
...
+          tours.push({
+            ...tour,
+            items: tasks,
+          })
+        })
+    
+        return {
+          standaloneTasks,
+          tours,
+          allTasks: items
```
jimakker, line 121: We are returning all tasks *again*. Should `standaloneTasks` and `tours` be arrays of ids referencing the tasks in `allTasks`?

`allTasks` is needed to show info about delivered percentage etc

alexsegura (commented):
There's a strange behavior when reordering tasks in the assigned section. It's ok to have tours appearing first as we can't do much better, but it seems impossible to reorder standalone tasks properly. 

tour_reorder

js/app/dashboard/components/TaskList.js
```diff
@@ -177,8 +184,30 @@ class TaskList extends React.Component {
                 }) }
                 { ...provided.droppableProps }
               >
+                { _.map(tours, (tour, index) => {
+                  return (
+                    <Draggable key={ `tour-${tour['@id']}` } draggableId={ `tour:${tour['@id']}` } index={ this.props.tours.length + index }>
```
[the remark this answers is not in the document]
alexsegura: The indexes should be unique for each draggable, this is what explains the strange dragging behavior. 
It's fixed in 29bac4b51d9773212824c2d5aa87f38f79335a4b

<table>
<tr><td>Without unique indexes</td><td>With unique indexes</td></tr>
<tr><td>
</td>
<td>
</td></tr>
</table>

alexsegura:
We have a problem: we are always rendering tours **above** other tasks, while it's still possible to drag'n'drop them. It generates an odd behavior, where you drag'n'drop a tour, and it would always go back to the top (I didn't even check why it shows (7) during a few seconds). 

tour_odd_behavior

We are not going to change the database design right now, however it should be possible to order a tour after/before standalone tasks. 

<table width="100%">
<tr><td>How it looks like in the UI</td><td>How it looks like in Redux</td></tr>
<tr>
<td><pre>
Task 1
Task 2
Tour 1
  Task 3
  Task 4
  Task 5
Task 6
Task 7
Tour 2
  Task 8
  Task 9
  Task 10
</pre></td>
<td>
<pre>
Task 1
Task 2
Task 3
Task 4
Task 5
Task 6
Task 7
Task 8
Task 9
Task 10
</pre></td>
</tr>
</table>

So basically, the selector should return a "flat" list of items, which tasks and tours "mixed" altogether. 
It's not as simple as creating an array of tasks belonging to the same tour, but it should be doable.

PR #3602, opened 2023-04-24
Avoid duplicated tasks in export

lucasferraro:
Some tasks have more than one `task_collection_item` associated. 
Now we are avoiding duplication of tasks sorting by `task.id` and `task_collection_item.position` and then for each task duplicated we keep the first one (the one with minimum position value in the collection).


src/Form/TaskExportType.php
```diff
@@ -118,7 +117,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             $csv->insertOne([
                 '#',
                 '# order',
-                'orderNumber',
+                'orderCode',
```
alexsegura, line 120: Noticing this now, but it would be better to name it `order.number`

PR #3607, opened 2023-04-27
Allow to create order from delivery in API context

r0xsh:
Fixes: #3608


r0xsh:
@alexsegura i'm pretty sure that the fix will works, but i got the following error. I guess its related to bad test data and wont happens in production env
image

PR #3609, opened 2023-04-27
Send customer to stripe

lucasferraro:
When user wants to pay with a saved credit card we have to send the customer id with the payment intent even if restaurant is connected but doesn't pay stripe fees.


alexsegura (approved):

alexsegura (commented):
Tests need to be fixed

alexsegura (commented):
Tests fixed by rebasing

PR #3697, opened 2023-06-22
Improve errors messages for delivery import

### LINKED ISSUE #1726: In order to faster find errors when uploading stops I want the error message to show the line in which the error occurs
**In order to**  make the error massages even more helpful it would be great to get the information on which line the error occurred to easy fix it. 

**As a** of a dispatcher I would benefit from it by easily finding the reason of a failed upload

**I want** to get the info in which line the problem happens. I wouldn't mind if it's just the first error line. 

**Whereas currently** you don't know the line that is responsible for the failed upload.

lucasferraro:
Closes #1726 
In order to make the error massages even more helpful when import deliveries: 

- [x] Show the multiple errors that are occurring in the import
- [x] Import only the correctly understood rows and not the erroneous rows
- [x] Inform well to the user which rows have been imported and which not

https://github.com/coopcycle/coopcycle-web/assets/4819244/ed0fe9b3-afa9-4fe2-b519-cfa5e2eceab8

**Explanation of new messages after imports**
Captura de pantalla de 2023-06-16 14-59-57

### DIFF
--- a/src/Form/DeliveryImportType.php
+++ b/src/Form/DeliveryImportType.php
@@ -66,14 +66,26 @@ class DeliveryImportType extends AbstractType
                 $deliveries = array_filter($result->getData(), function($delivery, $rowNumber) use ($result) {
                     $violations = $this->validator->validate($delivery);
                     if (count($violations) > 0) {
-                        $result->addErrorToRow($rowNumber, (string) $violations->get(0));
+                        foreach($violations as $violation) {
+                            $result->addErrorToRow($rowNumber,
+                                sprintf('%s: %s', $violation->getMessage(), (string) $violation->getInvalidValue())
+                            );
+                        }
                         return false;
                     }
                     return true;
                 }, ARRAY_FILTER_USE_BOTH);
 
                 if ($result->hasErrors()) {
-                    foreach($result->getSortedErrors() as $rowNumber => $failedRow) {
+                    $errors = $result->getSortedErrors();
+
+                    $event->getForm()->addError(new FormError(
+                        $this->translator->trans('import.unsuccessful.rows', [
+                            '%rows%' => implode(", ", array_keys($errors))
+                        ])
+                    ));
+
+                    foreach($errors as $rowNumber => $failedRow) {
                         $event->getForm()->addError(new FormError(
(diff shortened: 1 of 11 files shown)


r0xsh (approved):
Except for the error messages, which aren't very user-friendly. Everything seems fine to me :)
I'll create an issue for this (which I don't consider very urgent)

src/Form/DeliveryImportType.php
```diff
@@ -58,12 +61,29 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
 
-                $deliveries = $this->spreadsheetParser->parse($file->getPathname());
+                $result = $this->spreadsheetParser->parse($file->getPathname());
 
-                foreach ($deliveries as $delivery) {
+                $deliveries = array_filter($result->getData(), function($delivery, $rowNumber) use ($result) {
                     $violations = $this->validator->validate($delivery);
                     if (count($violations) > 0) {
-                        throw new \Exception((string) $violations->get(0));
+                        $result->addErrorToRow($rowNumber, (string) $violations->get(0));
```
r0xsh, line 69: There is never more than one violation error ?

lucasferraro:
> Except for the error messages, which aren't very user-friendly. Everything seems fine to me :) I'll create an issue for this (which I don't consider very urgent)

Yes, I think you mean to the errors like 'Object(AppBundle\...' which come from the violation errors that are not very user-friendly. I agree about to create a new issue to improve them in a next step.

src/Form/DeliveryImportType.php
lucasferraro: Yes, you're right. I'll change it so we take into account all the errors no just the first one.

lucasferraro:
New version of how errors look. 

- When there are violations from validators now we show all the violations (before we just took the first one)
- Messages from violations now are more user-friendly with a clear message and the invalid value (we still need to figure out how to communicate in a clear way the field in the row that is causing the error)
- At the beginning of the errors section now we specify which rows could not be processed.

Captura de pantalla de 2023-06-23 11-57-49

r0xsh:
Nice work :tada:

PR #3711, opened 2023-07-03
Allow tagging customers

### LINKED ISSUE #3684: In order to identify special clients during service, I want to be able to associate tags or flags with clients
To improve customer service I want to be able to identify users as 
1. VIP
2. ProblemClient
3. WeSkrewedUp

In such a way that later, when they create a new order, this special status can be communicated to the relevant users. 

**As a** `role`
Everyone

In a user´s profile page I want a clickable button that can identify them as a VIP, PC, or WSU
I DO NOT WANT THEM TO SEE THIS INFORMATION IN THEIR PROFILE

Also, should these categories be modifiable by the cooperative? If so, if would be easy to use the existing tags feature and the cooperative makes a tag associated with them.

HOWEVER then the visualization of those types of "Flags" (flags alert you to important information in real time that you might miss otherwise, tags are an organizational tool) must be different. maybe a bar instead of a circle. 

Nothing, we cannot do this

**Additional context**
image

AwesomeKuro:
Fixes #3684


src/Entity/User.php
```diff
@@ -62,8 +64,11 @@
  * @UniqueEntity(fields={"usernameCanonical"}, errorPath="username")
  * @UniqueEntity("facebookId")
  */
-class User extends BaseUser implements JWTUserInterface, ChannelAwareInterface
+class User extends BaseUser implements TaggableInterface
```
alexsegura, line 67: Actually, we want to tag customers, not users. A person ordering always maps to a customer, but not necessarily to a user.

docker/php/php.ini
```diff
@@ -1,5 +1,5 @@
 apc.enable_cli = 1
-date.timezone = Europe/Paris
+date.timezone = America/Monterrey
```
alexsegura, line 2: This shouldn't be commited

templates/admin/users.html.twig
```diff
@@ -101,6 +101,7 @@
 <thead>
   <th>{% trans %}profile.email{% endtrans  %}</th>
   <th>{% trans %}profile.fullName{% endtrans  %}</th>
+  <th>{% trans %}profile.tags{% endtrans  %}</th>
```
alexsegura, line 104: You can use `adminDashboard.tags.title` instead, it's already translated

app/config/services/twig.yml
```diff
@@ -16,6 +16,10 @@ services:
     arguments:
       $assetsFilesystem: '@assets_filesystem'
 
+  AppBundle\Twig\TagsRuntime:
+    tags:
+        - { name: twig.tagsruntime }
+
```
alexsegura, line 22: This is useless, the service is autowired by Symfony.

templates/profile/profile.html.twig
```diff
@@ -3,7 +3,15 @@
...
 </div>
+      
 
+{% if (user.isSuperAdmin) %}
+  <p><strong>{% trans %}profile.tags{% endtrans %}</strong>:
+    {% for tag in user.customer.tags|expand_tags %}
+      <span class="label label-default" style="background-color: {{ tag.color }};">{{ tag.slug }}</span>
+    {% endfor %}
+  </p>
+{% endif %}
```
alexsegura, line 14: - It's useless for admins to know about their own tags
- We are not using the "super admin" role in any case

src/Entity/Model/TaggableTrait.php
```diff
@@ -20,25 +23,35 @@ public function getTaggableResourceClass()
...
     public function getTags()
     {
         if (null === $this->tags) {
             $this->tags = [];
         }
-
+        $log = new Logger('getTags');
+        $log->pushHandler(new StreamHandler('php://stdout', Logger::WARNING)); // <<< uses a stream
+        $log->warning('getTags');
+        $log->warning(print_r($this->tags, true));
```
alexsegura, line 36: Please remove

src/Entity/Model/TaggableTrait.php
```diff
@@ -20,25 +23,35 @@ public function getTaggableResourceClass()
...
      * @SerializedName("tags")
-     * @Groups({"task_create", "task_edit"})
+     * @Groups({"task_create", "task_edit", "admin_user_edit", "update_profile"})
      */
     public function setTags($tags)
     {
+        $log = new Logger('setTags');
+        $log->pushHandler(new StreamHandler('php://stdout', Logger::WARNING)); // <<< uses a stream
+        $log->warning('setTags');
+        $log->warning(print_r($tags, true));
```
alexsegura, line 49: Please remove logging code

src/Form/UpdateProfileType.php
```diff
@@ -146,6 +155,44 @@ function (FormEvent $event) use ($options, $isAdmin) {
...
+            //$log->warning(var_dump($user));
+            //$log->warning(var_dump($event));
+            $tags = $event->getData();
+            //$tags = explode(' ', $tags);
+
+            $user->setTags($tags);
+            //$user->getUser()->setTags($tags);
+        });
+
+        */
```
alexsegura, line 195: Please remove dead code

templates/profile/profile.html.twig
```diff
@@ -4,6 +4,7 @@
   </a>
 </div>
 
+<p><strong>{% trans %}profile.tags{% endtrans %}</strong>: {{ user.tags|join(', ', ' and ') }}</p>
```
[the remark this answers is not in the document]
AwesomeKuro: Can i use an if to check if the user is an administrator?

AdrienSICKLO:
image

This client is a VIP customer. The tag appear on the foodtech dashboard, but does not appear on the dispatch screen.

image

PR #3732, opened 2023-08-08
Apply pricing rules for deliveries with multiple tasks.

### LINKED ISSUE #3047: Support multiple tasks when calculating prices
With the multistop option, it has become important to create pricing rules that can match this feature.

A distance condition is not relevant as the client is not aware of what the fastest route could be.

Therefore, a new condition based on the number of points is important.

There should be the possibility to add a "discount" like with the packages, as in "First point is 9, and other points will be 5"

alexsegura:
Fixes #3047

This is the simplest implementation possible, without modifying the pricing editor. 
When using the "All the matching rules" method of calculation AND having multiple tasks, the system will iterate over all the tasks, and try to apply the rules. 

⚠️ There are still some limitations. 

- Some conditions don't make sense at the task level, for example the distance. 

- Also, timing conditions won't work as expected when using "less than" operator. It is because when using the "Difference (hours)" condition, it translates to an expression like `diff_hours(pickup) < 3`. The current implementation returns `-1` when the task is not of the expected type (i.e, when evaluating a dropoff), so it will still match the "less than" operator. A solution would be to change the generated expression, to avoid using an operator, and use a boolean. For example it would generate an expression like `time_diff(pickup, "<", 3)`.

### DIFF
--- a/src/Entity/Delivery.php
+++ b/src/Entity/Delivery.php
@@ -417,11 +417,8 @@ class Delivery extends TaskCollection implements TaskCollectionInterface, Packag
     {
         $taskObject = new \stdClass();
         if ($task) {
-            $taskObject->address = $task->getAddress();
-            $taskObject->createdAt = $task->getCreatedAt();
-            $taskObject->after = $task->getAfter();
-            $taskObject->before = $task->getBefore();
-            $taskObject->doorstep = $task->isDoorstep();
+
+            return $task->toExpressionLanguageObject();
         }
 
         return $taskObject;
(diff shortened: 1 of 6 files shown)


Paul-Eraman-CoopCycle:
To be clear:
1. km based pricing rules do not work
2. We will now change the generated expression to the new solution, and time range based rules will function correctly
3. there are NOT other pricing rules that do not work with this?

Of course, we now need to implement a check that
**When** a user is modifying a shops configuration and has "multipoint deliveries" and "make orders" selected, it checks if the pricing rules associated with the shop contain KM rules, and if so, has a pop up alert that explains the problem, asks you to change the rules, and does not allow you to save. Correct?

Paul-Eraman-CoopCycle:
Does not work in demo:
Unless "create orders" is deactivated, the option to add another dropoff does not appear in the order creation page

image
image
image

PR #3759, opened 2023-09-05
Notify restaurante/dispatcher when an order is delayed

### LINKED ISSUE #3081: Notify restaurant when the dispatcher delays an order and vice versa if restaurant delays an order
_Originally posted by @AdrienSICKLO in https://github.com/coopcycle/coopcycle-web/issues/525#issuecomment-852805321_

lucasferraro:
Closes #3081 

When an order is delayed a new notification will be shown notifying to restaurant when the dispatcher delays an order and vice versa if restaurant delays an order. This notification will be visible only if the user is or has a tab open in the orders dashboard (admin or restaurant).

https://github.com/coopcycle/coopcycle-web/assets/4819244/5f6fb2c3-6cf7-4c10-bfa6-def3fa960f4c

### DIFF
--- a/js/app/foodtech/dashboard/redux/middlewares.js
+++ b/js/app/foodtech/dashboard/redux/middlewares.js
@@ -8,6 +8,8 @@ import {
   orderRefused,
   orderCancelled,
   orderFulfilled,
+  orderDelayed,
+  ORDER_DELAYED,
   ORDER_CREATED,
   initHttpClient,
   INIT_HTTP_CLIENT,
@@ -48,6 +50,9 @@ export const socketIO = ({ dispatch, getState }) => {
           break
         case 'order:fulfilled':
           dispatch(orderFulfilled(event.data.order))
+          break
+        case 'order:delayed':
+          dispatch(orderDelayed(event.data.order))
           break
       }
     })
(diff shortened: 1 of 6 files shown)


alexsegura (commented):
I think there should also be the same notification/alert on the dispatch dashboard.

r0xsh (approved):

js/app/foodtech/dashboard/redux/middlewares.js
```diff
@@ -143,6 +148,19 @@ export const notification = ({ getState }) => {
       }
     }
 
+    if (action.type === ORDER_DELAYED) {
```
alexsegura, line 151: Would be better at some point to generalize this behavior server-side, by marking PHP events with something, so that the fronted knows which event should trigger this type of notification. But it's ok for now 🙂

alexsegura (approved):

Paul-Eraman-CoopCycle:
can we add the name of the restaurant to the pop up to facilitate finding the order later?

lucasferraro:
> can we add the name of the restaurant to the pop up to facilitate finding the order later?

I couldn't capture in the video but at the top of the notification is the name of the restaurant.

PR #3776, opened 2023-09-25
Customize notifications in settings screen

lucasferraro:
With these changes administrators are able to enable/disabled the notifications they want to be notified for. The enable/disabled of a notification is global to the instance, which means that the enabled notifications will be shown for all the users that should be notified and viceversa for disabled notifications. 

How it looks:

https://github.com/coopcycle/coopcycle-web/assets/4819244/34ac0de8-80c8-49a8-9058-43a94842e0f7


app/DoctrineMigrations/Version20230920192807.php
```diff
@@ -0,0 +1,59 @@
...
+            'task:started',
+            'task:done',
+            'task:failed',
+            'task:cancelled',
+            'task_collections:updated',
+            'task_list:updated',
+            'task:rescheduled',
+            'task_import:success',
+            'task_import:failure'
+        ];
```
alexsegura, line 46: This should be stored in PHP in enum-style classes like `OrderEvents`/`TaskEvents`

src/Service/LiveUpdates.php
```diff
@@ -76,6 +81,10 @@ public function toOrderWatchers(OrderInterface $order, $message, array $data = [
     {
         $messageName = $message instanceof NamedMessage ? $message::messageName() : $message;
 
+        if (!$this->shouldNotifyEvent($messageName)) {
```
alexsegura, line 84: The live updates should still be pushed, or the UI won't update automatically. 
It's more at the email / top bar notifications level that this needs to apply.

app/DoctrineMigrations/Version20230920192807.php
```diff
@@ -0,0 +1,59 @@
...
+final class Version20230920192807 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('CREATE TABLE notification (id SERIAL NOT NULL, name VARCHAR(64) NOT NULL, enabled BOOLEAN DEFAULT \'true\' NOT NULL, PRIMARY KEY(id))');
```
alexsegura, line 22: A problem with this implementation, is that it would need the table to be filled on a fresh CoopCycle instance. 
Migrations are not executed on a fresh instance, so it would need to be filled in the `SetupCommand`
IMHO it would be better to not rely on the data in the table.

AdrienSICKLO:
Hey,
Would it be possible to add to the list the "delay bar activation"?

alexsegura:
> Would it be possible to add to the list the "delay bar activation"?

@AdrienSICKLO this is not currently an event. Do you mean that you want to log when people move the delay bar? If yes, please open a new issue.

AdrienSICKLO:
I don't know what you mean by log. But yes, having a possibility to know afterwards if/when the delay has been activated would be nice to analyse dispatch

alexsegura:
Log means having a trace somewhere. For example, in the activity page.

alexsegura (approved):

PR #3792, opened 2023-10-06
Failure reason

r0xsh:


src/Controller/AdminController.php
```diff
@@ -1096,6 +1098,117 @@ public function duplicatePricingRuleSetAction($id, Request $request)
...
+            ->getRepository(Delivery\FailureReasonSet::class)
+            ->find($id);
+
+        return $this->renderFailureReasonSetForm($failureReasonSet, $request);
+    }
+
+    /**
+     * @Route("/admin/deliveries/failures/{id}/delete", methods={"POST"}, name="admin_failures_delete")
+     */
+    public function deleteFailureReasonSetAction($id)
```
alexsegura, line 1188: Nice, we don't usually implement the delete action 🙂

src/Action/Task/FailureReasons.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function __construct(
+        private EntityManagerInterface $em,
+        private TranslatorInterface $translator
+    )
+    { }
+
+    private function getDefaultReasons(): array
+    {
+        $path = realpath(__DIR__ . '/../../Resources/config/failure_reasons.yml');
```
alexsegura, line 23: Could be cached to improve performance

rayon9coop: If a package is reported as damaged, a photo must be attached to prove the damage.
Generally speaking, it might be interesting (as a dispatch) to be able to configure the situations in which the courier can be forced to attach a photo to prove the event encountered (absent customer -> photo of the front door to prove that he has shown up, for example).
See what urbantz propose for that kind of situations
image

Issue #3789, opened 2023-10-05
Deleting tasks from admin dashboard has effect in accounting of coopcycle


Paul-Eraman-CoopCycle:
https://coopcycle.slack.com/archives/C4UT4DGQ3/p1696506098731159?thread_ts=1694603761.025129&cid=C4UT4DGQ3

Paul-Eraman-CoopCycle:
Quick fix for this is to cancel orders when **ALL** the tasks that make up the order are cancelled.

alexsegura:
Should be easy to implement with a Doctrine listener. 

When a task changes its status to `CANCELLED`, check if it is associated with an order. If it is associated to an order, check if all the tasks are `CANCELLED`, and if yes, cancel the order.

PR #3834, opened 2023-10-30
Cancel order when all linked tasks are cancelled

r0xsh:
Fixes: #3789


alexsegura (changes requested):
Can you please add a migration that retro-actively cancels all the orders that have all their tasks cancelled?

alexsegura (approved):

PR #3845, opened 2023-11-07
Price recalculation

r0xsh:


src/Enum/TaskCollectionState.php
```diff
@@ -0,0 +1,51 @@
...
+
+    function toGlyph() {
+        return match ($this) {
+            TaskCollectionState::PENDING => '⏲️',
+            TaskCollectionState::IN_DELIVERY => '🚴‍♂️',
+            TaskCollectionState::DELIVERED => '✅',
+            TaskCollectionState::FAILED => '⚠️',
+            TaskCollectionState::CANCELLED => '❌'
+        };
+    }
```
alexsegura, line 20: Dead code

src/Entity/TaskCollection.php
```diff
@@ -128,11 +129,21 @@ public function removeTask(Task $task)
...
-            return $item->getTask();
-        })->toArray();
+        return $this->getItems()
+            ->map(fn($item) => $item->getTask())
+            ->toArray();
+    }
+
+    public function getActiveTasks()
+    {
+        return collect($this->getTasks())
```
alexsegura, line 144: Where does the `collect` function come from? 
I would prefer to use the already imported `ArrayCollection` from Doctrine, which also has a `filter` method.

src/Entity/TaskCollection.php
```diff
@@ -128,11 +129,21 @@ public function removeTask(Task $task)
...
     {
-        return $this->getItems()->map(function (TaskCollectionItem $item) {
-            return $item->getTask();
-        })->toArray();
+        return $this->getItems()
+            ->map(fn($item) => $item->getTask())
+            ->toArray();
+    }
+
+    public function getActiveTasks()
```
alexsegura, line 142: Maybe it would be more reusable & "expressive" to implement this via a `filterTasks` method, that accepts an expression language function. For example, it would be: 

```php 
$delivery->filterTasks('not task.cancelled');
```

PR #3872, opened 2023-11-14
Use base32 for order numbers

### LINKED ISSUE #3829: Rethink order codes to reduce chance of similar codes
With high volume we get confusing codes. For some reason, there is a significant amount of chronology in the way the codes are created. 

I would propose that instead of the 1st two digits almost being 100% chronological, randomizing all 4 letter/digits and removing letters that often can be confusing. these would include the letters i, L, e, o, v or u, q, p or f, and m or n, and leave all the numbers

instead of getting what is seen in the photo below, we should have more like

3W10
7N8B
5RNU

etc

image

lucasferraro:
Closes #3829

### DIFF
--- a/src/Entity/Listener/OrderListener.php
+++ b/src/Entity/Listener/OrderListener.php
@@ -51,8 +51,12 @@ class OrderListener
         }
     }
 
-    private function findNumberWithoutCollision(EntityManagerInterface $objectManager, string $number)
+    private function findNumberWithoutCollision(EntityManagerInterface $objectManager, string $number, int $depth = 10)
     {
+        if ($depth <= 0) {
+            throw new \Exception('Could not find a unique number');
+        }
+
         $orderRepository = $objectManager->getRepository(Order::class);
         $orderWithSameNumber = $orderRepository->findOneByNumber($number);
 
@@ -62,6 +66,6 @@ class OrderListener
 
         $nextId = $orderRepository->fetchNextSeqId();
         $newNumber = Crockford::encode($nextId);
-        return $this->findNumberWithoutCollision($objectManager, $newNumber);
+        return $this->findNumberWithoutCollision($objectManager, $newNumber, $depth - 1);
     }
 }
--- a/src/Entity/Listener/OrderListener.php
+++ b/src/Entity/Listener/OrderListener.php
@@ -4,9 +4,10 @@ namespace AppBundle\Entity\Listener;
 
 use AppBundle\Entity\Sylius\Order;
 use AppBundle\Entity\Vendor;
-use Doctrine\Persistence\Event\LifecycleEventArgs;
-use Doctrine\ORM\Event\PreUpdateEventArgs;
+use Dflydev\Base32\Crockford\Crockford;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Event\PreFlushEventArgs;
+use Doctrine\ORM\Event\PreUpdateEventArgs;
 
 class OrderListener
 {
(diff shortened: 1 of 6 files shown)


src/Entity/Listener/OrderListener.php
```diff
@@ -32,4 +33,35 @@ public function preFlush(Order $order, PreFlushEventArgs $args)
...
+        $orderWithSameNumber = $orderRepository->findOneByNumber($number);
+
+        if (null === $orderWithSameNumber) {
+            return $number;
+        }
+
+        $nextId = $orderRepository->fetchNextSeqId();
+        $newNumber = Crockford::encode($nextId);
+        return $this->findNumberWithoutCollision($objectManager, $newNumber);
+    }
```
r0xsh, line 66: Always use safety vest in the presence of recursive functions :safety_vest: :repeat: 

```suggestion
    private function findNumberWithoutCollision(EntityManagerInterface $objectManager, string $number, int $depth = 10)
    {
        if ($depth <= 0) {
            throw new \Exception('Could not find a unique number');
        }

        $orderRepository = $objectManager->getRepository(Order::class);
        $orderWithSameNumber = $orderRepository->findOneByNumber($number);

        if (null === $orderWithSameNumber) {
            return $number;
        }

        $nextId = $orderRepository->fetchNextSeqId();
        $newNumber = Crockford::encode($nextId);
        return $this->findNumberWithoutCollision($objectManager, $newNumber, $depth - 1);
    }
```

src/Entity/Sylius/OrderRepository.php
```diff
@@ -245,4 +245,21 @@ public static function addVendorClause(QueryBuilder $qb, $alias, LocalBusiness $
             ->andWhere(sprintf('%s.restaurant = :restaurant', $vendorAlias))
             ->setParameter('restaurant', $restaurant);
     }
+
+    public function fetchNextSeqId(){
+        $dbConnection = $this->getEntityManager()->getConnection();
+        $nextValQuery = $dbConnection->getDatabasePlatform()->getSequenceNextValSQL('sylius_order_id_seq');
```
r0xsh, line 251: Is the sequence value updated in the table? I'm not entirely sure if I've grasped it correctly, but from what I gather, there seems to be a cycle of events:

For instance:
1. If Order `#100` encounters a collision with an existing number,
2. You fetch the next ID using `getSequenceNextValSQL()` and get `101`.
3. Subsequently, you store `Crockford::encode(101)` as the number for Order `#100`.
4. However, when Order `#101` is generated, won't its number collide with Order `#101` and so forth?

r0xsh (changes requested):
Recursion go brrrrr

src/Entity/Sylius/OrderRepository.php
```diff
@@ -245,4 +245,21 @@ public static function addVendorClause(QueryBuilder $qb, $alias, LocalBusiness $
...
+
+    public function findOneByNumber(string $number): ?OrderInterface
+    {
+        $qb = $this->createQueryBuilder('o');
+        $qb
+            ->andWhere('o.number = :number')
+            ->setParameter('number', $number);
+
+        return $qb->getQuery()->getOneOrNullResult();
+    }
```
alexsegura, line 264: This shouldn't be needed to declare this method, this getter is [already exposed magically by Doctrine](https://github.com/doctrine/orm/blob/609647a51a1b28a674de270204a5cea3af57d554/lib/Doctrine/ORM/EntityRepository.php#L275-L277).

src/Entity/Sylius/OrderRepository.php
lucasferraro: Sorry, I didn't include information about how this works (info that I didn't know either until I need to implement this by the way).
The `nextval` function in PostgreSQL advances the sequence to its next value and returns that value. So in that example would be:

1. Order #100 encounters a collision with an existing number
2. Fetch next ID available (`101`) and advance the sequence
3. Store Crockford::encode(101) as the number for Order #100.
4. Next Order will be generated with the ID 102, the Order with 101 won't exist in the platform since we advance the sequence manually and use that ID to resolve the collision.

r0xsh: Okay perfect then ! That was my question / concern :+1:

src/Entity/Sylius/OrderRepository.php
lucasferraro: That's magic is great! I thought that the suggestion was to use the `findOneBy` method

r0xsh (approved):

PR #3911, opened 2023-12-05
Fix/guest order tracking

vladimir-8:


alexsegura:
@vladimir-8 can you check why tests are failing? Thanks

app/config/security.yml
```diff
@@ -9,14 +9,20 @@ security:
         Nucleos\UserBundle\Model\UserInterface: sha512
 
     role_hierarchy:
-        ROLE_CUSTOMER: ROLE_USER
```
alexsegura, line 12: You removed the role because it's not used anywhere?

PR #3915, opened 2023-12-07
fix: check if error value is printable

r0xsh:


alexsegura:
Seems tests are passing now 🤷‍♂️

r0xsh:
yeah i'm so confused :confused:

PR #3917, opened 2023-12-07
Implement personal information and business account settings page

### LINKED ISSUE #3871: As a business account manager I want to have a configuration screen where I can update the information of the account
In order to have a way to edit:
- Business account settings
- Employees attached to the account
- Regenerate/Share invitation link

**As a** `role`
Administrador of a business account (`ROLE_BUSINESS_ACCOUNT`)

For a business account manager screens where can see/edit the account settings and the employees attached to the account.

This does not exist

**Additional context**
[Design for these screens](https://www.figma.com/file/l62gsQss0WpmiPhKQ6gufH/Foodtech?type=design&node-id=256-14988&mode=design&t=9kOA3udSQfqYH9tu-4)

lucasferraro:
Closes #3871 

New designs for personal and company (only business account managers) information.

https://github.com/coopcycle/coopcycle-web/assets/4819244/d5a36a5a-f113-4993-bf7e-b9299b8e048d

### DIFF
--- a/templates/_partials/profile/invitation_link.html.twig
+++ b/templates/_partials/profile/invitation_link.html.twig
@@ -1,0 +1,41 @@
+<div class="form rounded-bottom">
+  {% if title is defined %}
+    <h5>{{ title }}</h5>
+  {% endif %}
+  <div class="form-group">
+    <label class="control-label">{{ field_label(form.invitationLink) }}</label>
+    <div class="input-group">
+    <span class="input-group-addon bg-primary"><i class="fa fa-link fa-rotate-90 text-white"></i></span>
+    {{ form_widget(form.invitationLink) }}
+    </div>
+    <div class="d-flex justify-content-end mt-3">
+    <button id="copy-button" type="button" class="btn btn-primary" data-toggle="copy" data-clipboard-text="{{ field_value(form.invitationLink) }}">
+      <i class="fa fa-copy mr-2"></i>
+      {{ 'registration.step.invitation.copy.button'|trans }}
+    </button></div>
+  </div>
+  {% if flow is defined %}
+    {% include '_partials/profile/flow_buttons.html.twig' %}
+  {% endif %}
+</div>
+
+{% block scripts %}
+  <script>
+    document.getElementById('copy-button').addEventListener('click', function(e) {
+      const iconClass = this.childNodes[1].className
+      const text = this.childNodes[2].nodeValue
+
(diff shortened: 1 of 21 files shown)


lucasferraro:
@alexsegura I think `change-password` action is broken in our version of `nucleos/user-bundle`. If you try to change a password (through the web or command `php bin/console nucleos:user:change-password`) the password in database is not updated.
[There's already a possible fix for this](https://github.com/nucleos/NucleosUserBundle/pull/691) but in our case is not working as well.

In version `3.0` of the bundle [they made a rework on this](https://github.com/nucleos/NucleosUserBundle/pull/703) and the implementation is totally different. Should we update the version we're using?

I'll continue trying to figure out how to fix this, but just to let you know that seems like currently in the platform the `change-password` action is not working :/

alexsegura:
> Should we update the version we're using?

@lucasferraro yes, if possible, let's use the latest version. Can you make another pull request with only the bump, so that we merge it before this PR? Thanks.

lucasferraro:
> @lucasferraro yes, if possible, let's use the latest version. Can you make another pull request with only the bump, so that we merge it before this PR? Thanks.

Ok sure. I'll bump the version and test in my local. If everything work as expected I'll make another PR with that.

templates/_partials/profile/invitation_link.html.twig
```diff
@@ -0,0 +1,41 @@
...
+
+{% block scripts %}
+  <script>
+    document.getElementById('copy-button').addEventListener('click', function(e) {
+      const iconClass = this.childNodes[1].className
+      const text = this.childNodes[2].nodeValue
+
+      this.childNodes[1].className = 'fa fa-check mr-2'
+      this.childNodes[2].nodeValue = "{{ 'registration.step.invitation.link.copied'|trans }}"
+
```
r0xsh, line 30: I guess you forgot to trigger the copy event, there is a easy API for that.
- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText
- https://caniuse.com/mdn-api_clipboard_writetext

r0xsh (approved):

templates/_partials/profile/invitation_link.html.twig
alexsegura: Actually, the copy event is triggered by [ClipboardJS](https://clipboardjs.com/)

alexsegura (approved):

Issue #3877, opened 2023-11-17
Refactor Hubs into Platform Catering rules for a Business Account


Paul-Eraman-CoopCycle:
## Existing Hubs
Right now a hub has the following features:

1. General settings 
Main features to configure (not day to day)

- Name of the Hub
- Address of the Hub
- Enabled toggle
- List of associated restaurants (shops)

2. Fulfillment
Operational day to day settings

- Delivery and collection hours for fulfillment
- Delivery and collection minimum cart
- Delivery and collection time range duration
- Delivery and collection delay before ordering
- Delivery and collection toggle preording
- Delivery and collection toggle allow editing 
- Delivery and collection number of days in the future an order can be made in

3. Settlement
How the pricing works

- For restaurant and client if variable amount is being charged
- For restaurant and client the fixed price to charge
- For restaurant and client dropdown menu of pricing rules to use
- For restaurant and client flat delivery fee
- For restaurant and client fee for takeaway
- XXX **no ability to chose who is responsible for Stripe payment**

---

## Platform Catering Use Case for Administrators
📐   General: Modify address to the address of the **dropoff** where the business is located, rather than the pickup address of the hub
➕  Fulfillment: Add a cutoff time 
➖  Fulfillment: Remove "allow editing" button and feature
❓   How to design interface where a restaurant´s menu can be different, with different products and prices? My first thought is to create new products in the restaurant settings page and add them to a new menu, which could then be associated with the hub. This way we maintain the workflow for a restaurant. The administrator of a platform catering hub would need to be able to select the menu of a restaurant when adding a restaurant to the hub, and a restaurant would need to be able to have non public names for its menu items (when making a menu i need to know which item is which, but if i make one item sandwich normal and the other sandwich hub, the final client will see the name sandwich hub)

## Platform Catering Use Case for Business Owners
Does the business owner even need to see this?

Paul-Eraman-CoopCycle:
I've tested and it looks good. ive asked someone else to test too, with that ill move to closed unless something comes up

PR #3947, opened 2023-12-22
Linking business accounts with restaurant group and orders

lucasferraro:
Implementation of new relationships between business accounts, restaurants and orders. Introducing a new concept called `LocalBusinessGroup` which is an abstraction extended by `Hubs` and a new entity called `BusinessRestaurantGroup`.

The idea behind this is to use the new entity `BusinessRestaurantGroup` to group all the restaurants associated to a Business Account, instead of using the existing `Hub` entity which is not the same concept for business accounts.

The idea is to implement the [following design](https://excalidraw.com/#room=8e7bee2792e0d375fde6,iyiZEILWcmooq02lafIrvw)

----------------------------------------------------------------------------------------------
With this PR admin users are able to create and configure  `BusinessRestaurantGroup`, the actions available in the form are related to [this issue](https://github.com/coopcycle/coopcycle-web/issues/3877).

https://github.com/coopcycle/coopcycle-web/assets/4819244/0d1f11bf-4eb6-4429-8b68-c0019bd1159b

Also, as mentioned [in the same issue](https://github.com/coopcycle/coopcycle-web/issues/3877) with this PR admin users are able to select the menu of a restaurant when adding a restaurant to the `BusinessRestaurantGroup`.

https://github.com/coopcycle/coopcycle-web/assets/4819244/0823a4cb-a6ce-4b0f-ab8d-6193bd9fe7c0


src/Controller/RestaurantController.php
```diff
@@ -360,6 +362,43 @@ public function hubAction($id, $slug, Request $request,
...
+
+        if ($redirectToCanonicalRoute) {
+
+            return $this->redirectToRoute('business_restaurant_group', [
+                'id' => $id,
+                'slug' => $expectedSlug,
+            ], Response::HTTP_MOVED_PERMANENTLY);
+        }
+
+        return $this->render('restaurant/business_restaurant_group.html.twig', [
```
alexsegura, line 396: This template does not exist.

app/DoctrineMigrations/Version20231117160815.php
```diff
@@ -32,9 +32,6 @@ public function down(Schema $schema): void
         $this->addSql('ALTER TABLE hub DROP CONSTRAINT FK_4871CE4D5BC85711');
         $this->addSql('DROP INDEX UNIQ_4871CE4D5BC85711');
         $this->addSql('ALTER TABLE hub DROP business_account_id');
-        $this->addSql('ALTER TABLE restaurant ADD business_account_id INT DEFAULT NULL');
-        $this->addSql('ALTER TABLE restaurant ADD CONSTRAINT fk_eb95123f5bc85711 FOREIGN KEY (business_account_id) REFERENCES business_account (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
-        $this->addSql('CREATE INDEX idx_eb95123f5bc85711 ON restaurant (business_account_id)');
```
alexsegura, line 37: This migration should/may already have been executed, better not modify it, or migration `Version20231219182856` will not work, because it will try to drop this column.

src/Controller/RestaurantController.php
```diff
@@ -360,6 +362,45 @@ public function hubAction($id, $slug, Request $request,
         ]);
     }
 
+    /**
+     * @Route("/business_restaurant_group/{id}-{slug}", name="business_restaurant_group",
```
alexsegura, line 366: I'm not sure this URL is actually useful.

app/DoctrineMigrations/Version20231117160815.php
alexsegura: Sorry, I didn't notice those changes were in the `down` method.

PR #3952, opened 2024-01-05
Incident rework

r0xsh:
## Split Incidents from task failures.
New endpoint PUT `/api/tasks/:id/incidents` to register incident on a task.

## Update incident view in order timeline
Incident code is shown
Comment is shown with a bubble icon
image

## Add new filter for Incidents
New `Tasks with incident reports` filter
New `Only see these` button to easily switch between views

image

## Update the progress bar and tooltip for each rider
The progress bar now have multiple colors to reflect each states
The tooltip show more details about the rider's state.

image


Atala:
- the progress bar is wrong, it mixes failed/succesfull statuses and incidents. these should be separated
- the field `failureReason` is "wrong" because
       - a task can have several incidents
       - a task can be not "failed/failure", i.e. a success and have an incident
 -> migrate old "failureReason" data to TaskEvent. maybe add a boolean field on the Task so we can know front end if the task has an event.

Atala (changes requested):
see my comments on the PR

good luck !

src/Domain/Task/Handler/RescheduleHandler.php
```diff
@@ -25,7 +25,7 @@ public function __invoke(Reschedule $command)
         $task->setBefore($rescheduledBefore);
         $task->unassign();
         $task->setMetadata('rescheduled', true);
-        $task->setFailureReason(null);
+        $task->setIncidented(false);
```
Atala, line 28: I am surprised by this change : 
1. why should we override the previous "incident statute" if the task was rescheduled? task had an incident and it should be kept 
2. also, I guess rescheduling should create an incident as well, so we can keep track of it. I think it will nice but not 100% sure what do you think ? what @Paul-Eraman-CoopCycle  do you think about registering reschedule as incident ?

src/Entity/Task.php
```diff
@@ -438,10 +448,10 @@ class Task implements TaggableInterface, OrganizationAwareInterface, PackagesAwa
...
 
     /**
-     * @var string|null
-     * @Groups({"task"})
-     */
-    private $failureReason;
+    * @var bool
+    * @Groups({"task"})
+    */
+    private $incidented = false;
```
Atala, line 454: sorry to say that but I think it is a poor name, as it doesn't mean anything, it should be `hasIncident` or `hasIncidents`

Atala (changes requested):
- `incidented` should be changed to `hasIncident` or `has Incidents`
- discussion about rescheduling is an incident or not - alos I don't think we should erase `hasIncident/incidented` there. because we want to keep the history on this task

features/tasks.feature
```diff
@@ -39,7 +39,7 @@ Feature: Tasks
             "assignedTo":"bob",
             "previous":null,
             "group":{"@*@":"@*@"},
-            "tags":@array@,
+            "tags"@array@,
```
Atala, line 42: typo ?

src/Domain/Task/Handler/RescheduleHandler.php
r0xsh: The actual behavior of the `hasIncidents` fields is to say if the task is in a "error"/incident state. Rescheduling a Task remove this state to avoid polluting the dashboard with task that has been reschuduled for another day.
This is not optimal, but QA-Coops seems happy with this behavior for now. Things will change later with a full incident feature.

r0xsh: The `hasIncidents` field indicates whether a task is in an error state or has encountered an incident. To prevent the dashboard from being cluttered with tasks that have been rescheduled for a different day, rescheduling a task will reset this field, removing the error/incident state.
This is not optimal, but QA-Coops seems happy with this behavior for now. Things will change later with a full incident feature.

r0xsh: Rescheduling tasks are recorded but not categorized as incidents; they are treated as consequences of incidents. For instance, if a customer is not at home, the rider may report a `NO_CUSTOMER` incident. Subsequently, the dispatcher can address this incident by rescheduling the task for a later time.

Issue #3929, opened 2023-12-13
Sometimes an order has a delivery fee applied twice


vladimir-8:
**Describe the bug**

For some orders I see multiple `delivery` fee items in the `sylius_adjustment` table.

For example, see orders #173396, 175444, 176052 on the eraman instance.

**To Reproduce**
Unknown. Creating this issue as a placeholder to collect more information.

**Expected behavior**
As I understand normally there should be only one delivery fee per order.

vladimir-8:
I suspect there is a race condition between some endpoints. I encountered this bug from the app (API calls). 


```
{
"@context":"\/api\/contexts\/ConstraintViolationList",
"@type":"ConstraintViolationList",
"hydra:title":"An error occurred",
"hydra:description":"adjustments: Unexpected number of \u0022delivery\u0022 adjustments\nadjustments: Unexpected number of \u0022fee\u0022 adjustments",
"violations":[
{"propertyPath":"adjustments","message":"Unexpected number of \u0022delivery\u0022 adjustments","code":null},
{"propertyPath":"adjustments","message":"Unexpected number of \u0022fee\u0022 adjustments","code":null}]}
```


Here are the endpoints that were called before it happened:

![Screenshot 2024-01-09 at 18 08 01](https://github.com/coopcycle/coopcycle-web/assets/70273239/92eda19d-2743-4b69-938f-bff9ae2f0354)

![Screenshot 2024-01-09 at 18 10 16](https://github.com/coopcycle/coopcycle-web/assets/70273239/417ff035-e435-4f73-9ff3-5f6fd7666fda)

I noticed an issue with the app, that it calls `/assign` endpoint twice when a user logs in during the checkout flow. Still, I think that the backend should be able to handle such incorrect calls.

vladimir-8:
One case when it happens: https://github.com/coopcycle/coopcycle-web/pull/4561

Paul-Eraman-CoopCycle:
can we query how often this happens @vladimir-8 ?

Paul-Eraman-CoopCycle:
wow thats super annoying. do we know how to fix it?

vladimir-8:
I know where to start investigating, but we still have to find a place that needs a fix

Issue #3873, opened 2023-11-15
Display bug : order total products and number of items does not match what the customer actually paid for


tcmjlnt:
**Describe the bug**
For instance 2ORQ (#125414) at Les Coursiers Rennais

Display is wrong but the customer paid the right price.
However it is problematic because the restaurant doesn’t always check the total amount and can prepare more items than paid for and end up with a net loss.
It’s the second time in 2 days. I don’t know if any one else had this bug.
Order displays 3 pizzas when the customer ordered and paid  2. The restaurant prepared 3 pizza.

**To Reproduce**
Idk

**Expected behavior**
correct display

**Screenshots**
![admin 2ORQ](https://github.com/coopcycle/coopcycle-web/assets/101735534/7197f7e6-1875-492b-95b3-a2856f33cbc7)
<img width="1078" alt="CSV 2ORQ" src="https://github.com/coopcycle/coopcycle-web/assets/101735534/6786e0e4-eaa8-4945-945e-318218bf0eae">
![customer confirmation 2ORQ](https://github.com/coopcycle/coopcycle-web/assets/101735534/4b706452-6b38-468a-b446-5802591c3ed5)
![customer receipt 2ORQ](https://github.com/coopcycle/coopcycle-web/assets/101735534/a3fadf71-1c5e-44e2-898b-7c9497a112b9)
<img width="1280" alt="restaurant order 2ORQ" src="https://github.com/coopcycle/coopcycle-web/assets/101735534/908addc6-d809-40a2-9a09-cf8d72122d0f">
<img width="1280" alt="stripe payment 2ORQ" src="https://github.com/coopcycle/coopcycle-web/assets/101735534/efc8f3dc-72b1-44f1-a045-586abede25e5">

**Desktop (please complete the following information):**
 n/a

**Smartphone (please complete the following information):**
 - n/a

**Additional context**

See attached slack thread here :
https://coopcycle.slack.com/archives/CQDUL440P/p1699869814791059

Paul-Eraman-CoopCycle:
Is this a frequent bug?

Paul-Eraman-CoopCycle:
I would close this due to how old it is, but messing with the money is bad. @vladimir-8 do we have any way of looking at logs to see if this is still occuring? if not, i guess we close until we see it happen again?

Paul-Eraman-CoopCycle:
wait its closed, but still a good question haha

PR #3965, opened 2024-01-12
Add more logs

vladimir-8:
Add more logs in an attempt to catch issues leading to:

https://github.com/coopcycle/coopcycle-web/issues/3873

https://github.com/coopcycle/coopcycle-web/issues/3929

https://github.com/coopcycle/coopcycle-app/issues/1603

### DIFF
--- a/src/Service/LoggingUtils.php
+++ b/src/Service/LoggingUtils.php
@@ -19,7 +19,10 @@ class LoggingUtils
          * (1) | function: persistAndFlushCart | file: /var/www/html/src/Controller/RestaurantController.php | line: 639
          * (2) | function: addProductToCartAction | file: /var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php | line: 163
          */
-        return $this->getCallerAtFrame(2);
+
+        $frameNumber = 2 + 1; // 2 frames from this function + 1 frame from getCallerAtFrame
+        
+        return $this->getCallerAtFrame($frameNumber);
     }
 
     public function getOrderId($order): string {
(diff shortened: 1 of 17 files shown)


alexsegura (commented):
This is a lot of changes, and it is quite invasive just to add logs.

src/Service/LoggingUtils.php
alexsegura, line 1: Why is `LoggingUtils` actually a service? It doesn't need to be instantiated, and there is nothing injected inside of it, it could be a helper class with just static methods. This would avoid having to inject an instance of `LoggingUtils` everywhere.

vladimir-8:
> This is a lot of changes, and it is quite invasive just to add logs.

I agree. I also don’t like that, in some cases, there are more logs than the code. But what other approach could we use? 

Basically, I want to have a way to know when each element of the total price was added. So that when we have a situation like this https://coopcycle.slack.com/archives/C4UT4DGQ3/p1704279294849429 we can understand what went wrong.

Let me try to use doctrine event listeners; maybe I can use them instead of some logs.

src/Service/LoggingUtils.php
vladimir-8: Indeed, a helper class would also work here. Just if it's injected, it's easier to mock/disable in the unit tests. Not sure if it's possible with the static methods.

vladimir-8:
@alexsegura I moved some of the logs to the event listener: https://github.com/coopcycle/coopcycle-web/pull/3965/files#diff-870004e7df38010bf25350e656d9f8a919cb6697f901bfc98f7ec2410ec0a23f Please take a look again

alexsegura (approved):

PR #3967, opened 2024-01-12
Improve delivery imports.

### LINKED ISSUE #3957: Improve deliveries import
There are multiple problems when importing deliveries. Let's fix them all. 

### 1. Spreadsheets should be processed asynchronously

Currently they are processed synchronously, so if the geocoder is slow, it may generate a HTTP 504 error. 
Files should be processed asynchronously, just like in the dispatch dashboard, so that we get rid of potential timeout errors. 

### 2. Error reporting should be easier to understand

Currently, when there are import errors, it looks like this: 

Capture d’écran 2024-01-09 à 18 06 27

---

Instead, errors should be shown as a spreadsheet (for example using [Luckysheet](https://dream-num.github.io/LuckysheetDocs/guide/#introduction)), with problematic rows/cells highlighted in red. 

As a bonus, the user should be able to download a partial spreadsheet, with the lines that could not be imported.

alexsegura:
Fixes #3957

- Handle import asynchronously.
- Show a list of recent imports.
- Show errors line per line with a spreadsheet view.

delivery_import_async


Paul-Eraman-CoopCycle:
This doesn't have the ability for me to edit the cell when it is visuallised there directly? and the it reuploads?

alexsegura:
> This doesn't have the ability for me to edit the cell when it is visuallised there directly? and the it reuploads?

No, it would require too much work. 
For now, as suggested in the issue, we will allow to re-download a spreadsheet containing only the rows that could not be imported and need to be fixed.

PR #3971, opened 2024-01-16
remove an order validation after creating a payment intent

vladimir-8:
The same behaviour already exists in the API: https://github.com/coopcycle/coopcycle-web/blob/9faa2d465fe560b9d006694f9aab6cbda7fe3618/src/Action/Order/Pay.php#L51


lucasferraro (approved):
LGTM

Issue #3861, opened 2023-11-13
In order to [set up an additional option] as a [restaurant/admin] I want [the default max value to be set to 1]


tcmjlnt:
**In order to**  `set up an additional option`
When checking the box "This option is additional" in a restaurant's options.

**As a** `restaurant/admin`
Especially when restaurants are setting up options, they are not always aware that the default max value is infinite.

**I want** `the default max value to be set to 1`
More often than not, the max value for an additional is 1, not infinite. For instance you can choose additional toppings for your pizza, you only choose one. When it's more than 1, for instance steaks for a burger, it's never gonna be infinite. So the default value should be the lowest >0 = 1. If the restaurants want the customer to add more than 1 steak, it will be more obvious to change in that way, than from infinite to 3 for instance.

**Whereas currently** `this happens`
Max default value is infinite.

**Additional context**

<img width="1280" alt="Capture d’écran 2023-11-13 à 16 08 30" src="https://github.com/coopcycle/coopcycle-web/assets/101735534/3102ffbd-773c-4793-97e1-817a1151420c">
<img width="1280" alt="Capture d’écran 2023-11-13 à 16 09 38" src="https://github.com/coopcycle/coopcycle-web/assets/101735534/e3d74f82-72c0-4fc0-bce4-28cd7fb09af7">

PR #3972, opened 2024-01-17
set the default max value for the additional option to 1

vladimir-8:
Fixes: https://github.com/coopcycle/coopcycle-web/issues/3861


Atala:
image


hello some small remarks on the product option form :
- the "no max value" checkbox should be at the same level than the number inputs
- ticking the checkbox should hide the "max value" input not disable it
- once the product is saved, when you untick "max value" then value in the "max" input is INF, should be 1 no?

max-value

Atala (approved):
see my comment on the PR

I guess you have a lot on your plate; so if it takes you too long ignore the comment and merge directly :)

vladimir-8:
Thanks for the review! The idea was to make a quick fix, so indeed I don't want to spend a lot of time on the various improvements.

But this seems directly related to the original issue. I'll implement it:

> once the product is saved, when you untick "max value" then value in the "max" input is INF, should be 1 no?

vladimir-8:
> once the product is saved, when you untick "max value" then value in the "max" input is INF, should be 1 no?

@Atala I implemented this. Please take a look

Atala (changes requested):
If i go from INFINITE to 1, I am under the impression it is not correctly set. See GIF below. correct me if i am wrong!

I am in holyday next week, so please fix and merge directly! 

max-value-not-set

vladimir-8:
It happens because the menu page is cached. If you want to test it immediately you can temporary remove the cache element from https://github.com/coopcycle/coopcycle-web/blob/57da3a4974cc03750905f867b0767a0ace6df62f/templates/restaurant/index.html.twig#L194

Atala:
OK thanks I didnt know that, sorry for the false alert

PR #3979, opened 2024-01-23
Upgrade to React 18

vladimir-8:


assets/react/controllers/NavBarCart.jsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react'
+import CartTop from '../../../js/app/cart/CartTop'
+
+export default function NavBarCart(props) {
```
Atala, line 4: - why is this change needed?
- should we add this for all the components we attach to the dom ?

webpack.config.js
```diff
@@ -76,11 +76,32 @@ Encore
...
+    resolve: {
+      fullySpecified: false,
+    },
+  })
+
   .enableStimulusBridge('./assets/controllers.json')
 
-  .enableSingleRuntimeChunk()
+  // When enabled, Webpack "splits" your files into smaller pieces for greater optimization.
   .splitEntryChunks()
```
Atala, line 97: I am curious, does the order between `splitEntryChunks` and `enableSingleRuntimeChunk changes anything ?

Atala (approved):
I have been poking around and everything seems to work allright

i just have 50 of these warnings in webpack container logs, maybe handle them now before merge?
```
coopcycle-web-webpack-1  | /srv/coopcycle/js/app/widgets/DateTimePicker.js
coopcycle-web-webpack-1  |   2:10  warning  ReactDOM.render is deprecated since React 18.0.0, use createRoot instead, see https://reactjs.org/link/switch-to-createroot  react/no-deprecated
``

vladimir-8:
> i just have 50 of these warnings in webpack container logs, maybe handle them now before merge?
> 
> ```
> coopcycle-web-webpack-1  | /srv/coopcycle/js/app/widgets/DateTimePicker.js
> coopcycle-web-webpack-1  |   2:10  warning  ReactDOM.render is deprecated since React 18.0.0, use createRoot instead, see https://reactjs.org/link/switch-to-createroot  react/no-deprecated
> ``
> ```

I tried, but I run into the issue that the new concurrent render introduced in react 18 could create race conditions with jquery when they modify the same elements. That's why I reverted these changes and propose to migrate gradually when we work on each page and can dedicate more time for testing

assets/react/controllers/NavBarCart.jsx
vladimir-8: It's an experiment of using Symfony\UX\React\ReactBundle to render react components in twig templates instead of adding them in scripts: https://github.com/coopcycle/coopcycle-web/pull/3979/files#diff-540f9bd9a739c50f0821fcd03e7894fcf56f3a8e0a22acbe7423092cfd33d093R77-R81

For now, twig will only pickup components from `assets/react/controllers/` that's why this wrapper is needed. Later on we can set it up to look for components in `/js/app`, but it might have consequences on the code splitting

webpack.config.js
vladimir-8: I don't think so. I just added comments and put them in the same order as in the docs: https://github.com/symfony/symfony-docs/blob/deca2b1acdb09e15432b0240731209207f8e5880/frontend/encore/installation.rst

PR #3989, opened 2024-01-29
Tours : enable dragndrop, modifications, ... in assigned tours

Atala:
This PR adds :
- remove task from assigned tour by click on the cross
- add task to assigned tour by dragndrop
- move "groups of tasks" in the "unassigned tours" column (this column will be renamed soon) 
- make sure we dont create a group twice when hitting twice the "add" button in the create group modal

Known limitation : when you add a task to an assigned tour it closes the tour -> I will fix this soon

tour-modification


js/app/dashboard/components/Tour.js
```diff
@@ -1,45 +1,184 @@
...
+              </div>
+            )}
+          </Droppable>
       </div>
     </div>
   )
 }
 
-export default withTranslation()(Tour)
+function mapStateToProps (state) {
```
vladimir-8, line 167: That could be replaced with `useSelector` and `useDispatch` hooks in the function components. https://react-redux.js.org/api/hooks

js/app/dashboard/components/Tour.js
```diff
@@ -1,45 +1,184 @@
...
+
+    return {
+        isDroppable: state.logistics.ui.areToursDroppable,
```
vladimir-8, line 170: It's better to access state via a selector function

js/app/dashboard/components/Tour.js
```diff
@@ -1,45 +1,184 @@
...
+                                onClick={ e => e.preventDefault() }>
+                                <i className="fa fa-trash"></i>
+                              </a>
+                          </Popconfirm>
+                     }            
+                </div>
+            </>
+            }
+            { 
+              toggleInputForName && renderEditNameForm()
```
vladimir-8, line 118: It seems that `renderEditNameForm()` should be a function component

js/app/dashboard/components/UnassignedTasks.js
```diff
@@ -142,7 +142,7 @@ class UnassignedTasks extends React.Component {
           <Droppable droppableId="unassigned">
             {(provided) => (
               <div className="list-group nomargin" ref={ provided.innerRef } { ...provided.droppableProps }>
-                { _.map(this.props.groups, (group, index) => {
+                { !this.props.toursEnabled && _.map(this.props.groups, (group, index) => {
```
vladimir-8, line 145: I suggest using a conditional (?) rather than && to avoid subtle bugs; https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

js/app/dashboard/redux/actions.js
```diff
@@ -286,14 +290,21 @@ export function importError(token, message) {
   return { type: IMPORT_ERROR, token, message }
 }
 
-export function modifyTaskList(username, tasks) {
+export function modifyTaskList(username, tasks, updateTourUI=null) {
```
vladimir-8, line 293: Could you please explain why do we need to pass `updateTourUI` callback and can not execute it before dispatching this action?

js/app/dashboard/redux/actions.js
```diff
@@ -309,26 +320,30 @@ export function modifyTaskList(username, tasks) {
...
-        // eslint-disable-next-line no-console
-        console.error(error)
-      })
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error)
+    }
+
+    // return a promise so we can chain after dispatching this function
+    return new Promise((resolve) => {
```
vladimir-8, line 343: There is no need to create a Promise explicitly. As it's an async function JS will wrap returned result into a promise

js/app/dashboard/redux/actions.js
```diff
@@ -1593,41 +1471,54 @@ export function createTour(tasks, name, date) {
...
+          'Content-Type': 'application/ld+json'
+        }
+      })
+    } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error(error)
+        dispatch(modifyTourRequestError(tour))
+    }
+
+    return new Promise((resolve) => {
```
vladimir-8, line 1511: Same as earlier. There is no need to create a Promise explicitly. As it's an async function JS will wrap returned result into a promise

js/app/dashboard/redux/actions.js
```diff
@@ -309,26 +320,30 @@ export function modifyTaskList(username, tasks) {
...
         position,
       }
     })
 
     dispatch(modifyTaskListRequest(username, newTasks))
 
-    axios
-      .put(url, data, {
+    try {
+      var response =  await axios.put(url, data, {
```
vladimir-8, line 331: Better use `let` instead of `var`

js/app/dashboard/redux/handleDrag.js
vladimir-8, line 1: If you have time, it would be great to add tests for this file

js/app/dashboard/components/Tour.js
```diff
@@ -1,45 +1,184 @@
...
+import { getDroppableListStyle } from '../utils'
+
+
+const Tour = ({ tour, removeTaskFromTour, unassignTasks, isDroppable, modifyTour, deleteTour }) => {
 
-const Tour = ({ tour, tasks, username = null, unassignTasks = null }) => {
 
-  const { t } = useTranslation()
+  const { t } = useTranslation(),
+        { show } = useContextMenu({
```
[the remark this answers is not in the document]
Atala: OK I'll change it 
I post a message on Slack so we all agree on this? or we should redact guidelines somewhere?

js/app/dashboard/components/Tour.js
```diff
@@ -1,45 +1,184 @@
...
+          id: 'dashboard',
```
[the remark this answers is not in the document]
Atala: yes I think so, it was preparatory work to have 3 different context menus, one for tasks, one  for groups and one for tours.
I will change it now for tasks and the other context menus will be introduced in the coming weeks

js/app/dashboard/components/Tour.js
```diff
@@ -1,45 +1,184 @@
...
+        }),
+        collapseId = `tour-panel-${tour['@id'].replaceAll('/', '-')}`,
+        [toggleInputForName, setToggleInputForName] = useState(false),
+        [tourName, setTourName] = useState(tour.name),
+        onEditSubmitted = async (e) => {
+          e.preventDefault()
+          $('.task__draggable').LoadingOverlay('show', {image: false})
```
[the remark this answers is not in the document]
Atala: yes definitely, it is a bad habit of the old me :D I have changed this code on Friday in order to use redux state to control tour panel (was necessary because of refresh), I will cherry pick this commit

js/app/dashboard/components/Tour.js
Atala: OK I ll look into that, I am not very familiar with hooks yet but we need now to change everywhere for hooks I guess

js/app/dashboard/components/Tour.js
Atala: OK, I saw it was kinf od "ok" for small UI vars but I ll change

js/app/dashboard/redux/handleDrag.js
Atala: yes indeed they are [here](https://github.com/coopcycle/coopcycle-web/blob/tour-v1/js/app/dashboard/redux/__tests__/actions.test.js#L63-L192)

js/app/dashboard/redux/actions.js
Atala: Oh ok I didnt know that thx

js/app/dashboard/components/Tour.js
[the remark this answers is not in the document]
vladimir-8: Sure, we can discuss it on slack. Not the most important thing, to be honest. If you prefer that style it's fine to keep it

js/app/dashboard/components/Tour.js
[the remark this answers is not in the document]
vladimir-8: Ideally, yes. Not sure if we get time for this, though. For now, I would at least avoid using it in the new code

js/app/dashboard/components/Tour.js
vladimir-8: Ideally, yes. But the same as the previous one, not sure if we get time for this. For now, I would at least avoid using it in the new code

js/app/dashboard/redux/handleDrag.js
vladimir-8: Great, maybe better to move them into a separate file?

js/app/dashboard/components/UnassignedTasks.js
```diff
@@ -161,7 +161,7 @@ class UnassignedTasks extends React.Component {
                       )}
                     </Draggable>
                   )
-                })}
+                }) : <></>}
```
vladimir-8, line 164: `null` should also be fine

js/app/dashboard/index.js
```diff
@@ -130,6 +130,21 @@ function start() {
     }
   }
 
+  // the empty tour panels are initially open
+  let expandedToursIds = []
+  tours.forEach((tour) => {if (tour.itemIds.length == 0) {expandedToursIds.push(tour['@id'])}})
+
```
vladimir-8, line 136: Also a matter of style preferences. If I had to do it, I would write it like this:

```
const expandedToursIds = tours
.filter((tour) => tour.itemIds.length == 0)
.map((tour) => tour['@id'])
```

js/app/dashboard/index.js
```diff
@@ -130,6 +130,21 @@ function start() {
...
+    ...preloadedState,
+    logistics: {
+      ...preloadedState.logistics,
+      ui: {
+        ...preloadedState.ui,
+        expandedTourPanelsIds: expandedToursIds
+      }
+    }
+  }
+
```
vladimir-8, line 147: I think it could be made a bit simpler using lodash merge https://lodash.com/docs/4.17.15#merge

```
_.merge(preloadedState, {
  logistics: {
    ui: {
      expandedTourPanelsIds: expandedToursIds
    }
  }
})
```

vladimir-8 (approved):

PR #4000, opened 2024-02-08
Restaurant menu redesign

vladimir-8:
PLEASE REVIEW, BUT DO NOT MERGE (as it's a large change we need to notify cooperatives in advance)

Also, it's might be easier to review storybook-related changes separately in https://github.com/coopcycle/coopcycle-web/pull/4002 (they are included in both PRs)

Fixes:

- https://github.com/coopcycle/coopcycle-web/issues/3980
- https://github.com/coopcycle/coopcycle-web/issues/4006


assets/css/styles.scss
```diff
@@ -26,6 +27,34 @@ h1, h2, h3, h4, h5, h6 {
   vertical-align: middle;
 }
 
+.button-reset {
```
Atala, line 30: why was it added? was it specifically needed?

js/app/restaurant/components/Order/CartTotal.js
```diff
@@ -103,43 +103,35 @@ class CartTotal extends React.Component {
...
   }
-
 }
 
 function mapStateToProps (state) {
 
   return {
     items: selectItems(state),
     itemsTotal: selectItemsTotal(state),
-    total: state.cart.total,
```
Atala, line 142: was not used?

Atala:
- it was asked that products and cart have different background color than page, are you going to implement it?

- cart content may disappear on page refresh when cart button finishes loading
cart-disappear-on-page-refresh

Atala:
choose fulfillment modal covers whole page i dont know if it is intended ?

image

Atala:
allergens not shown on list imtem (check on the left of the modal), not sure if intenteded or not?
image

Atala:
bug : product options are not cleared out when you select a product with options then another product

steps to reproduce :
- open a modal for a product with options
- close the modal
- open a modal for another product -> it shows the option for the first product even if product 2 has no options

image

Atala:
if you come back from checkout with "back arrow", the cart doesnt load

nop-cart-comeback

assets/css/styles.scss
vladimir-8: I use it to reset some default button styles for the restaurant menu items. Now I think that it does not need to be a global style. Moved it to the menu.scss

vladimir-8:
> * it was asked that products and cart have different background color than page, are you going to implement it?

I did that, see the light grey background behind the products. Previously it was all white:

Screenshot 2024-03-12 at 11 09 56

Paul-Eraman-CoopCycle:
Thanks @vladimir-8 and valentin, this looks much better this way

vladimir-8:
> * cart content may disappear on page refresh when cart button finishes loading
>       cart-disappear-on-page-refresh
>             

Can you please check what data do you get from the backend/redux state in that case? I suspect that it happens on the backend, because some validations are not met for some reason

vladimir-8:
> choose time modal covers whole page i dont know if it is intended ? it doesnt show on the screenshot but when you visit the page the whole modal background is blank
> 
> image

Not sure I understand what do you mean. I see the same as on your screenshot:

Screenshot 2024-03-12 at 11 06 45

vladimir-8:
> allergens not shown on list item (here behind the modal on the left), not sure if intenteded or not? image

Yes. In the list it shows only diets and zero waste and in the details both diets and allergens

Paul-Eraman-CoopCycle:
Thats ok for now, long term having the little alergen icons in front is probably a good idea

vladimir-8:
> bug : product options are not cleared out when you select a product with options then another product
> 
> steps to reproduce :
> 
>     * open a modal for a product with options
> 
>     * close the modal
> 
>     * open a modal for another product -> it shows the option for the first product even if product 2 has no options
> 
> 
> image

Are you sure that there is no options setup? I can not reproduce it locally. It also works on the test instance: http://206.189.30.248/en/restaurant/51-burger-bar "Make your own burger" and "FRIED CHICKEN STICKS" have options, other products are without options

vladimir-8:
> if you come back from checkout with "back arrow", the cart doesnt load
> 
> nop-cart-comeback 

I suspect that it also might be due to some validation in your local setup. Please check if it works here: http://206.189.30.248/en/restaurant/51-burger-bar

vladimir-8:
from @Atala :

> Screenshot 2024-03-12 at 14 20 08
>     1. price is on the left but for single choice option price is on the right, I guess price should be on the  right or at least always same side

Both options are fine on the desktop. But on the mobile I think the current version is a bit better because the buttons are closer to the side of the screen and are easier to touch. What do you think @Paul-Eraman-CoopCycle ?

Paul-Eraman-CoopCycle:
Agreed, the things you are modifying should be on the right hand side

Atala:
@vladimir-8 I didnt check but do you use the new react 18 `createRoot` to mount all of these components?

Atala:
> > bug : product options are not cleared out when you select a product with options then another product
> > steps to reproduce :
> > ```
> > * open a modal for a product with options
> > 
> > * close the modal
> > 
> > * open a modal for another product -> it shows the option for the first product even if product 2 has no options
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > image
> 
> Are you sure that there is no options setup? I can not reproduce it locally. It also works on the test instance: http://206.189.30.248/en/restaurant/51-burger-bar "Make your own burger" and "FRIED CHICKEN STICKS" have options, other products are without options

sorry mistake on my end, seems fine

Atala (approved):
Hello

I could not reproduce the bugs I was talking about, so your good to go :+1: 
for the white background i messaged you on slack, it is due to my sceen..., very weird I need to check that

I have an issue locally to upload image, so I couldnt test the slider with product images (there is one?), make sure it works!

I think also that you check that the card reset well when switching restaurants, need to fix #4058 first

js/app/restaurant/components/Order/FulfillmentMethod.js
```diff
@@ -8,47 +8,45 @@ const Icon = ({ type }) => (
...
       'fa-bicycle': type === 'delivery',
       'fa-cube': type === 'collection',
-    }) }></i>
-  </span>
-)
+    }) } />
+  </span>)
 
-const FulfillmentMethod = ({ value, shippingAddress, onClick, allowEdit }) => {
+function FulfillmentMethod({ value, shippingAddress, onClick, allowEdit }) {
```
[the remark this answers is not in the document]
vladimir-8: There is a whole discussion about it: https://stackoverflow.com/questions/49306148/why-is-arrow-syntax-preferred-over-function-declaration-for-functional-react-com

I prefer regular functions, mostly due to the examples in the docs: https://react.dev/learn

But there is not significant difference between then. So I should probably kept this component as it was here

js/app/restaurant/components/Order/Time.js
```diff
@@ -51,4 +56,5 @@ function mapDispatchToProps(dispatch) {
   }
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(withTranslation()(Time))
+export default connect(mapStateToProps, mapDispatchToProps)(
```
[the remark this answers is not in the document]
vladimir-8: Non really, just got auto-formatted

PR #4002, opened 2024-02-13
Storybook improvements

vladimir-8:
- [x] Added support for objects in params
- [x] Added storybook for React, available at http://localhost:6007/


Atala (commented):
here is small review; it works on my machine so I think it is alright :
image
1. price is on the left but for single choice option price is on the right, I guess price should be on the  right or at least always same side
2. +/- control dont work, not sure if it is normal? it is also broken on the restaurant page?

image

I dont really understand what is this "header" component?

image

this is broken?

- i didnt llok into detail but ` showPricesTaxExcluded` for CartItem seems weird to me as peperoni has the same price for both options - extra cheese is 0€ then 1€

vladimir-8:
> image
> 
> I dont really understand what is this "header" component?

They are all parts of the product details modal. I tried to split it into smaller components so that they could be mocked/tested separately

Screenshot 2024-03-12 at 14 15 45

vladimir-8:
> image
> 
> this is broken?

It is. I think it was never implemented

vladimir-8:
>     * i didnt llok into detail but ` showPricesTaxExcluded` for CartItem seems weird to me as peperoni has the same price for both options - extra cheese is 0€ then 1€

Fixed now

Atala:
> They are all parts of the product details modal. I tried to split it into smaller components so that they could be mocked/tested separately

OK why  not keeping it if you think it is useful. Maybe call it differently like "Product modal header" ? Or move it outside of this and it becomes our "Header 1" component? (but it has the cross with it, so maybe it is our "Modal header" component?)


> It is. I think it was never implemented

ask Lucas and clean it?

PR #4020, opened 2024-02-22
Implement shipping time calculation for multiple pickups

alexsegura:
Currently, to calculate the shipping time, we calculate the travel time by bike between the pickup address and the dropoff address. 

In the context of business accounts, the pickup address is kind of "random", because it is resolved from the first restaurant of the group

https://github.com/coopcycle/coopcycle-web/blob/d2c6c354408a611cf4d1ed23e0a48701a9196443/src/Entity/BusinessRestaurantGroup.php#L65-L75

This pull request makes a better calculation of the shipping time in this context, and also in the context of multiple pickups in the future: the travel time is calculated using pickup addresses sorted from the farest to the closest, with a final dropoff.

### DIFF
--- a/src/Entity/Sylius/Order.php
+++ b/src/Entity/Sylius/Order.php
@@ -1736,4 +1736,9 @@ class Order extends BaseOrder implements OrderInterface
     {
         return null !== $this->businessAccount;
     }
+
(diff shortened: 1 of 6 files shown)


lucasferraro (approved):
LGTM

PR #4034, opened 2024-02-29
Tours fixes 1

Atala:
This PR introduce the following changes :
 - **validate multiselected tasks : can only select tasks in the same tour and same rider (or unassigned)**
 - **disable right-click context menu when task is in the tour**

And the following small clean-ups
 - highlight task(s) on the right panel when you click marker on the map
 - animated loader on task/tour loading
 - use function based components for UnassignedTours
 - use function based components for TaskList
 - move the <Draggable> component wrapper to child component to avoid repetition (for example <Draggable> is now in <Task> component, not 3 times in UnassignedTasks, Tasklist, Tour)

### DIFF
--- a/js/app/dashboard/components/TaskList.js
+++ b/js/app/dashboard/components/TaskList.js
@@ -143,7 +143,7 @@ export const TaskList = ({ tasks, items, uri, username, polylineEnabled, isEmpty
     unassignTasks(username, task)
   }
 
-  const t = useTranslation()
+  const { t } = useTranslation()
 
   const uncompletedTasks = _.filter(tasks, t => t.status === 'TODO')
   const completedTasks = _.filter(tasks, t => t.status === 'DONE')
--- a/js/app/dashboard/components/TaskList.js
+++ b/js/app/dashboard/components/TaskList.js
@@ -2,7 +2,7 @@ import React from 'react'
 import { connect } from 'react-redux'
 import moment from 'moment'
 import { Draggable, Droppable } from "@hello-pangea/dnd"
-import { withTranslation } from 'react-i18next'
+import { useTranslation } from 'react-i18next'
 import _ from 'lodash'
 import { Tooltip } from 'antd'
 import Popconfirm from 'antd/lib/popconfirm'
@@ -18,7 +18,7 @@ import Task from './Task'
 
 import Avatar from '../../components/Avatar'
 import { unassignTasks, togglePolyline, optimizeTaskList, onlyFilter } from '../redux/actions'
-import { selectFiltersSetting, selectVisibleTaskIds } from '../redux/selectors'
+import { selectVisibleTaskIds } from '../redux/selectors'
 import { makeSelectTaskListItemsByUsername } from '../../coopcycle-frontend-js/logistics/redux'
 import Tour from './Tour'
 import { getDroppableListStyle } from '../utils'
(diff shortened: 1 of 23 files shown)


js/app/dashboard/redux/reducers.js
```diff
@@ -123,50 +125,53 @@ export const polylineEnabled = (state = {}, action) => {
...
+    return [action.task['@id']]
   case SELECT_TASKS:
-
-    return action.tasks.map(task => task['@id'])
-
-  case SELECT_TASKS_BY_IDS:
-
-    return action.taskIds
-
+    newSelectedTasks = action.tasks
```
vladimir-8, line 154: If possible, I suggest saving only task ids instead of the full objects to avoid duplication

js/app/dashboard/redux/actions.js
```diff
@@ -1567,16 +1580,16 @@ export function deleteTour(tour) {
   }
 }
 
-export function removeTaskFromTour(tour, task, username) {
+export function removeTaskFromTour(tour, task, username, unassignTasksAction=unassignTasks, modifyTourAction=modifyTour) {
```
vladimir-8, line 1583: If `unassignTasksAction=unassignTasks, modifyTourAction=modifyTour` are only needed for testing maybe we can find another way to test this function? For example, checking `expect` on the `dispatch` mock?

js/app/dashboard/redux/reducers.js
```diff
@@ -123,50 +125,53 @@ export const polylineEnabled = (state = {}, action) => {
...
     }
-    break
+  break
   }
 
+  if (newSelectedTasks && newSelectedTasks.length > 0) {
+    if (isValidTasksMultiSelect(newSelectedTasks)) {
+      return newSelectedTasks.map(task => task['@id'])
+    } else {
+      toast.warn(i18next.t('ADMIN_DASHBOARD_INVALID_TASKS_SELECTION'), {autoclose: 15000})
```
vladimir-8, line 171: I would avoid touching UI from a reducer, let's move it at least to the action

js/app/dashboard/redux/selectors.js
```diff
@@ -36,7 +36,12 @@ export const selectIsRecurrenceRulesVisible = state => state.settings.isRecurren
 export const selectTaskListGroupMode = state => state.taskListGroupMode
 export const selectSplitDirection = state => state.rightPanelSplitDirection
 export const selectSearchIsOn = state => state.searchIsOn
+export const selectPolylineEnabledByUsername = username => state => state.polylineEnabled[username]
 
+export const getProductNameById = id => store => {
```
vladimir-8, line 41: It seems that it's not used

js/app/dashboard/components/TaskList.js
```diff
@@ -137,199 +115,163 @@ const ProgressBarMemo = React.memo(({
...
 
-class TaskList extends React.Component {
+export const TaskList = ({ uri, username, distance, duration }) => {
+  const dispatch = useDispatch()
+  const unassignTasksFromTaskList = (username => tasks => dispatch(unassignTasks(username, tasks)))(username)
 
-  remove(task) {
-    this.props.unassignTasks(this.props.username, task)
-  }
+  const selectTaskListItems = makeSelectTaskListItemsByUsername()
```
vladimir-8, line 122: I don't really understand why we need to wrap `createSelector` inside a function in `makeSelectTaskListItemsByUsername`. What does not work if we make it a usual selector?

vladimir-8 (approved):

js/app/dashboard/redux/reducers.js
Atala: OK no problem

in this case we need to change `selectedTasks' state key to `selectedTasksIds`

js/app/dashboard/redux/reducers.js
Atala: I put it there because they are 3 different actions touching this slice of state, so I wanted to have this on one place only.

and it is pretty capital that no one breaks this unintentionnaly because i rely on it in other places of the code.

what do you think?

js/app/dashboard/components/TaskList.js
```diff
@@ -137,199 +115,163 @@ const ProgressBarMemo = React.memo(({
     )
   })
 
-class TaskList extends React.Component {
+export const TaskList = ({ uri, username, distance, duration }) => {
+  const dispatch = useDispatch()
+  const unassignTasksFromTaskList = (username => tasks => dispatch(unassignTasks(username, tasks)))(username)
```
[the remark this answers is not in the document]
Atala: I dont know I was worried that `username` would be undefined afterwards when we call the function in the child component
but maybe it is not the case
I will check

js/app/dashboard/redux/reducers.js
Atala: here I need tasks to perform validation on task data before changing the state (I dont return the tasks for state update)
maybe no the good place, let discuss it below

js/app/dashboard/redux/actions.js
Atala: in this particular case I don't remember why I added this as arg because it doesnt seem to be used (???) I will delete it

generally however I think I can only test with `expect` that dispatch has been called with a `function` as argument, but I would like to test the arguments that were passed to this `function`, so I think i need the `function`  to be a mock

is it clear?

js/app/dashboard/redux/reducers.js
vladimir-8: I don't know if it's the best approach, but as we have only ids in the state and you need task objects to perform the validation. I would put all of these inside action creators and make it a function if you need it for multiple actions

vladimir-8: I think it will be more clean rather than adding state into the actions:
```
const currentlySelectedTasks = selectSelectedTasks(getState())
dispatch({ type: TOGGLE_TASK, task, multiple, currentlySelectedTasks })
```

js/app/dashboard/redux/actions.js
vladimir-8: I see. The only other approach that I found is to move it into a separate file and mock the `unassignTasks` and `modifyTour` imports: https://jestjs.io/docs/mock-functions#mocking-modules

js/app/dashboard/redux/reducers.js
Atala: I will do in my next PR

js/app/dashboard/redux/actions.js
Atala: yes  I started with that, I usually do mock modules, and yes if you move stuff to another file it works
but while working on this I realise that dependency injection is much better because it doesnt make you depend on the import logic/order which is a pain because you never  are sure you get the mocked version of the function in your test (like moving stuff or changing the import chain will break the mocking)

i realised that for this project and in general ahah

js/app/dashboard/components/TaskList.js
Atala: because of this : https://redux.js.org/usage/deriving-data-selectors#createselector-behavior

actually if you pass ID directly to the selector you never get the cached result you re-run the selector each time because the selector cache size is one and now you are changing arguments

so it is better to just create a new selector for each different ID/component

it is how far i am understanding, tell me if it is not clear

Issue #3950, opened 2024-01-05
In order to have the correct workflow for incidents "controlled" by dispatch in food delivery; when "failing" tasks it should act as "done" in regards to Stripe


Paul-Eraman-CoopCycle:
In the new workflow of task management, a "failed" task status reflects an incident, which may not mean that it should not be charged. If it should be charged, the dispatcher will review the incident, and apply a reembursement or whatever needs doing. Therefore, an incident by a courier should trigger the capturing of a stripe payment, and if a reemburement needs to happen it can happen somewhere else by the dispatcher.

- [x] failing a pickup should not effect anything else other than creating an incident (should still be possible to do the dropoff)
- [x] failing a dropoff should trigger a stripe capture, just like completing

PR #4036, opened 2024-03-01
Incident V1

r0xsh:
https://github.com/coopcycle/coopcycle/issues/9
https://github.com/coopcycle/coopcycle-web/issues/3950


assets/react/controllers/Incident/List.jsx
```diff
@@ -0,0 +1,231 @@
+import React, { useEffect, useState, useMemo } from "react";
+import { Table, Tag, Avatar, Row, Col, Badge, Switch } from "antd";
+import TaskContext from "./TaskContext";
+import _ from "lodash";
+import "antd/dist/antd.css";
```
alexsegura, line 5: This breaks the whole platform CSS, the fonts are overridden by ant.design. You shouldn't need to require it, or you should require the individual CSS files for Ant components, like in other JS code. 

Capture d’écran 2024-04-23 à 17 26 27

src/Entity/OrganizationRepository.php
```diff
@@ -7,4 +7,18 @@
...
+    {
+        $query = $this->createQueryBuilder('o')
+        ->select('s.id as store_id', 'r.id as restaurant_id')
+        ->leftJoin(Store::class, 's', 'WITH', 's.organization = o.id')
+        ->leftJoin(LocalBusiness::class, 'r', 'WITH', 'r.organization = o.id')
+        ->setMaxResults(1)
+        ->where('o = :org')
+        ->setParameter('org', $org);
+
+        return $query->getQuery()->getResult();
```
alexsegura, line 21: You can use `getOneOrNullResult` to return directly the entity instead of an array.

src/Entity/OrganizationRepository.php
```diff
@@ -7,4 +7,18 @@
 
 class OrganizationRepository extends EntityRepository
 {
+
+    public function reverseFindByOrganizarionID($org)
```
alexsegura, line 11: You could return either a `LocalBusiness` or a `Store` and use `instanceof` in the client code.

src/Action/Task/FailureReasons.php
```diff
@@ -18,36 +21,89 @@ public function __construct(
...
-            'organization' => $org
-        ]);
+        $reverse = $this->em->getRepository(Organization::class)
+            ->reverseFindByOrganizarionID($org);
 
-        if (is_null($store->getFailureReasonSet())) {
+        if (empty($reverse)) {
             return $this->getDefaultReasons();
         }
+        $reverse = $reverse[0];
```
alexsegura, line 92: Using `getOneOrNullResult` in the implementation of `reverseFindByOrganizarionID` would avoid accessing the first element of the array

src/Entity/Incident/IncidentRepository.php
```diff
@@ -0,0 +1,74 @@
...
+                }
+                $tempArray = &$tempArray[$nestedKey];
+            }
+            $tempArray = is_array($value) ? $this->transformArray($value) : $value;
+        }
+
+        return $outputArray;
+    }
+
+    public function getAllIncidents(): array {
```
alexsegura, line 34: At some point if there are lots of incidents, this will die. We should have query limit / pagination here.

r0xsh: Yep i'm aware of that problem but i guess it's okay for now

src/Entity/OrganizationRepository.php
```diff
@@ -4,7 +4,29 @@
...
+    {
+        $query = $this->createQueryBuilder('o')
+        ->select('s', 'r')
+        ->leftJoin(Store::class, 's', 'WITH', 's.organization = o.id')
+        ->leftJoin(LocalBusiness::class, 'r', 'WITH', 'r.organization = o.id')
+        ->setMaxResults(1)
+        ->where('o = :org')
+        ->setParameter('org', $org)
+        ->getQuery()
+        ->getResult();
```
alexsegura, line 23: Why not use `getOneOrNullResult`?

alexsegura (commented):
Why use `*.jsx` extension when everything has the `*.js` extension?

src/Entity/OrganizationRepository.php
r0xsh: When i'm using this method the following exception is raised
> More than one result was found for query although one row or none was expected.

r0xsh:
> Why use *.jsx extension when everything has the *.js extension?

I prefer using `.jsx` file to explicitly tell when a file is a react component. This is helpful when searching in the codebase. I can rename them in `.js` if you prefer

assets/react/controllers/Incident/incidentStore.js
```diff
@@ -0,0 +1,59 @@
...
+  setIncident,
+  setOrder,
+  setImages,
+  setEvents,
+  setTransporterEnabled,
+} = incidentSlice.actions;
+
+export default store;
+
+export const useStore =
```
vladimir-8, line 53: That's a nice trick. Is it to access redux store from the components added with `symfony/ux-react`?

assets/react/controllers/Incident/ActionBox.jsx
```diff
@@ -0,0 +1,157 @@
...
+  );
+}
+
+const styles = {
+  btn: {
+    width: "100%",
+  },
+};
+
+export default function () {
```
vladimir-8, line 25: Should we use `useStore` instead if importing the store directly?

assets/react/controllers/Incident/ActionBox.jsx
```diff
@@ -0,0 +1,157 @@
...
+  const { loaded, incident, order, images, transporterEnabled } =
```
vladimir-8, line 26: Also, it's better to read from the store using selectors, it makes it easier to refactor store state shape

assets/react/controllers/Incident/CommentBox.jsx
```diff
@@ -0,0 +1,58 @@
...
+async function _handleCommentSubmit(id, comment) {
+  const httpClient = new window._auth.httpClient();
+  return await httpClient.post(
+    window.Routing.generate("api_incidents_add_comment_item", { id }),
+    { comment },
+  );
+}
+
+export default function () {
+  const { incident } = store.getState();
```
vladimir-8, line 16: same as above; should we use `useStore` instead if importing the store directly?

assets/react/controllers/Incident/incidentStore.js
vladimir-8: Maybe `useStore` is not a good name, though, because it's not a hook https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level

assets/react/controllers/Incident/List.jsx
```diff
@@ -0,0 +1,234 @@
...
+
+  const showedIncidents = useMemo(() => {
+    if (showClosed) {
+      return incidents;
+    }
+    return _(incidents)
+      .filter((i) => i.status !== "CLOSED")
+      .value();
+  }, [incidents, showClosed]);
+
```
vladimir-8, line 115: it looks like the `incidents` should be saved in redux, and all these `useMemo` should become redux selectors (maybe at some point later 😄)

assets/react/controllers/Incident/OrderDetails.jsx
```diff
@@ -0,0 +1,167 @@
...
+      </h5>
+      <p>{task.address.name}</p>
+      <p>{task.address.streetAddress}</p>
+      <p>{task.address.telephone}</p>
+      {task.weight && <p>{weight(task.weight)}</p>}
+      <div className="mt-3">{<TaskStatusBadge task={task} />}</div>
+      <hr />
+      {order?.customer && <ShowCustomerDetails customer={order.customer} />}
+    </div>
+  );
```
vladimir-8, line 166: I think the word `show` is unnecessary here, it could be simply:
CustomerShowName > CustomerName
ShowOrderDetails -> OrderDetails
ShowCustomerDetails -> CustomerDetails

assets/react/controllers/Incident/TaskContext.jsx
vladimir-8, line 1: I'd call it something like `IncidentItem`, it's just a row in a table, is not it?

assets/react/controllers/Incident/TaskContext.jsx
```diff
@@ -0,0 +1,71 @@
...
+import { money } from "./utils";
+
+async function _fetchTaskContect(id) {
+  const httpClient = new window._auth.httpClient();
+  return await httpClient.get(
+    window.Routing.generate("api_tasks_get_task_context_item", { id }),
+  );
+}
+
+function formatContext({ delivery, order }) {
```
vladimir-8, line 13: that can be a React component, for example `Context` or `ContextDetails`

assets/react/controllers/Incident/TaskTimeline.jsx
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+import TaskTimeline from "../../../../js/app/dashboard/components/TaskTimeline";
+
+export default function ({ events }) {
+  events = Object.values(JSON.parse(events));
```
vladimir-8, line 5: I would not modify props, just in case. Why not create a new variable?

src/Entity/Incident/IncidentEvent.php
vladimir-8, line 1: Is it an `IncidentAction` rather than event?

templates/admin/incident.html.twig
vladimir-8, line 1: There are a few strings here that are not localised

vladimir-8 (changes requested):

src/Entity/OrganizationRepository.php
```diff
@@ -4,7 +4,29 @@
...
+
+        $query = array_filter($query);
+        if (count($query) === 1) {
+            return $query[0];
```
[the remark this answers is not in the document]
vladimir-8: `return array_shift($query);` seems to fix it locally, but I'm not sure if it's a proper solution

vladimir-8:
One more thing: did you test it with the released mobile app version? What will happen if somebody try to mark task as failed?

assets/react/controllers/Incident/incidentStore.js
r0xsh: You're correct, and the name is even more inaccurate with 'store.' You can access the store directly without wrapping the component with that function. It simply enables store reactivity.

We could use:
- connectWithRedux
- reduxWrapper
- reactiveReduxWrapper

assets/react/controllers/Incident/ActionBox.jsx
r0xsh: `useStore` do not auto-map state to component local bindings.
As said on a previous comment, only allow to enable reactivity.

assets/react/controllers/Incident/List.jsx
r0xsh: For now this screen fit in one component, so there is no need to use a store. (but yeah maybe later (: )

assets/react/controllers/Incident/TaskContext.jsx
r0xsh: I hesitated to extract it into a component. But I don't think we'll need it elsewhere, and I wanted to avoid cluttering the codebase with too many small JS files.

assets/react/controllers/Incident/TaskTimeline.jsx
r0xsh: The initial props binding is unusable without parsing it soo idk :shrug: no strong opinion on this :)

src/Entity/Incident/IncidentEvent.php
r0xsh: It's possible I've spent too much time immersed in the CQRS pattern, but in my perspective, an 'Action' serves as a Command that ultimately triggers an Event. Here's how I see it:
| Command               | Event                   |
|-----------------------|-------------------------|
| Reschedule            | Rescheduled             |
| Cancel                | Cancelled               |
| Report to transporter | Reported to transporter |

Also, it's worth noting that a single command can sometimes trigger multiple events.

Rather than storing the individual actions themselves in the database, I focus on capturing the consequential events they generate.

r0xsh:
> One more thing: did you test it with the released mobile app version? What will happen if somebody try to mark task as failed?

The old "mark as failed" behavior hasn't changed. I'm waiting for user feedback before making any changes to this part. Since incidents add a new step in the validation lifecycle that the rider doesn't have access to, I expect feedback from cooperatives doing "on the fly" dispatches asking me not to deactivate the ability for a rider to mark a task as failed. In the long term, I believe there will be an option to enable or disable incidents on the instance. Incidents are primarily there to manage billing and administrative issues, not logistical business problems. Cooperatives using CoopCycle solely for its logistical aspect will find incidents getting in their way.

assets/react/controllers/Incident/OrderDetails.jsx
r0xsh: Haha, yeah, there's this internal private function that gets converted on the fly into a component, and nobody bothers renaming it 🫠
Nice catch +1

assets/react/controllers/Incident/CommentBox.jsx
r0xsh: same as above; this component do not need reactivity.

src/Entity/Incident/Incident.php
```diff
@@ -0,0 +1,237 @@
...
+ *     }
+ *   }
+ * )
+ */
+class Incident implements TaggableInterface {
+    use TaggableTrait;
+
+    /**
+    * @Groups({"incident"})
+    */
```
alexsegura, line 59: Wrong PHP docblock indentation. It can prevent PHP reflection from working as expected.

templates/admin/incidents.html.twig
```diff
@@ -0,0 +1,23 @@
+{% extends "deliveries.html.twig" %}
+
+{% block breadcrumb %}
+{{ parent() }}
+<li>{% trans %}adminDashboard.incidents.title{% endtrans %}</li>
+{% endblock %}
+
+{% block content %}
+  <div class="text-right mb-4">
+    <a href="" class="btn btn-success" onClick="alert('Only available for QACoops for now')">
```
alexsegura, line 10: This `alert()` couples the app with our own internal release system. Also, this doesn't seem to do anything. 
Should be removed IMHO.

assets/react/controllers/Incident/ActionBox.jsx
vladimir-8: why not to add reactivity? So that it behaves like a standard react component

assets/react/controllers/Incident/TaskContext.jsx
vladimir-8: It does not need to be in a separate file 😉 If you rename this function to start with a capital letter it will be already a component

assets/react/controllers/Incident/TaskTimeline.jsx
vladimir-8: what I meant is just creating a separate variable instead of modifying the props, something like:

```
export default function ({ _events }) {
const events = Object.values(JSON.parse(_events));
``

src/Entity/Incident/IncidentEvent.php
vladimir-8: That makes sense, I got confused by the `TYPE..` names, maybe they should be
`const TYPE_CANCEL_TASK = "task_cancelled"`
instead of
`const TYPE_CANCEL_TASK = "cancel_task"`
(and the same for other types)

vladimir-8: Also, not sure what is the better approach. For the order events, for example, we have them implemented as an inheritance from the base event, instead of the `type` property

vladimir-8 (approved):

src/Entity/Incident/IncidentEvent.php
r0xsh: The incident event doesn't need to be as complex as the order event. They're designed this way to be sent through the message bus, enabling real-time communication. However, for incidents, such complexity isn't necessary. Simple text type works fine here.

PR #4044, opened 2024-03-07
Enable "basic unassign" with drag'n'drop

Atala:
Enable the following moves:
- from unassigned tour to unassigned
- from assigned tour to unassigned
- from assigned task to unassigned


Atala:
I did a new pull request with more changes

PR #4051, opened 2024-03-12
Unassign with dnd bis

Atala:
Add the following features for drag'n'drop :
- unassign task from tasklit by dnd
- unassign tour to unassigned tours by dnd
- move individual task outside of tour to assign it or remove from tour

at the dev level : 
- remove the need for `task.tour_id` field (will be removed in DB)
- thus remove the need to chain `modifyTaskList` and `modifyTour` calls


vladimir-8:
When I assign a linked task other linked tasked get assigned automatically. For example, if I make an order and assign Pickup, Dropoff is also assigned to the same person. But if unassign one of these tasks the other is not unassigned. Not sure if it's a correct behaviour.

It's not caused by drag and drop. It behaved the same way with the buttons. Just noticed it

js/app/dashboard/redux/selectors.js
```diff
@@ -50,13 +51,16 @@ export const selectTaskLists = taskListSelectors.selectAll
 
 export const selectBookedUsernames = taskListSelectors.selectIds
 
+
 export const belongsToGroup = task => Object.prototype.hasOwnProperty.call(task, 'group') && task.group
-export const belongsToTour = task => Object.prototype.hasOwnProperty.call(task, 'tour') && task.tour
+export const belongsToTour = task => state => selectTaskIdToTourIdMap(state).has(task['@id'])
```
vladimir-8, line 56: I think it makes code a bit easier to understand if we keep using `belongsToTour` as a utility function, instead of `taskIdToTourIdMap.has(task['@id']` And call a selector something like `selectTourWithTask`

js/app/dashboard/redux/utils.js
```diff
@@ -207,7 +207,8 @@ export const isValidTasksMultiSelect = (selectedTasks) => {
 
   return (
     selectedTasks.every(t => t.assignedTo === selectedTasks[0].assignedTo) &&
-    selectedTasks.every(t => t.tour === null) ||
-      (selectedTasks.every(t => t.tour !== null) && selectedTasks.every(t => t.tour["@id"] === selectedTasks[0].tour["@id"]))
+    selectedTasks.every(t => !taskIdToTourIdMap.has(t['@id'])) || // no tour
+      (selectedTasks.every(t => taskIdToTourIdMap.has(t['@id'])) &&
+        selectedTasks.every(t => taskIdToTourIdMap.get(t['@id']) === taskIdToTourIdMap.get(t[0]['@id'])))
```
vladimir-8, line 212: I think this is a bit easier to read:

```
  const isAllAssignedToSameUserOrNotAssigned = selectedTasks.every(t => t.assignedTo === selectedTasks[0].assignedTo)
  const isNoneInTour = selectedTasks.every(t => !taskIdToTourIdMap.has(t['@id']))
  const isAllInTour = selectedTasks.every(t => taskIdToTourIdMap.has(t['@id']))
  const isAllInSameTour = isAllInTour && selectedTasks.every(t => taskIdToTourIdMap.get(t['@id']) === taskIdToTourIdMap.get(t[0]['@id']))
  
  return (isAllAssignedToSameUserOrNotAssigned && isNoneInTour) || isAllInSameTour
```

vladimir-8 (approved):

Atala:
> When I assign a linked task other linked tasked get assigned automatically. For example, if I make an order and assign Pickup, Dropoff is also assigned to the same person. But if unassign one of these tasks the other is not unassigned. Not sure if it's a correct behaviour.
> 
> It's not caused by drag and drop. It behaved the same way with the buttons. Just noticed it

-> yes it is the correct behavior! most of the time we want them to go together same rider, but sometimes you want more fine-grained control ( https://github.com/orgs/coopcycle/discussions/12#discussioncomment-8244940 )

js/app/dashboard/redux/handleDrag.js
```diff
@@ -85,43 +94,18 @@ export function handleDragEnd(result, modifyTaskList=modifyTaskListAction, modif
...
-    }
-
-    if (source.droppableId.startsWith('tour:') && destination.droppableId.startsWith('tour:') && source.droppableId !== destination.droppableId) {
-      toast.warn("Can not move directly tasks between tours at the moment")
-      return
-    }
-
     if (source.droppableId.startsWith('group:') && destination.droppableId.startsWith('group:') && source.droppableId !== destination.droppableId) {
-      toast.warn("Can not move directly tasks between groups at the moment")
+      toast.warn("Can not move directly tasks between groups")
```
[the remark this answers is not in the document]
Atala: yes indeed i will change this one
(actually what i did is create a lot of temporary toasts to forbid some drag n drop moves, and i deleted them just aftwards - not the case here though)

js/app/dashboard/redux/selectors.js
Atala: yes it is not very nice indeed :/

you mean `selectTaskWithTour` no? because it is the problem at hand here

js/app/dashboard/redux/utils.js
Atala: yes thx better than adding lots of comments

js/app/dashboard/redux/selectors.js
vladimir-8: we have a task and want to get("select") a tour which contains this task, correct? Hence, `selectTourWithTask` or maybe `selectTourByTask' or `selectTourContainingTask` something like this..

js/app/dashboard/components/context-menus/TasksContextMenu.js
```diff
@@ -26,6 +26,7 @@ import {belongsToTour, selectLinkedTasksIds, selectNextWorkingDay, selectSelecte
 import {selectUnassignedTasks} from '../../../coopcycle-frontend-js/logistics/redux'
```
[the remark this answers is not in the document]
Atala: sorry the last commit was a quickfix and I didn't notice

js/app/dashboard/redux/selectors.js
Atala: should be `selectTourByTask` then in my understanding (?)

but in this case `belongsToTour` should be deleted no? because it has no use

vladimir-8: could be `selectTourByTask`

vladimir-8: `belongsToTour` could still be used as a util function, for example: `belongsToTour(taskIdToTourIdMap, task)` instead of `taskIdToTourIdMap.has(task['@id'])`

Atala: OK I will unify this, it not very clean at the moment in the code (kind of always different each time we check)

https://github.com/coopcycle/coopcycle-web/issues/4061

PR #4054, opened 2024-03-12
Dispatcher users should be able to perform recurring tasks operations

### LINKED ISSUE #3900: Users with dispatcher role cannot create recurring tasks
**Describe the bug**

When creating recurring tasks, the modal works fine until the user clicks on "Save" at which point the very helpful message "An error occurred" appears. The tasks are not and cannot be saved, the only option is to leave the modal.

We believe this is a problem specific to the dispatcher role, as admin it works fine.

**To Reproduce**

- Log in as a user with dispatcher rights
- Create a recurring task or set of tasks from the admin dashboard
- Click save
- Observe error message

**Expected behavior**

A user with dispatch rights should (IMO) be able to create recurring tasks. The recurring task(s) should be saved as they currently are with admin rights.

**Desktop (please complete the following information):**

 - OS: Ubuntu
 - Browser: Firefox

lucasferraro:
Fixes #3900

### DIFF
--- a/src/Entity/Task/RecurrenceRule.php
+++ b/src/Entity/Task/RecurrenceRule.php
@@ -21,31 +21,31 @@ use Symfony\Component\Validator\Constraints as Assert;
  *   collectionOperations={
  *     "get"={
  *       "method"="GET",
- *       "security"="is_granted('ROLE_ADMIN')"
+ *       "security"="is_granted('ROLE_ADMIN') or is_granted('ROLE_DISPATCHER')"
  *     },
  *     "post"={
  *       "method"="POST",
- *       "security"="is_granted('ROLE_ADMIN')"
+ *       "security"="is_granted('ROLE_ADMIN') or is_granted('ROLE_DISPATCHER')"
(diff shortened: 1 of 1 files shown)


src/Entity/Task/RecurrenceRule.php
```diff
@@ -21,31 +21,31 @@
  *   collectionOperations={
  *     "get"={
  *       "method"="GET",
- *       "security"="is_granted('ROLE_ADMIN')"
+ *       "security"="is_granted('ROLE_ADMIN') or is_granted('ROLE_DISPATCHER')"
```
vladimir-8, line 24: Should we maybe specify only `is_granted('ROLE_DISPATCHER')` here? And rely on role hierarchy to grant admins the same permissions: https://github.com/coopcycle/coopcycle-web/blob/master/app/config/security.yml#L19

vladimir-8 (approved):

Atala (approved):
Agree with the changes, I think implement vladimir's comment

Issue #4066, opened 2024-03-16
Tasks have strange order when put into a route


Paul-Eraman-CoopCycle:
When a series of tasks is put into a route they end up with first the dropoffs and later the pickups in a strange and unhelpful order.

To replicate: 
1. select a number of tasks
2. while they are selected, click the plus button to create a route. 
3. name the route
4. see the strange order

What is expected to happen:
They are ordered in the same order that they were above in the unassigned list, especially with the dropoffs first and pickups second

Video:
![routeproblemas](https://github.com/coopcycle/coopcycle-web/assets/77277854/01c60363-6f64-4067-b49b-8964e987ec96)

Atala:
actually the default order for unassignedtasks is sorted by date of completion. i will put this as the default order as well for newly created tours

Atala:
we need to dig with real world use cases because with my test data it is difficult for me to see what is useful

for example if we dragNdrop one pickup it picks the dropoff with it and I think it will place the dropoff just after the pickup **even if** in the task selection we have stuff to be done before

I am not exaclty sure on what happenned on your gif

we will check together

Paul-Eraman-CoopCycle:
what was expressed to me by mensakas was that if it could be in the order that it is in unassigned, that would make sense 

but i dont know honestly

we should do a real workshop about routes with multiple people soon

Atala:
OK seems good for now, I can do that

so when you create tours + when you DND to a tour from unassigned ?

Paul-Eraman-CoopCycle:
yep, thats what has been communicated to me

PR #4097, opened 2024-03-27
Add tasks reorder + move between riders

Atala:
This PR adds the following :
- move directly tasks between two riders
- add the capability to reorder unassigned tasks + reorder unassigned tours
-


js/app/dashboard/redux/uiReducers.js
```diff
@@ -46,6 +59,45 @@ export default (state = initialState, action) => {
...
+      return {
+        ...state,
+        unassignedTasksIdsOrder: unassignedTasksIdsOrder,
+      }
+
+    case INSERT_IN_UNASSIGNED_TASKS:
+      const tasksToInsertIds = action.tasksToInsert.map(t => t['@id'])
+      unassignedTasksIdsOrder = [...state.unassignedTasksIdsOrder]
+      _.remove(unassignedTasksIdsOrder, t => tasksToInsertIds.includes(t))
+      Array.prototype.splice.apply(unassignedTasksIdsOrder, Array.prototype.concat([ action.index, 0 ], tasksToInsertIds))
```
vladimir-8, line 75: Is not it more readable?

```
unassignedTasksIdsOrder.splice(action.index, 0, ...tasksToInsertIds)
```

js/app/dashboard/index.js
vladimir-8, line 1: when unassign linked tasks it does not respect the order in which I drop an item. No sure if it was in the scope of this issue

https://github.com/coopcycle/coopcycle-web/assets/70273239/b69be9ae-7c63-4348-8ec9-36e913e9f3ca

vladimir-8, line 1: Minor 'glitch', when I move a task between riders for a second it appears in unassigned

https://github.com/coopcycle/coopcycle-web/assets/70273239/9966e10b-e738-4080-a553-a2a5373ade45

vladimir-8 (approved):

js/app/dashboard/components/UnassignedTasks.js
```diff
@@ -10,12 +10,17 @@ import TaskGroup from './TaskGroup'
...
 import classNames from 'classnames'
 
-const StandaloneTasks =  ({tasks, offset}) => {
-  return _.map(tasks, (task, index) => <Task task={ task } draggableIndex={ (offset + index) } key={ task['@id'] } />)
+const StandaloneTasks =  ({tasks, unassignedTasksIdsOrder, offset}) => {
+  return _.map(unassignedTasksIdsOrder, (taskId, index) => {
+    const task = tasks.find(t => t['@id'] === taskId)
+    if (task) {
+      return <Task task={ task } draggableIndex={ (offset + index) } key={ task['@id'] } />
+    }
```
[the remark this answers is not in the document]
Atala: I think it does already if `task` is `undefined` then it evaluates to false ?

js/app/dashboard/redux/uiReducers.js
Atala: yes indeed, I was following the way it was written before (it was even more convoluted before, I couldnt understand the code ^^)

js/app/dashboard/redux/actions.js
```diff
@@ -286,26 +287,17 @@ export function importError(token, message) {
   return { type: IMPORT_ERROR, token, message }
 }
 
-export function modifyTaskList(username, tasks) {
+export const setUnassignedTasksLoading = createAction(SET_UNASSIGNEDTASKS_LOADING)
```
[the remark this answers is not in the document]
Atala: OK

js/app/dashboard/redux/actions.js
```diff
@@ -12,6 +12,7 @@ import {
   createTaskListFailure
 } from '../../coopcycle-frontend-js/logistics/redux'
 import { selectNextWorkingDay, selectSelectedTasks } from './selectors'
+import { createAction } from 'redux-actions'
```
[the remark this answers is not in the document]
Atala: OK

js/app/dashboard/redux/handleDrag.js
```diff
@@ -149,7 +150,19 @@ export function handleDragEnd(
       selectedTasks =  withOrderTasksForDragNDrop(selectedTasks, allTasks, taskIdToTourIdMap)
     }
 
-    if (destination.droppableId === 'unassigned') {
+    // reordered inside the unassigned tours list
+    if (
+      source.droppableId === destination.droppableId && source.droppableId === 'unassigned_tours'
+    ) {
+      const itemId = result.draggableId.startsWith('tour:') ? result.draggableId.replace('tour:', '') : result.draggableId.replace('group:', '')
+      dispatch({type: "INSERT_IN_UNASSIGNED_TOURS", itemId: itemId, index: result.destination.index})
```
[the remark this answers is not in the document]
Atala: yes sorry it was some old code when i was drafting the solution

js/app/dashboard/index.js
Atala: I will build this on top of the PR; issue #4066

Atala: yes, I could not solve this right now...

if you have any idea (and time !!!), here is the problem :
- we unassign from the original tasklist
- send AJAX to server
- **task:unassigned** arrives from the server (websocket to syncjronize web and mobile dashboard)
- it triggers a UPDATE_TASK event that removes it from any tasklists (see here https://github.com/coopcycle/coopcycle-web/blob/master/js/app/dashboard/redux/taskListEntityReducers.js#L77 )

so even if i put it ahead of time (optimistic UI update) the task in the destination task list, it will be removed from it by the websocket event.

I have no way for now to know if I triggered the websocket event (in this case i could ignore it) or somebody else did (in this case i need to apply it, or i have a conflict).

I did not find a solution that was not an awful hack or might risk to break subtle thing, so i kept the glitch

js/app/dashboard/redux/uiReducers.js
Atala: i have simplified

js/app/dashboard/components/UnassignedTasks.js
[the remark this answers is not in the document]
vladimir-8: Not sure what I meant here 😄

js/app/dashboard/index.js
vladimir-8: Yeah, I see.

Let's keep it for now.

But if we are going to work on it, we can try to do on the backend. If we know that the task is going to be assigned immediately maybe we don't need to send task:unassigned event to the websocket

assets/css/utils.scss
```diff
@@ -60,6 +60,6 @@
   font-weight: 600 !important;
 }
 
-.fa.fa-arrows {
+.cursor--grabbing {
```
vladimir-8, line 63: maybe `cursor-grabbing`? similarly to other utility classes

js/app/dashboard/components/context-menus/TasksContextMenu.js
```diff
@@ -53,7 +53,12 @@ const { hideAll } = useContextMenu({
   id: 'task-contextmenu',
 })
 
-async function _assign(tasksToAssign, username, tasksLists, date, dispatch) {
+const useAssignAction = async function(username, tasksToAssign) {
```
vladimir-8, line 56: is this a hook? then it should be a regular function itself and return an async function that will be called from a react component

js/app/dashboard/components/context-menus/TasksContextMenu.js
```diff
@@ -184,9 +188,13 @@ const DynamicMenu = () => {
   const couriers = useSelector(selectCouriersWithExclude)
   const tasksListsLoading = useSelector(selectTaskListsLoading)
```
vladimir-8, line 190: and then used as 

`const assign = useAssignAction()`

...

`  const assignSelectedOrders = (username) => assign(username, selectedOrders)`

js/app/dashboard/index.js
Atala: yes maybe

but in this case you need to pass some flag to the backend; which doesnt look that nice

I think the core problem is that we have assignment saved both on `task.assignedTo` and `Tasklist` which is nice but as to be synchronized, and sometimes one is the reference, other times reference is different.

Issue #4086, opened 2024-03-21
In order to speed up the dispatch web page As a dispatcher I want to fetch packages data asynchronously


Atala:
If my understanding is good [these (https://github.com/coopcycle/coopcycle-web/blob/master/src/Serializer/TaskNormalizer.php#L97-L105) creates 1 (dropoff) or 2 (pickup) queries per task.

<s>So on Mensakas instance it is like 600 queries instead of 300 queries just for tasks.</s> -> look below!

It is hard to predict but it could gives us a speed boost when loading tasks from the dashboard.

As these data are only displayed in the task modal for now, maybe we could fetch packages data at modal opens? thus we could create an 2nd endpoint to get details on tasks?
 
Or maybe we could improve the PHP, so the query gets packages data at once?

Atala:
with packages VS without packages

![image](https://github.com/coopcycle/coopcycle-web/assets/4426585/35997081-15b7-4bc5-a6a1-1740788fd7ce)

![image](https://github.com/coopcycle/coopcycle-web/assets/4426585/f282d207-6828-4799-ade2-c954517e0e22)

alexsegura:
The problem is that we are retrieving the `packages` in API responses since the beginning, and I can't say what is consuming this.

Atala:
Done differently with a big SQL query

PR #4104, opened 2024-03-28
clean: use custom query to get package on "Tasks" endpoint

Atala:
issue #4086

improve query time and reduce doctrine SQLs

before

image

after

image

### DIFF
--- a/src/Serializer/TaskNormalizer.php
+++ b/src/Serializer/TaskNormalizer.php
@@ -36,18 +36,6 @@ class TaskNormalizer implements NormalizerInterface, DenormalizerInterface
         $this->entityManager = $entityManager;
     }
 
-    private function _normalizeTaskPackages($packages) {
-        $packagesNormalized = [];
-        foreach ($packages as $package) {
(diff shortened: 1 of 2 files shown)


src/Serializer/TaskNormalizer.php
```diff
@@ -98,11 +86,48 @@ public function normalize($object, $format = null, array $context = array())
...
+                $qb =  $this->entityManager
+                ->getRepository(Task::class)
+                ->createQueryBuilder('t');
+
+                $query = $qb
+                    ->select('p.name AS name', 'p.name AS type', 'sum(tp.quantity) AS quantity')
+                    ->join('t.packages', 'tp', 'WITH', 'tp.task = t.id')
+                    ->join('tp.package', 'p', 'WITH', 'tp.package = p.id')
+                    ->join('t.delivery', 'd', 'WITH', 'd.id = :deliveryId')
+                    ->groupBy('p.name')
```
Atala, line 100: should be grouped by `package.id` maybe ?

because we can have different packages with the same name ?

PR #4108, opened 2024-03-29
Edenred restaurants sync

### LINKED ISSUE #3153: Implement FTP synchronization based on SIRET for Edenred
To get new restaurants, work on the Edenred solution to have them on the platform

lucasferraro:
Fixes #3153


src/Form/RestaurantType.php
```diff
@@ -189,9 +189,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
                     if ($options['edenred_enabled']) {
                         $form
-                            ->add('edenredMerchantId', TextType::class, [
```
alexsegura, line 192: Maybe we should still allow to set merchant ID manually, but it's fine. At least, we should show it (disabled).

PR #4150, opened 2024-04-16
Fix cypress tests

Atala:


Atala:
will do a new and clean PR

PR #4172, opened 2024-04-19
Allow search saved addresses in delivery form

### LINKED ISSUE #3825: In order to improve usability of the task creation form, as a dispatcher or shop owner I want to be able to search names of clients that allows their corresponding information to appear from "saved addresses"
Right now you can search your "saved addresses" by searching in the address field with the street of the client, but this is the only way to search. 

When saving an address you are saving a client name, address, and other characteristics. 

We would like to be able to search in the name field for the names of clients from those saved addresses, click on which would then automatically populate the rest of the fields in the task.

lucasferraro:
Closes #3825

This change allows to dispatchers and shop owners searching saved addresses by its name in the delivery form.
Note that Autosuggest component already show saved addresses but is limited only to 5 addresses. This new search component will search for all the saved address.

https://github.com/coopcycle/coopcycle-web/assets/4819244/9d70aa87-36d1-4ea1-bf9f-bf10ae834f4f

### DIFF
--- a/js/app/delivery/AddressBook.js
+++ b/js/app/delivery/AddressBook.js
@@ -413,7 +413,7 @@ export default function(el, options) {
 
   render(
     <AddressBook
-      allowSearchSavedAddresses={ allowSearchSavedAddresses }
+      allowSearchSavedAddresses={ Boolean(allowSearchSavedAddresses) }
       addresses={ addresses }
       initialAddress={ address }
       details={ details }
--- a/js/app/delivery/AddressBook.js
+++ b/js/app/delivery/AddressBook.js
@@ -12,6 +12,7 @@ import AddressAutosuggest from '../components/AddressAutosuggest'
 import { getCountry } from '../i18n'
 
 import './AddressBook.scss'
+import { SavedAddressesBox } from './SavedAddressBox'
 
 const AddressPopoverIcon = ({ prop }) => {
   switch (prop) {
@@ -150,12 +151,14 @@ const AddressBook = ({
   onDuplicateAddress,
   onClear,
   details,
+  allowSearchSavedAddresses,
   ...otherProps
 }) => {
 
   const { t } = useTranslation()
   const [ address, setAddress ] = useState(initialAddress)
   const [ isModalOpen, setModalOpen ] = useState(false)
+  const [ searchSavedAddresses, setSearchSavedAddresses ] = useState(false)
 
   const onAddressPropChange = (address, prop, value) => {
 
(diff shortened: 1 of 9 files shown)


js/app/delivery/AddressBook.js
```diff
@@ -393,6 +413,7 @@ export default function(el, options) {
 
   render(
     <AddressBook
+      allowSearchSavedAddresses={ allowSearchSavedAddresses }
```
vladimir-8, line 416: maybe better to explicitly convert it to boolean like 

`allowSearchSavedAddresses={ Boolean(allowSearchSavedAddresses) }` ?

I suppose it will be `undefined` in all places where it is not set

js/app/delivery/SavedAddressBox.js
vladimir-8, line 1: Maybe it's because I don't have any address, it crashes when I click on `Search a saved address` and then enter any symbol into the `Address name...` text field




Screenshot 2024-04-19 at 10 45 14

vladimir-8: Actually, I have some addresses:

Screenshot 2024-04-19 at 10 49 38

vladimir-8 (commented):
The code looks good me, but I can not test it at the moment. Please see the screenshot with the crash and let me know if you need more info

js/app/delivery/SavedAddressBox.js
lucasferraro: Hmm nice catch, I'll try to reproduce it

lucasferraro: Seems like your default address (second line) does not have a name. I don't know in which cases that can happen but I will add a check for handling these things and avoid crushing.

Paul-Eraman-CoopCycle: when saving an address in the address book you can save without putting in a name. i would make it obligatory there, since it is already obligatory in the order creation page

Atala:
I would add a question/remarks : on your example there are quite a few saved addresses, and it seems that the "non saved addresses" (gocode search) section is far from reachable. I think maybe you should have a `max- height` attribute for the saved section and make it scrollable, so user see can either scroll first section for saved addresses or reach directly for the 2nd section/search section

Paul-Eraman-CoopCycle:
Can you explain that in less technical terms alois?

Atala:
if you have too many saved addresses you will only see these in the dropdown, not the second section which is for searching new addresses (you will need to scroll to see them)

(if I am not mistaken)

Paul-Eraman-CoopCycle:
I think that's fine...you just click out of the box to make that disappear right?

lucasferraro:
> if you have too many saved addresses you will only see these in the dropdown, not the second section which is for searching new addresses (you will need to scroll to see them)
> 
> (if I am not mistaken)

I don't know which saved addresses section you are talking about, because now there are two places where users can select them.

1) New search save addresses component (it has a fixed `height` with an auto scroll to avoid cases where the list of results may take a large part of the screen)
2) Existing autosuggest address input with a "Saved addresses" section at the top. This section doesn't change with this PR and in that section we are only displaying 5 addresses, so there's no such case where the list can take a lot of the screen) 

Maybe I'm missing something but I don't know where you see that the "non saved addresses" section is far from reachable.

js/app/delivery/SavedAddressBox.js
lucasferraro: Ok I'll add that mandatory rule there so we avoid this kind of errors

Atala:
sorry i was mistaken (only top five saved addresses are shown)

vladimir-8:
Tried again, it works now.

Maybe it's too late for this question, but why are we doing it in a separate field instead of extending/replacing autosuggest component? It look a bit confusing for me to have two separate fields for a similar functionality.

vladimir-8 (approved):

Paul-Eraman-CoopCycle:
@vladimir-8 we tried that but it got complicated
since we will be redoing the entire page soon, we went with this "dirty fix" 

id love to see a gif, but if it works as described, all good for me

PR #4197, opened 2024-04-23
Optimize variant resolver

### LINKED ISSUE #4090: When I try to add a product to the cart, I see a progress indicator for a few seconds and then the cart appear empty as if the product was not added (in reality it's added, but not displayed) It happens only for one specific product
**Describe the bug**

When I try to add a product to the cart, I see a progress indicator for a few seconds and then the cart appear empty as if the product was not added (in reality it's added, but not displayed) It happens only for one specific product

**To Reproduce**
1. Go to https://sicklo.coopcycle.org/fr/restaurant/47-gustavo
2. Click on 'MENU KEBAB VEGAN'
3. Select all required option and try to add it to the cart
4. See the cart is still empty
5. (try to add another product and 'MENU KEBAB VEGAN' will appear in the cart)

**Expected behavior**

The product should be added to the cart immediately

**Additional context**

I found `504 Gateway Timeout` logs for this request, and also for some others:

https://app.datadoghq.com/logs?query=%40http.status_code%3A504%20&cols=host%2Cservice&fromUser=true&index=%2A&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1710969294876&to_ts=1711228494876&live=true

Maybe we can try some of these suggestions:

https://stackoverflow.com/questions/24453388/nginx-reverse-proxy-causing-504-gateway-timeout

alexsegura:
Fixes #4090

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,6 +68,8 @@ jobs:
         run: docker-compose logs php
       - name: Create database
         run: docker-compose exec -T php bin/console doctrine:schema:create --env=test
+      - name: Create typesense collections
+        run: docker-compose exec -T php bin/console typesense:create --env=test
       - name: Validate database schema
         run: docker-compose exec -T php bin/console doctrine:schema:validate --env=test --skip-sync
       - name: Run PHPUnit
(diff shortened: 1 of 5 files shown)


Atala:
Tested OK, I added some combinaison to the cart and then created another cart with the same option choice and it found the match

Issue #4003, opened 2024-02-13
In order to clean our coopcycle As an admin I want to be able to delete stores


lillebike:
**In order to**  `achieve something`
in order to Delete store that are not in use, I want a delete store button on the web version

as an admin

I want a button to delete stores that are out of use

**Whereas currently**  there is no way to delete a store

**Additional context**
We worked with companies that do not work with us anymore or literally closed, so in our store list we got plenty of stores that are useless, we want to delete them

Paul-Eraman-CoopCycle:
- [ ] stores
- [ ] shops
- [ ] pricing rules
- [ ] time slots
- [ ] Formulas
- [ ] package sets
- [ ] zones
- [ ] tags

Atala:
Actually it is more complex than i expected. A store is linked to all these objects, so we need to define behaviour for each of them

|object | behaviour |
| --- | ---- |
"store_address" |
"delivery" |
"store_time_slot" |
"delivery_quote" |
"api_app" |
"urbantz_hub" |
"woopit_integration" |
"delivery_import_queue" |
"api_user_store" |
"task_rrule" |

Paul-Eraman-CoopCycle:
What does "define behavior " mean? Decide what happens to them?

Paul-Eraman-CoopCycle:
🥳

PR #4198, opened 2024-04-23
feat: add the ability to soft delete stores

Atala:
- add "deletedAt" field in table store
- add a "delete" button on store detail page (with confirmation pop up)
- remove "owner" link with store on soft delete
- hides store on stores list after soft delete

#4003 

and
- add the ability to delete store from stores list
- add the ability to delete restaurant from restaurants list

### DIFF
--- a/js/app/store/list.js
+++ b/js/app/store/list.js
@@ -1,0 +1,24 @@
+import i18n from '../i18n'
+
+document.querySelectorAll('.delete-store').forEach((el) => {
+    el.addEventListener('click', (e) => {
+
+    if (!window.confirm(i18n.t('CONFIRM_DELETE_WITH_PLACEHOLDER', { object_name: e.target.dataset.storeName }))) {
+      e.preventDefault()
+      return
+    }
+
+    const jwtToken = document.querySelector("#stores-list").dataset.jwt
+    const headers = {
+      'Authorization': `Bearer ${jwtToken}`,
+      'Accept': 'application/ld+json',
+      'Content-Type': 'application/ld+json'
+    }
+
+    const url = '/api/stores/' + e.target.dataset.storeId
+    fetch(url, {method: "DELETE", headers: headers}).then(
+      function () { location.reload(); }
+    );
+
+  });
+})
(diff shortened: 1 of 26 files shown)


src/Controller/Utils/StoreTrait.php
```diff
@@ -49,6 +50,9 @@
 
 trait StoreTrait
 {
+    /**
+    * @HideSoftDeleted
+    */
```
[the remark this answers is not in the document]
Atala: Ok will do I didnt know

src/Doctrine/EventSubscriber/PostSoftDeleteSubscriber.php
```diff
@@ -89,5 +91,15 @@ public function postSoftDelete(LifecycleEventArgs $args)
 
             $unitOfWork->computeChangeSets();
         }
+
+        if ($entity instanceof Store) {
```
alexsegura, line 95: Maybe the store should be also "reset" here. I mean, we should set the pricing rule set, time slots & packages to NULL. 
I'm foreseeing the deletion of pricing rules, time slots, packages. This way, if soft deleted stores are already not using them anymore, it would simplify things.

js/app/store/list.js
```diff
@@ -0,0 +1,24 @@
+import i18n from '../i18n'
+
+document.querySelectorAll('.delete-store').forEach((el) => {
```
alexsegura, line 3: Let's see if we can generalize this when you implement deletion for other objects, to avoid re-implementing it multiple times.

alexsegura (approved):

js/app/store/list.js
Atala: could be a small util function that gets selector, object name, object id, ect

src/Doctrine/EventSubscriber/PostSoftDeleteSubscriber.php
Atala: OK will do in the next PRs

Issue #1753, opened 2020-09-25
Use Centrifugo instead of WebSocket / Socket.io


alexsegura:
[Centrifugo](https://centrifugal.github.io/centrifugo/) looks very nice to replace what we have for live update. 

Among other things, it has message recovery on reconnect, and channels.

alexsegura:
Legacy WebSockets have been removed 

96954b01a8f4d94b6154ecdea9cd5f0c2cddfe98...392fbb8590bff0029b049b39ec17f76cc79e2702

PR #4208, opened 2024-04-24
fix: set pickup from store on POST_DESERIALIZE not PRE_VALIDATE

Atala:
issue #1753

### DIFF
--- a/features/deliveries.feature
+++ b/features/deliveries.feature
@@ -1153,7 +1153,7 @@ Feature: Deliveries
       }
       """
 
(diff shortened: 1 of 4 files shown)


alexsegura:
@Atala did you reproduce the issue in Behat in the end?

Atala:
@alexsegura yes the issue was 100% reproducible on behat, i added the test

I don't understand exactly how i managed to place an order locally - maybe i didnt place an order in the end

I will resolve the php tests problem in master

alexsegura:
@Atala I fixed the tests, you can rebase on master

Issue #4214, opened 2024-04-26
Prevent clicking on the Order button multiple times on the Address page (mobile web only)


vladimir-8:
**To Reproduce**
1. Add a few items to the order
2. Go to Address page
3. Click `Order` button multiple times
4. In come cases duplicate adjustments are created in the database (https://github.com/coopcycle/coopcycle-web/issues/4128) and the payment is wrong

**Expected behavior**
The button should be disabled after the first click

**Device (please complete the following information):**
 - OS: iOS 17 (might be also on Android)
 - Browser: safari

**Additional context**

https://coopcycle.slack.com/archives/G855MRP7V/p1714157060423889?thread_ts=1713918961.968519&cid=G855MRP7V

Issue #4140, opened 2024-04-09
Prevent adding tips after click on the `Order` button on the Address page


vladimir-8:
**To Reproduce**
1. Add a few items to the order
2. Go to Address page
3. Click `Order`
4. Click Add tips, before the next page is loaded
5. In come cases duplicate adjustments are created

**Expected behavior**

There should be always a single: fee, tax and delivery adjustment per order

**Additional context**

Found the following sequence of requests in the logs:

```
click `Order` button on the restaurant page:
web02 {"message"":"Order #42955 | OrderLogStartProcessor | processing started | triggered by: POST /fr/restaurant/69-cantine-numero-10; at: AppBundle\\Sylius\\OrderProcessing\\OrderLogStartProcessor::process() called at: file: /vendor/sylius/order/Processor/CompositeOrderProcessor.php line: 41 | Sylius\\Component\\Order\\Processor\\CompositeOrderProcessor::process() called at: file: /src/Form/Order/CartType.php line: 123 | AppBundle\\Form\\Order\\CartType::AppBundle\\Form\\Order\\{closure}() called at: file: /vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php line: 230","datetime":"2024-04-04T12:14:45.547423+02:00"}


click on add tips? (#1)
web02 {"message"":"Order #42955 | OrderLogStartProcessor | processing started | triggered by: POST /order/; at: AppBundle\\Sylius\\OrderProcessing\\OrderLogStartProcessor::process() called at: file: /vendor/sylius/order/Processor/CompositeOrderProcessor.php line: 41 | Sylius\\Component\\Order\\Processor\\CompositeOrderProcessor::process() called at: file: /src/Controller/OrderController.php line: 154 | AppBundle\\Controller\\OrderController::indexAction() called at: file: /vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php line: 163","datetime":"2024-04-04T12:14:54.974252+02:00"}


click on add tips? (#2)
[...]

PR #4231, opened 2024-04-30
address page: order button

vladimir-8:
Fixes:

- https://github.com/coopcycle/coopcycle-web/issues/4214
- https://github.com/coopcycle/coopcycle-web/issues/4140

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -249,11 +249,11 @@ h1, h2, h3, h4, h5, h6 {
   border-bottom: 1px solid darken(#ededed, 10%);
 }
 
-.btn-payment > .fa-spinner {
+.btn > .fa-spinner {
   display:none;
 }
 
-.btn-payment__loading > .fa-spinner {
+.btn__loading > .fa-spinner {
   display:inline-block;
 }
 
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -249,11 +249,11 @@ h1, h2, h3, h4, h5, h6 {
   border-bottom: 1px solid darken(#ededed, 10%);
 }
 
-.btn > .fa-spinner {
+.btn > .btn__loading_indicator {
   display:none;
 }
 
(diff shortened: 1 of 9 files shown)


assets/css/styles.scss
```diff
@@ -249,11 +249,11 @@ h1, h2, h3, h4, h5, h6 {
   border-bottom: 1px solid darken(#ededed, 10%);
 }
 
-.btn-payment > .fa-spinner {
+.btn > .fa-spinner {
```
Atala, line 252: if i am not mistaken this break the spinner in the cart button when you add products on the restaurant page (the button is empty)

js/app/checkout/summary.js
```diff
@@ -23,10 +24,7 @@ var im = new Inputmask("currency", {
     suffix: currency.position === 'postfix' ? `${space}${getCurrencySymbol()}` : ''
 })
 
-const submitBtns = document.querySelectorAll(
-  'form[name="checkout_address"] input[type="submit"],button[type="submit"]'
-)
-const mainSubmitBtn = _.find(Array.from(submitBtns), btn => !btn.hasAttribute('name'))
+const submitPageBtn = document.querySelector('.btn-submit-page')
```
Atala, line 27: :+1:

Atala (changes requested):
hello

the desired behavior is well implemented but the CSS change broke the spinner in js/app/restaurant/components/Order/CartButton.js and js/app/restaurant/components/Order/OrderState.js (and maybe other places, I just noticed this one by chance)

the "loading state" on the mobile web cart button seems to work fine

please have a look and then merge!

assets/css/styles.scss
vladimir-8: wow, nice catch. I'll fix it

vladimir-8:
@Atala I fixed the issue with the loading indicator caused by my changes, and one more issue that already existed and the loading indicator had not been shown when the `Order` button was clicked on the restaurant page

Issue #4251, opened 2024-05-07
Business account employee invitation should work for existing platform accounts


lucasferraro:
**In order to**  `achieve something`
Existing users should be able to accept Business Account invitation as employee using their existing account.

**As a** `role`
User

**I want** `something to be added or changed`
When an invitation link for joining to a Business Account is shared to an existing account in the platform, the invitation link should not ask for personal account creation. Instead it should ask for login with existing account and then just associate that account to the Business Account where the invitation link has been shared from.

**Whereas currently** `this happens`
Now we always ask for a personal account creation when the invitation link is opened.

**Additional context**
Video where you can see how an invitation link shared by a business account manager to an existing platform user opens the personal account creation screen instead of just ask for login (if it is needed) and associate the existing account to the busines account.

https://github.com/coopcycle/coopcycle-web/assets/4819244/85144c94-ff83-4021-9459-4e6842ea7d65

lucasferraro:
This how it looks this feature implemented (code in [PR](https://github.com/coopcycle/coopcycle-web/pull/4253))

**An existing user opens the link of the invitation while they are not logged in to the platform**

https://github.com/coopcycle/coopcycle-web/assets/4819244/c2ab6d5b-1a25-46d8-83c7-4329a83ba635

**An existing and logged in user opens the link of the invitation**

https://github.com/coopcycle/coopcycle-web/assets/4819244/fce861b8-9da6-41d4-8635-ed3e017bec37

PR #4253, opened 2024-05-10
Associate existing users as business accounts employees

lucasferraro:
With these changes now an existing user should be able to accept the invitation to associate its account to a Business Account (https://github.com/coopcycle/coopcycle-web/issues/4251).

**An existing user opens the link of the invitation while they are not logged in to the platform**

https://github.com/coopcycle/coopcycle-web/assets/4819244/c2ab6d5b-1a25-46d8-83c7-4329a83ba635

**An existing and logged in user opens the link of the invitation**

https://github.com/coopcycle/coopcycle-web/assets/4819244/fce861b8-9da6-41d4-8635-ed3e017bec37


Atala (approved):
hello lucas

it works as expected but I think the UI could be slightly improve, because the "yellow notif" which leads you to login is not enough visible i think (i was focus on the error message)

it should be written closer to the error message, maybe in the same text string as the error message if possible

PR #4264, opened 2024-05-14
e2e tests setup; minor improvements

vladimir-8:
- updated: .env setup; https://symfony.com/doc/current/configuration.html#overriding-environment-values-via-env-local
- an attempt to fix flaky test

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,1 @@
+APP_ENV=test
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,8 @@
 .docker-sync
 
 .env
+*.env.local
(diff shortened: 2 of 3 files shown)


Atala:
@vladimir-8 
- I have cherry-picked the increased delay in master
- the .env.test.local doesnt seem so useful to me (I dont undersnatd if it is used somewhere?), so I d rather keep it for when it is in use/documented/aso. and call it a day

what do you think?

Atala:
it is just i dont get where this .env.test is loaded?
i just .env file being loaded for example in web/index.php or app/autoload.php

PR #4301, opened 2024-05-22
New method to get setup vendor information

lucasferraro:
In [this discussion](https://github.com/coopcycle/coopcycle-web/discussions/4302) you can find the idea behind these changes.

### DIFF
--- a/src/Entity/Sylius/Order.php
+++ b/src/Entity/Sylius/Order.php
@@ -1254,7 +1254,7 @@ class Order extends BaseOrder implements OrderInterface
         return $this->customer->getUser();
     }
 
-    public function getSetupVendor(): ?Vendor
+    public function getVendorConditions(): ?Vendor
     {
         if (null !== $this->getBusinessAccount()) {
             return $this->getBusinessAccount()->getBusinessRestaurantGroup();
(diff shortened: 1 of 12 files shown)


Paul-Eraman-CoopCycle:
can you retitle this PR so it is more human readable to non-technical users?

Atala:
great! I think with the comments on top of the function that will do it

I just think `getVendorSetup` is not the best name. I see we are using `settlement` for the tab in the restaurant admin interface ( https://github.com/search?q=repo%3Acoopcycle%2Fcoopcycle-web%20settlement&type=code ). I also do think this is a mistake, because `settlement` indicates there is a conflict with the restaurant which is not the case

so I think the right word is `conditions` and thus the method should be `getVendorConditions` (+ we need to change the `settlement`thing everywhere) - or maybe `deliveryconditions`. to be double checked with a native speaker (paul)

lucasferraro:
> great! I think with the comments on top of the function that will do it
> 
> I just think `getVendorSetup` is not the best name. I see we are using `settlement` for the tab in the restaurant admin interface ( https://github.com/search?q=repo%3Acoopcycle%2Fcoopcycle-web%20settlement&type=code ). I also do think this is a mistake, because `settlement` indicates there is a conflict with the restaurant which is not the case
> 
> so I think the right word is `conditions` and thus the method should be `getVendorConditions` (+ we need to change the `settlement`thing everywhere) - or maybe `deliveryconditions`. to be double checked with a native speaker (paul)

@Atala yes, naming is a nightmare and I couldn't decide for an option. But I like your suggestion, I think `getVendorConditions` describes better the method. thanks

PR #4319, opened 2024-05-23
generate test orders and tasks

vladimir-8:


Atala:
hello,

If i understand well we can run this command any day and it will
 - create orders for today
 - not recreate/delete any previous existing restaurants (or data in general)
 ?
 
if so it is ok for me to have this in the `setup` command, otherwise maybe a dedicated command is better

vladimir-8:
It will delete any existing restaurants, which is the current behaviour of `php bin/console coopcycle:demo:init "$@"` command. So this PR only adds more data to this command. I'd like to be able to only generate orders for a day without touching existing data, but that's for another PR 😉

Issue #4310, opened 2024-05-23
When connected as business account/in business mode i see ALL restaurants in the frontpage not just the ones of the Business Restaurant Group


Atala:
Here only one restaurant should show (name  La Cuisine d'Asmar )

![image](https://github.com/coopcycle/coopcycle-web/assets/4426585/825c133c-ec22-4957-ad6c-ea21f69df312)

PR #4355, opened 2024-05-30
When connected in business mode user should see only available restaurants from the Business Account configuration

lucasferraro:
Fix for #4310 

Also, a warning message has been added in restaurant screen so if a user changes the mode to 'Business' while is in a restaurant screen we display a warning message if that current restaurant is not available in the Business Account config. (look at it in the video)

https://github.com/coopcycle/coopcycle-web/assets/4819244/590c43a6-0ee9-45fd-9d15-5d141eda5374

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -486,6 +486,8 @@ services:
     tags:
       - { name: doctrine.event_subscriber, connection: default }
 
+  AppBundle\EventListener\LogoutSubscriber: ~
+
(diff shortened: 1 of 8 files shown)


Atala (commented):
image

when you go click on "coopcycle" in the topbar it doesn't set  you in business mode so you still see all restaurants (business mode is correctly set when you click "order now")

you should force the business mode when the user clicks on "finish my inscription" I think

Issue #4296, opened 2024-05-22
In business restaurant group invitation process; check if the user exists before the user enters all of his detail 


Atala:
Now : the invited user enters his email, then all his details THEN we check the email

Better : the invited user enters his email and we check right away the email, so he doesn't waste time

lucasferraro:
@Atala 
We are validating the email in the same way for all kind of new users registration. 
Do we want to change it to validate at the beginning for one of them, for some or for all?
- Business Account managers invitation
- Business Account employee invitation 
- Regular users with invitation link
- Public registration screen

Paul-Eraman-CoopCycle:
THis would be best for all users in all situations. The same way we check if the username is taken once they have finished typing it, when they finish writing their email and click out of the box (not while they are writing) it should check if we already have an account made with that email

Atala:
yes i think same as username is nice, maybe with a longer delay before checking i had the impression it was a little bit too short?

i dont know how it is done technically this check, i guess with an API endpoint ?

lucasferraro:
Yes, for username we are using an API endpoint  `/register/suggest`. So I think we could have a new one like `/register/check-email` which can be called after a user type a valid email (I think does not make any sense check for the email while is not complete) in the input.

PR #4391, opened 2024-06-04
Check if email exists before submit registration form

lucasferraro:
#4296 

https://github.com/coopcycle/coopcycle-web/assets/4819244/c5cbacb6-6253-49f0-b144-5dabb16f892a

https://github.com/coopcycle/coopcycle-web/assets/4819244/d12635af-aba8-434c-953c-a128710d97ae

### DIFF
--- a/js/app/register/index.js
+++ b/js/app/register/index.js
@@ -7,8 +7,6 @@ const emailInput =
 const usernameInput =
   document.querySelector('[id$=_username]')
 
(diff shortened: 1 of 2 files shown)


Atala (changes requested):
no-email-check

Atala:
The error message on existing email is not the good one, it should be the one that says "already have an account?" please login (the one we added previously)

image

PR #4422, opened 2024-06-17
Support multi points in time based pricing functions 

### LINKED ISSUE #3977: Make multi-dropoff orders work with time-based and km based pricing rules
Currently multi-point pricing works UNLESS you use time-baed rules or km-based rules

we need to implement multi-dropoff pricing for:

- [ ] time-based rules
- [ ] km-based rules

Time based is a technical solution that is already defined by @alexsegura , please explain in comments.

Km based is harder, and the current suggestion is to have the platform offer to optimize the route for the client. This would look like:

**Case 1:**  I have one pickup and three dropoffs with the same time characteristics, but I have put them in an order that is geographically inefficient. If I decline the optimization, this will charge me XXX€ for the kms. 

Screenshot 2024-01-21 at 18 12 09

However, if I push the optimization button, it will reorder the dropoffs into a more efficent route and only charge me X€
Screenshot 2024-01-21 at 18 13 42

**Case 2:** I have one pickup and 3 dropoffs. Dropoffs 1 and three can be delivered throughout the entire day, but dropoff 2 has a precise time range of 15:30 to 16:00 and the pickup is after 9:00. In this case, the VROOM optimizer we have currently would put that dropoff last no matter the geographic location, given that it must be delivered after the others. Basically, we use the existing VROOM engine (it works with time ranges), and the km of this route creates the price. 

Screenshot 2024-01-21 at 18 16 04

Having documentation for this INSIDE the platform for clients is a MUST

alexsegura:
Fixes #3977

Since 0437ce115b9b970306e5a1343bf3de7c697af467, it's possible to define time-based functions in pricing rules, i.e: 
- `diff_days(pickup)`
- `diff_hours(pickup)`
- `time_range_length(pickup, 'hours')`

As those functions were first designed to work at **delivery** level (i.e 2 tasks, pickup & dropoff), to make them work at **task** level while not breaking everything, [we inject a fake/empty task in the expression language context](https://github.com/coopcycle/coopcycle-web/blob/b493c7a30ee43439e0e7231d6f6f9c641ac7ea96/src/Entity/Task.php#L1106-L1127), and we return `-1` when we encounter this fake/empty task, as a way to "skip" the rule matching. 

It [works well for distance evaluation](https://github.com/coopcycle/coopcycle-web/blob/b493c7a30ee43439e0e7231d6f6f9c641ac7ea96/src/Entity/Task.php#L1117), but not for time. 
Because a pricing rule could contains this: 

```
diff_hours(pickup) < 3
```

 If we return `-1`, it matches. 
 
 --- 
 
This pull request allows passing an additional parameter to those functions, so that they return a **boolean** instead of a numeric value. Not passing this argument is still supported, but deprecated. 
 
| BEFORE | AFTER |
| -------- | ------- |
| `diff_days(pickup) < 3` | `diff_days(pickup, '< 3')` |
| `diff_hours(pickup) < 3` | `diff_hours(pickup, '< 3')` |
| `time_range_length(pickup, 'hours') < 3` | `time_range_length(pickup, , 'hours', '< 3')` |

### DIFF
--- a/app/DoctrineMigrations/Version20240617121101.php
+++ b/app/DoctrineMigrations/Version20240617121101.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20240617121101 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $stmt = $this->connection->prepare('SELECT id, expression FROM pricing_rule');
+
+        $result = $stmt->execute();
+
+        while ($pricingRule = $result->fetchAssociative()) {
+
+            $expression = $pricingRule['expression'];
+
+            if (str_contains($expression, 'diff_days')
+            ||  str_contains($expression, 'diff_hours')
+            ||  str_contains($expression, 'time_range_length')) {
+
+                $expression = preg_replace('/(?<func>diff_(hours|days))\(pickup\) (?<expression>([<>]|==|in) [0-9]+(\.\.)*[0-9]*)/', '${1}(pickup, \'${3}\')', $expression);
+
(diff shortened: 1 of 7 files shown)


Paul-Eraman-CoopCycle:
Both for time and km??

alexsegura:
@Paul-Eraman-CoopCycle 

> Both for time and km??

No, for km it's not possible.

alexsegura:
@Paul-Eraman-CoopCycle 

Sorry I didn't read fully #3977. Can we rethink about it once we have decent route optimization?

Paul-Eraman-CoopCycle:
Phone call? jaja

vladimir-8 (approved):

PR #4447, opened 2024-06-21
Suggest route optimization on delivery form

### LINKED ISSUE #4432: Suggest users to optimize multi-point deliveries
When a user inputs a delivery with multiple points, we run an optimization in the background, we compare the distance of the optimized route to the one entered by the user, and if it's lower, we suggest the user to accept the suggested route.

alexsegura:
Fixes #4432

suggest_optim

### DIFF
--- a/js/app/delivery/form.js
+++ b/js/app/delivery/form.js
@@ -6,6 +6,9 @@ require('gasparesganga-jquery-loading-overlay')
 
 import DeliveryForm from '../forms/delivery'
 import PricePreview from './PricePreview'
+import axios from 'axios'
+
+const baseURL = location.protocol + '//' + location.host
 
 import './form.scss'
 
@@ -150,6 +153,36 @@ const updateData = (form, delivery) => {
       })
     })
 
+    const loadSuggestions = new Promise((resolve) => {
+
+      $.getJSON(window.Routing.generate('profile_jwt'))
+        .then(result => {
+
+          axios({
+            method: 'post',
+            url: `${baseURL}/api/deliveries/suggest_optimizations`,
+            data: {
+              ...delivery,
+              tasks: delivery.tasks.slice(0).map(t => ({
+                ...t,
+                address: serializeAddress(t.address)
+              }))
+            },
+            headers: {
+              Accept: 'application/ld+json',
+              'Content-Type': 'application/ld+json',
+              Authorization: `Bearer ${result.jwt}`
+            }
+          })
+          .then(response => {
+            if (response.data.suggestions.length > 0) {
+              form.showSuggestions(response.data.suggestions)
+            }
+            resolve()
+          })
+        })
+    })
+
     const updatePrice = new Promise((resolve) => {
       if (delivery.store && pricePreview) {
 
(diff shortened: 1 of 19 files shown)


src/Entity/Task.php
```diff
@@ -1192,4 +1194,13 @@ public function setPrefetchedPackagesAndWeight($prefetchedPackagesAndWeight)
 
         return $this;
     }
+
+    public static function fixTimeWindow(Task $task)
```
Atala, line 1198: can we change the name? it is hard to tell what this function is doing

`presetDoneAfter` ?

src/Action/Delivery/SuggestOptimizations.php
```diff
@@ -0,0 +1,93 @@
...
+use AppBundle\Service\RoutingInterface;
+use AppBundle\Vroom\RoutingProblem;
+use AppBundle\Vroom\Vehicle;
+
+class SuggestOptimizations
+{
+    public function __construct(private RouteOptimizer $optimizer, private RoutingInterface $routing)
+    {}
+
+    public function __invoke(Delivery $data): OptimizationSuggestions
```
Atala, line 19: why not use `RouteOptimizer.optimizePickupsAndDelivery` ? it seems it does more or less the same thing except storing the gain

js/app/delivery/form.js
```diff
@@ -132,6 +135,37 @@ form = new DeliveryForm('delivery', {
...
+              headers: {
+                Accept: 'application/ld+json',
+                'Content-Type': 'application/ld+json',
+                Authorization: `Bearer ${result.jwt}`
+              }
+            })
+            .then(response => {
+              if (response.data.suggestions.length > 0) {
+                this.showSuggestions(response.data.suggestions)
+                // this.reorder(response.data.suggestions[0].order)
```
Atala, line 162: to be removed

Atala (changes requested):
see the comments, it may be useful to use existing "optimizePickupsAndDelivery" function to do the optimization as it looks a lot the same. or maybe move at least the code in the optimizer.

I have found the following bug :
- enter a delivery with 4 points that is suboptimized
- click "yes" in optimization modal
- the points are not in the right order in the form

image

js/app/delivery/form/SuggestionModal.js
```diff
@@ -0,0 +1,78 @@
...
+            { suggestedTasks.map((task, index) =>
+              <Timeline.Item key={ `suggested-order-${index}` }>{ task.address?.streetAddress }</Timeline.Item>
+            ) }
+          </Timeline>
+        </div>
+      </div>
+      <div className="d-flex justify-content-between align-items-center">
+        <span>{ t('DELIVERY_OPTIMIZATION_SUGGESTION_CONFIRM') }</span>
+        <div>
+          <button className="btn btn-primary mr-1" type="button" onClick={ onClickYes }>
```
Atala, line 40: you should witch yex/no button (no/cancel should be on the left of the modal footer)

src/Action/Delivery/SuggestOptimizations.php
alexsegura: It doesn't do the same: it's for multiple pickups + one dropoff.

Atala: OK
maybe create a new method in the RouteOptimizer ? so we have everything there

Paul-Eraman-CoopCycle:
TO DO

- [ ] try out on LCR

Atala:
- the fix for the failing test is in master I think

- by adding/removing dropoffs I managed to trigger the following error :

alexsegura:
> by adding/removing dropoffs I managed to trigger the following error :
> Uncaught TypeError: can't access property "toLowerCase", task.type is undefined

Ok I see. It happens when doing this:
- Add a new dropoff
- Remove the first one
- Try to edit a property of the added dropoff

It's because of index changes. Looking for a fix.

PR #4452, opened 2024-06-25
In the admin, add the ability to delete packages in packages set

Atala:


alexsegura (commented):
The problem I see here is that we will lose information / history. 
Maybe this is more complicated than that, and the actual packages attached to a task would need to be a sort of "frozen clone" of the original packages. 

Wouldn't soft delete be enough? We just want to mask them in the list actually.

Atala:
My opinion on soft delete is that it is a "false friend" because when you write new code you have to remember to enable soft deletion if needed. it is easy to remember for me, because i implemented it but other contributors would struggle to think about it.

That said i agree with you that we should have a frozen state, i.e. write packages details in TaskPackage table without linking to a package with foreign key  (because for now we can already edit the package name or volume units !).

(also is the e-commerce part working like this with "frozen state" ?? (i hope so!))

what do you think will be the easiest here? i dont want to spend too much time on this issue

Atala:
@alexsegura i implemented soft deletion ;)

PR #4455, opened 2024-06-25
Show task(s) in the right hand columns when selecting a task on the map

Atala:
when selecting a bunch of tasks :
- open all related unassigned groups
- open all related tours
- open all related tasklists

scrollIntoView of the first task

### DIFF
--- a/js/app/dashboard/components/context-menus/TasksContextMenu.js
+++ b/js/app/dashboard/components/context-menus/TasksContextMenu.js
@@ -20,6 +20,7 @@ import {
   openCreateTourModal,
   openReportIncidentModal,
   openTaskRescheduleModal,
+  openTaskTaskList,
   removeTasksFromGroup,
   restoreTasks,
   setCurrentTask,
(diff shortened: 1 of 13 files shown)


js/app/dashboard/components/context-menus/TasksContextMenu.js
```diff
@@ -177,13 +180,24 @@ export function getAvailableActionsForTasks(selectedTasks, unassignedTasks, link
   return actions
 }
 
+function usePrevious(old) {
```
vladimir-8, line 183: What does it do?

Atala: i don't understand very well this code ( shame :( ) but it keeps in the ref the previous value of `selectedTasks`  so we can compare with the new value. like what we have before with `componentDidUpdate(prevProps, props)`.

vladimir-8: I see. Then maybe put it in a separate file so we can re-use it

vladimir-8:
Please check `dispatch/assign_task_with_dnd.cy.js` test

js/app/dashboard/components/context-menus/TasksContextMenu.js
Atala: OK

Atala: I think i didnt understand the code because the variable was badly named, it is indeed the "new value" which was called "old"

I moved it to utils.js for other usage

Issue #4331, opened 2024-05-23
Introduce Saved orders page


vladimir-8:
1. Dispatch; 'Delivery' page: Add a button: `Duplicate an order`: opens "Create a new order" page while coping/pasting data from the current order 

PR: https://github.com/coopcycle/coopcycle-web/pull/4474

2. Dispatch; 'Delivery' page: Add `save order` option to right side hand of local commerce web page orders (only for Dispatch)

![image](https://github.com/coopcycle/coopcycle-web/assets/77277854/230d0b9c-9347-4584-a448-b20658b8a70a)

PR: https://github.com/coopcycle/coopcycle-web/pull/4489

3. Dispatch; 'Store' page: Add a page with the list of saved orders

![Screenshot 2024-06-27 at 15 35 01](https://github.com/coopcycle/coopcycle-web/assets/70273239/a097618b-0bcd-44f2-be38-c8cebf95b056)

PR: https://github.com/coopcycle/coopcycle-web/pull/4489

Paul-Eraman-CoopCycle:
as discussed, this is more like "favorites" since everything is in practice saved

PR #4474, opened 2024-06-29
'Duplicate an order' button

vladimir-8:
Added a button: `Duplicate an order`: opens "Create a new order" page while coping/pasting data from the current order (no.1 in https://github.com/coopcycle/coopcycle-web/issues/4331 )

Screenshot 2024-07-04 at 12 28 02


templates/delivery/form.html.twig
```diff
@@ -116,11 +116,18 @@
...
-    <div class="col-md-5">
-      <button id="delivery-submit" type="submit" class="btn btn-block btn-lg btn-primary h-100">
+    <div class="col-md-5 delivery__form__buttons">
+      <button id="delivery-submit" type="submit" class="btn btn-block btn-lg btn-primary">
         <i class="fa fa-spinner fa-spin hidden" id="loader"></i>
         {% trans %}basics.save{% endtrans %}
       </button>
+      {% if is_granted('ROLE_ADMIN') and not is_new and delivery.order is not empty %}
+        <a href="{{ path('admin_store_delivery_new',
+          { id: delivery.store.id, frmrdr: delivery.order|hashid(16) }
```
r0xsh, line 126: `delivery.store` is only set for "lastmile" deliveries. This fields is empty for foodtech

r0xsh: Which lead to this error for foodtech deliveries
image

vladimir-8: Fixed in 5ef7b02f81b01894d4031360b8d65a0561da98d7

HugoLCR:
I have a problem with the added function:
the date and time are not respected -> my duplicate delivery switches to the next working day at the time of duplication

Paul-Eraman-CoopCycle:
Hi hugo, could you be more specific? Like, this order was supposed to be X time / day but instead was Y time / day

Or record a gif using licecap?

HugoLCR:
I just need a cloned order, when I duplicate. Not a new order, on the next date.

Paul-Eraman-CoopCycle:
I think the way it is working is the expected behavior. You can modify the day once the order is created no?

Remember, it has to work for multiple different types of situations at once

what types of orders are you copying?

HugoLCR:
I need to duplicate order from the B2B form 3-4x per week, for urgent customer needs, or poorly estimated volumes (2 services).
This meets urgent needs.

Paul-Eraman-CoopCycle:
That can be done by just adding another dropoff at the end with the original address

Paul-Eraman-CoopCycle:
"I need to duplicate order from the B2B form 3-4x per week, for urgent customer needs, or poorly estimated volumes (2 services).
This meets urgent needs."

I dont understand, can we do a video call next week?

HugoLCR:
I'll detail it for you tomorrow

HugoLCR:
https://github.com/user-attachments/assets/5afe4a2c-5b3b-45aa-aa1c-331f6f94bb40

I actually think I didn't make myself understood well.

I just need a duplicate with more :

- a same date
- a same hour
- a same price

PR #4479, opened 2024-07-02
As a dispatcher, I can show a line for tours + the line has a specific color & markers change colors

Atala:


js/shared/src/logistics/redux/selectors.js
```diff
@@ -84,6 +84,17 @@ export const selectAllTours = createSelector(
   (allTours) => allTours
 )
 
+export const selectTourPolylines = createSelector(
+  selectAllTours,
+  (allTours) => {
+    let polylines = {}
+    allTours.forEach(tour => {
```
r0xsh, line 91: nit picking but the use of `reduce()` would be cleaner here

r0xsh (approved):
LGTM

js/shared/src/logistics/redux/selectors.js
Atala: I use an object as accumulator and i set the key when iterating?

PR #4482, opened 2024-07-03
Move the search input from right panel to top bar

Atala:
Built on top of #4479


r0xsh (commented):
Overall okay, but i fell it would be way cleaner and simpler to use [auto-complete component from antd](https://4x.ant.design/components/auto-complete/)

PR #4489, opened 2024-07-04
Add the ability for admins to bookmark/save orders

vladimir-8:
Added: Dispatch; 'Delivery' page: Add save order option to right side hand of local commerce web page orders (only for Dispatch); no 2 in https://github.com/coopcycle/coopcycle-web/issues/4331

Screenshot 2024-07-05 at 14 13 38

Added: Dispatch; 'Store' page: Add a page with the list of saved orders; no 3 in https://github.com/coopcycle/coopcycle-web/issues/4331

Screenshot 2024-07-05 at 14 13 42

Screenshot 2024-07-05 at 14 13 48


src/Controller/Utils/StoreTrait.php
```diff
@@ -560,6 +587,45 @@ public function storeDeliveriesAction($id, Request $request, PaginatorInterface
...
+
+        $qb = $deliveryRepository->createQueryBuilder('d')
+            ->where('d.store = :store')
+            ->join(Order::class, 'o', Expr\Join::WITH, 'o = d.order')
+            ->join(Tagging::class, 'tagging', Expr\Join::WITH, 'tagging.resourceId = o.id AND tagging.resourceClass = :orderResourceType')
+            ->join(Tag::class, 'tag', Expr\Join::WITH, 'tag = tagging.tag')
+            ->andWhere('tag.slug = :tagName')
+            ->setParameter('store', $store)
+            ->setParameter('orderResourceType', 'AppBundle\Entity\Sylius\Order')
+            ->setParameter('tagName', '__bookmark')
```
Atala, line 616: add orderBy Id DESC so it is ordered from newest to latest ?

templates/store/layout.html.twig
```diff
@@ -51,6 +51,11 @@
             {{ 'stores.deliveries'|trans }}
           </a>
         </li>
+        <li role="presentation">
+          <a href="{{ path('admin_store_deliveries_bookmarks', { id: store.id }) }}">
+            {{ 'stores.deliveries.bookmarks'|trans }}
```
Atala, line 56: would that be possible to add a small help text?

translations/messages.en.yml
```diff
@@ -1284,6 +1285,7 @@ adminDashboard.nonprofits.confirm_delete: Are you sure you want to delete this n
 order.cart.donation.disclaimer: We offer you to donate %percentage% of the revenue from your order to one of our partner associations. You will not pay more.
 form.checkout_address.nonprofit.placeholder: Choose your favorite nonprofit
 form.delivery.arbitrary_price.label: Define the price manually
+form.delivery.bookmark.label: Bookmark
```
Atala, line 1288: Don't forget to add french, spanish, italian, german translations at least! - we use chatgpt to do that with Antoine

Atala (commented):
hello,

regarding the design and the use of tags to store the "bookmarked" info, I am not 100% sure it is the best solution; it feels kind of hacky to me - but maybe it is! ^^ it would have been worth it to discuss it has a team I think. but let s go this way if we need to get this out fast!

It didnt work for me the first try but after that it worked all the time so... I guess it is good.

I think regarding the UI/UX and the way we want people to use this feature, I think we should improve a little bit on the display so pple understand and it is more widely used:
- explain what the bookmark is for (help text)
- on an existing order the "duplicate order" button should be more visible. like in an action bar or something
- we should add the functionality to duplicate order in the list view (both orders and saved orders)
these are just some thoughts that can change!

Atala:
Actually it crashes the dispatch map : `Uncaught TypeError: can't access property "length", task.group.tags is undefined`

src/Serializer/Csv/TaskNormalizer.php
```diff
@@ -46,6 +46,10 @@ public function denormalize($data, $class, $format = null, array $context = arra
                 unset($item['address']['telephone']);
             }
 
+            if (isset($item['tags'])) {
```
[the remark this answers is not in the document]
vladimir-8: Some tests started to fail after I added types in TaggableTrait: `protected array $tags = [];`  so I parse the string into an array explicitly here

templates/store/layout.html.twig
vladimir-8: Added an empty state message. Will check on content with Vincent next week

PR #4506, opened 2024-07-10
Implement Paygreen

alexsegura:
📋 **TODO**

- [x] Add a page that sums up Paygreen transactions that have been paid by meal voucher, and need invoicing.


js/app/components/Payment/paygreen.scss
r0xsh, line 1: Is this needed or it can be cleaned out ?

src/Controller/Utils/RestaurantTrait.php
```diff
@@ -1882,4 +1771,123 @@ public function restaurantImageFromUrlAction($id, Request $request,
             ['imageName' => $restaurant->getBannerImageName()]
         );
     }
+
+    public function mealVouchersTransactionsAction(
```
r0xsh, line 1775: This function seems to only work with Edenred, Conecs payments won't appear in the export

src/Controller/Utils/RestaurantTrait.php
```diff
@@ -1882,4 +1771,123 @@ public function restaurantImageFromUrlAction($id, Request $request,
...
+                'Completed at',
+                'Total amount',
+                'Edenred amount',
+                'Platform fee',
+            ];
+
+            $records = [];
+            foreach ($hash[$exported] as $order) {
+
+                $edenredPayment = $order->getLastPaymentByMethod('EDENRED', PaymentInterface::STATE_COMPLETED);
```
r0xsh, line 1863: I'm pretty sure this will crash for conecs, swile and retroflash since edenredPayment will be `null`

src/Service/PaygreenManager.php
```diff
@@ -0,0 +1,281 @@
...
+        $po = $this->getPaymentOrder($payment->getPaygreenPaymentOrderId());
+
+        // If there are multiple operations, the parent payment order might already have been captured
+        if ($po['status'] !== 'payment_order.authorized') {
+            return;
+        }
+
+        $response = $this->paygreenClient->capturePaymentOrder($payment->getPaygreenPaymentOrderId());
+        if ($response->getStatusCode() !== 200) {
+            throw new \Exception('Could not capture payment');
```
r0xsh, line 54: Neat-picking but i think it always a good idea to create custom Exception class for integrations. Here for example it'll be helpful for debug purposes to have a exception named `PaygreenExcetpion`

src/Controller/Utils/RestaurantTrait.php
r0xsh: Yup, can confirm.
Exporting meal vouchers payments that are not exclusively Edenred lead to a crash
image

PR #4512, opened 2024-07-16
Add barcodes generation

r0xsh:
Linked with: https://github.com/coopcycle/coopcycle-app/pull/1834


src/Action/Task/Note.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace AppBundle\Action\Task;
+
+use AppBundle\Entity\Task;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Security\Core\Security;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class Note extends Base {
```
Atala, line 11: maybe it should be called AppendToComment ?

src/Controller/BarcodeController.php
```diff
@@ -0,0 +1,156 @@
...
+        $task = $this->getDoctrine()
+            ->getRepository(Task::class)
+            ->findByBarcode($barcode->getRawBarcode());
+
+        if (is_null($task)) {
+            return $this->json(['error' => 'No data found.'], 404);
+        }
+
+        $client_action = null;
+        if ($task->isAssignedTo($this->getUser())) {
```
Atala, line 61: where is unassignement done?

Atala, line 61: - does not seem to correspond to the expected behavior described above
- also please write code in the form
if... else if ... else if : I think it is more understandable this way to get the different cases

src/Controller/BarcodeController.php
```diff
@@ -0,0 +1,156 @@
...
+            $client_action = 'ask_unassign';
+        } elseif ($task->isAssigned()) {
+            $client_action = 'ask_assign';
+        }
+
+        if (!$task->isAssignedTo($this->getUser())) {
```
Atala, line 67: to check : is this dead code?

src/Controller/BarcodeController.php
```diff
@@ -0,0 +1,156 @@
...
+            if (is_null($task->getDelivery())) {
+                $task->assignTo($this->getUser());
+            } else {
+                $task->getDelivery()->assignTo($this->getUser());
+            }
+        }
+
+        $this->taskManager->scan($task);
```
Atala, line 75: this will send the "TASK SCANNED" message through EDIFACT to the transporter? (see command handler)
in this case I am not 100% sure it is good to have such side effect here (because the endpoint is mostly about assign/unassign) so maybe move it to a different endpoint for separation of concern

tell me if i a misunderstood things

src/Domain/Task/Handler/ScanBarcodeHandler.php
```diff
@@ -0,0 +1,46 @@
...
+    public function __invoke(ScanBarcode $command): void
+    {
+        $task = $command->getTask();
+
+        $this->eventRecorder->record(new Event\TaskBarcodeScanned($task));
+
+
+        $edi = $task->getImportMessage();
+        if (!is_null($edi)) {
+            //TODO: Send edifact message when scanned
```
Atala, line 30: TODO before merging I think?

src/Entity/TaskRepository.php
```diff
@@ -73,4 +75,26 @@ public function findBySubscriptionAndDate(Task\RecurrenceRule $subscription, Dat
...
+        if ($barcode->isInternal() && $barcode->getEntityType() === Barcode::TYPE_TASK) {
+            return $this->find($barcode->getEntityId());
+        }
+
+        /* TODO: I'm using raw sql here, but doctrine doesn't support json fields and i don't
+                 want to install doctrine json functions just for this
+            @see https://github.com/ScientaNL/DoctrineJsonFunctions
+        */
+        $conn = $this->getEntityManager()->getConnection();
+        $sql = "SELECT id FROM task WHERE metadata->>'barcode' = :barcode";
```
Atala, line 92: Check if you get the correct function here https://github.com/martin-georgiev/postgresql-for-doctrine/tree/main/src/MartinGeorgiev/Doctrine/ORM/Query/AST/Functions

then add the function like this https://github.com/coopcycle/coopcycle-web/blob/master/app/config/config.yml#L335

src/Transporter/ImportFromPoint.php
```diff
@@ -57,6 +57,7 @@ public function import(
         $task->setAddress($address);
         $task->setComments($point->getComments());
         $task->setMetadata('imported_from', $imported_from);
+        $task->setMetadata('barcode', $point->getId());
```
Atala, line 60: will it work if we have one barcode per parcel? in this case the barcode should be saved on `TaskPackage` object

src/Utils/Barcode/BarcodeUtils.php
```diff
@@ -0,0 +1,74 @@
...
+            Barcode::TYPE_TASK, $task->getId()
+        );
+
+        return self::parse($code);
+    }
+
+    /**
+     * @return Barcode[]
+     */
+    public static function getBarcodeFromPackage(Package $package, int $start = 0): array {
```
Atala, line 61: getBarcode**S**FromPackage

src/Utils/Barcode/BarcodeUtils.php
```diff
@@ -0,0 +1,74 @@
...
+            $matches['package'][0] ?? null,
+            $matches['unit'][0] ?? null
+        );
+    }
+
+    /**
+     * @return Barcode[]|Barcode|null
+     * @param object $entity
+     */
+    public static function getBarcodeFromEntity(object $entity): array|Barcode|null {
```
Atala, line 37: I didnt found where it is called?

assets/react/controllers/BarcodesModal.jsx
```diff
@@ -0,0 +1,106 @@
...
+          <Col span={8}>
+            <PackageBarcode label={name} barcode={barcode} index={index + 1} />
+          </Col>
+        ))}
+      </Row>
+    </>
+  )
+}
+
+export default function ({ items }) {
```
Atala, line 82: maybe show it in the dispatch panel somewhere?

Atala (commented):
To be checked the following points : 
- getBarcodeFromEntity seems not to be called
- maybe we could display the barcode stuff in the dispatch somewhere?
- the logic in assign/unassign endpoint

src/Transporter/ImportFromPoint.php
r0xsh: There is only the possibility for "one barcode per task" in the EDIFACT specification. So in this context no issue

src/Controller/BarcodeController.php
```diff
@@ -0,0 +1,178 @@
...
+        private ManagerRegistry $doctrine,
+        private TaskManager $taskManager,
+        private BarcodeUtils $barcodeUtils,
+    )
+    { }
+
+
+
+    /**
+     * @Route("/api/barcode", name="barcode_api")
```
alexsegura, line 35: This shouldn't be mapped to a web controller

src/Domain/Task/Handler/ScanBarcodeHandler.php
```diff
@@ -0,0 +1,43 @@
...
+    { }
+
+    public function __invoke(ScanBarcode $command): void
+    {
+        $task = $command->getTask();
+
+        $this->eventRecorder->record(new Event\TaskBarcodeScanned($task));
+
+
+        // $edi = $task->getImportMessage();
```
Atala, line 25: we should create an issue to track this otherwise we will forget?

PR #4518, opened 2024-07-19
Add the ability to undo optimization

Atala:
undo-optim


vladimir-8 (approved):

vladimir-8:
Just a note, `Optimise` icon has a nice tooltip, but other icons don't

Screenshot 2024-07-19 at 14 37 47

src/Service/RouteOptimizer.php
```diff
@@ -94,7 +94,10 @@ function($desc) use ($iriConverter, $items) {
             })->toArray()
         );
 
-        return $res;
+        return [
+            "solution" => $res,
+            "unassigned_count" => $data["summary"]["unassigned"]
+        ];
```
alexsegura, line 100: Maybe we could return an object here.

src/Action/TaskList/Optimize.php
```diff
@@ -3,27 +3,36 @@
...
-        return $data;
+        return new JsonResponse([
+            "solution" => $this->taskListNormalizer->normalize($data, 'jsonld', [
+                'resource_class' => TaskList::class,
+                'operation_type' => 'item',
+                'item_operation_name' => 'get',
+                'groups' => ['task_list']
+            ]),
+            "unassigned_count" => $optim["unassigned_count"]
+        ]);
```
alexsegura, line 36: We should return an object here, and let the serializer do its job.

src/Action/TaskList/Optimize.php
```diff
@@ -3,27 +3,36 @@
...
 
-        return $data;
+        return new JsonResponse([
+            "solution" => $this->taskListNormalizer->normalize($data, 'jsonld', [
+                'resource_class' => TaskList::class,
+                'operation_type' => 'item',
+                'item_operation_name' => 'get',
+                'groups' => ['task_list']
+            ]),
+            "unassigned_count" => $optim["unassigned_count"]
```
alexsegura, line 35: In other parts of the code, JSON responses are generally camel case. 

Should be `unassignedCount`

Atala:
Merged manually

Issue #4519, opened 2024-07-19
Change "Spinner" component to use "loader" class so it is animated


Atala:
Instead of static font awesome icon; we could use <span class="loader loader--dark></span> which is animated

PR #4520, opened 2024-07-19
chore: Update LoadingIcon component to use new loader class #4519

kumaryogesh17:
#4519,
Updated LoadingIcon components to use new loader class `<span className="loader loader--dark"></span>`


vladimir-8 (approved):

vladimir-8:
Thanks a lot @kumaryogesh17 👍

PR #4548, opened 2024-08-01
stripe api key issue

vladimir-8:
Fixes https://github.com/coopcycle/coopcycle/issues/62


src/Service/StripeManager.php
```diff
@@ -356,6 +354,8 @@ public function createSetupIntent(PaymentInterface $payment, $paymentMethod)
 
     public function attachPaymentMethodToCustomer(PaymentInterface $payment)
     {
+        $this->configure();
```
vladimir-8, line 357: @Atala this line is the actual fix. If `configure` method is not called an API key is not set on the Stripe client. 

All other changes is an attempt to limit such calls to the `StripeManager` class to avoid similar mistakes and avoid the need to "not forget" to call a `configure`method everywhere in the codebase

Atala (commented):
hello 

ok i understand better now

is it possible to move other configure() calls into this StripeManager class (and maybe make the method private/internal). I see two remainings :
 - https://github.com/coopcycle/coopcycle-web/blob/master/src/Controller/PaymentController.php#L78
  - https://github.com/coopcycle/coopcycle-web/blob/master/src/Controller/StripeController.php#L178

also maybe rename this method to setStripeClientKeys() or something like this because we have this `configure` which is very vague + `configurePayment`, `configureCreateIntentPayload` which are different things and actually not related to call to stripe

also maybe add a comment on the method : "Please call this before using the Stripe client lib".

vladimir-8:
> is it possible to move other configure() calls into this StripeManager class (and maybe make the method private/internal). I see two remainings

These two are different because they use Stripe directly, and I don't want to touch them now

vladimir-8:
> also maybe rename this method to setStripeClientKeys() or something like this because we have this configure which is very vague + configurePayment, configureCreateIntentPayload which are different things and actually not related to call to stripe
> 
> also maybe add a comment on the method : "Please call this before using the Stripe client lib".

done

Atala (approved):
it is ok for me

i can not test payment right now locally because i have some config issues i need to solve, i ll try to fix that ASAP, but I think we are good for beta testing on demo and maybe a few coops

PR #4554, opened 2024-08-08
In foodtech, the (global) delay is not applied correctly when presenting choices to the customer

Atala:


src/Utils/OrderTimeHelper.php
```diff
@@ -37,24 +37,34 @@ public function __construct(
...
 
-    private function filterChoices(OrderInterface $cart, array $choices): array
+    private function filterChoices(OrderInterface $cart, array $choices, $fulfillmentMethod): array
     {
         $choicesLogged = 0;
         $acceptedChoicesLogged = 0;
+        $orderingDelayMinutes = $this->getOrderingDelayMinutes($fulfillmentMethod->getOrderingDelayMinutes());
 
-        return array_filter($choices, function (TsRangeChoice $choice) use ($cart, &$choicesLogged, &$acceptedChoicesLogged) {
+        $this->logger->info('here');
```
vladimir-8, line 46: remove? 😄

src/Utils/OrderTimeHelper.php
```diff
@@ -37,24 +37,34 @@ public function __construct(
...
 
-            $result = $this->shippingDateFilter->accept($cart, $choice->toTsRange());
+        return array_filter($choices, function (TsRangeChoice $choice) use ($cart, $orderingDelayMinutes, &$choicesLogged, &$acceptedChoicesLogged) {
 
-            if ($choicesLogged < self::MAX_CHOICES_LOGGED && $acceptedChoicesLogged < self::MAX_ACCEPTED_CHOICES_LOGGED) {
```
vladimir-8, line 49: that was added to limit the number of choices logged. If we log all possible choices for the next 7 days on each call, we'll get gigabytes of logs on busy instances

src/Utils/ShippingDateFilter.php
```diff
@@ -52,11 +52,18 @@ public function accept(OrderInterface $order, TsRange $range, \DateTime $now = n
...
-            $this->logger->info(sprintf('ShippingDateFilter::accept | preparation time "%s" is in the past',
-                $preparation->format(\DateTime::ATOM)),
+        // if a customer wants a dropoff at 12:00, and preparation time + shipping time is 30 minutes, I need to order at 11:30 at least
+        // if the dispatcher adds an delay of 30min
+        // then I want here to filter out all of the possible dropoffs between 11:30 and 12:00
+        // so I need to substract the delay from the preparation time
+        $preparationWithDelay = clone $preparation;
+        $preparationWithDelay = $preparationWithDelay->sub(date_interval_create_from_date_string(sprintf('%s minutes', $orderingDelayMinutes)));
+
+        if ($preparationWithDelay <= $now) {
```
vladimir-8, line 62: It's exactly the same logic, but maybe it's a bit easier to understand:

```
$preparationCanStartAt = clone $now;
if ($orderingDelayMinutes > 0) {
    $preparationCanStartAt = $preparationCanStartAt->add(date_interval_create_from_date_string(sprintf('%s minutes', $orderingDelayMinutes)));
}
if ($preparation <= $preparationCanStartAt) {
```

vladimir-8 (approved):
Nice solution. Just a few small suggestions:

src/Utils/OrderTimeHelper.php
Atala: yes yes I saw

i moved it out to debug but I surely should have not commit that... :sweat_smile: thanks for spotting this out

src/Utils/ShippingDateFilter.php
Atala: cool thx it is nice

in this case which variable do we need to pass on l.82 according to you? (where I added a comment "I am not sure") I think it is still "preparationWithDelay" no?

vladimir-8: I'd say `$preparation`. We'll filter out all options before `$preparationCanStartAt` on line 62, and on line 82 we only need to check that the restaurant is still opened at the expected `$preparation` time

Atala: i gave it a second thought, it is `$preparationWithDelay` because delay is the delay bar but also "prior notice" setting on the restaurant page, so you want the restaurant to be open "some time in advance before the preparation can start"

src/Utils/OrderTimeHelper.php
```diff
@@ -37,23 +37,33 @@ public function __construct(
     {
     }
 
-    private function filterChoices(OrderInterface $cart, array $choices): array
+    private function filterChoices(OrderInterface $cart, array $choices, $fulfillmentMethod): array
```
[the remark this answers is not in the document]
Atala: yes i think also, i ll double check

src/Utils/OrderTimeHelper.php
```diff
@@ -76,7 +86,7 @@ private function roundUp($n, $x = 5): int
         return (int) $value;
     }
 
-    private function getExtraTime(): int
+    private function getShippingDelay(): int
```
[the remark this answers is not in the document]
Atala: yes

PR #4561, opened 2024-08-09
fixed: avoid calling `sync()` action twice

vladimir-8:
fixed: avoid calling sync() action twice on loading a restaurant page: http://*/en/restaurant/3-burger-bar

Which leads to a double POST`/*/restaurant/*/cart` requests with in turn leads to a creation of double adjustments (delivery fees)

Screenshot 2024-08-09 at 12 12 50

❗️you will probably see double POST requests in the development environment, that due to the `StrictMode`

### DIFF
--- a/js/app/restaurant/components/Order/index.js
+++ b/js/app/restaurant/components/Order/index.js
@@ -4,28 +4,21 @@ import Sticky from 'react-stickynode'
 import classNames from 'classnames'
 
 import { sync } from '../../redux/actions'
-import {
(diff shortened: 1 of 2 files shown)


Atala (approved):
could fix the global issue about duplicated delivery fees?

vladimir-8:
> could fix the global issue about duplicated delivery fees?
At least partially. I remember I saw that it happens from the mobile app sometimes, but that fix will not fix that

PR #4606, opened 2024-09-07
Export tasks command w/ Apache Parquet support

r0xsh:


src/Command/ExportTasksCommand.php
```diff
@@ -0,0 +1,165 @@
...
+
+        /** @var HandledStamp $handledStamp */
+        $handledStamp = $envelope->last(HandledStamp::class);
+        $export = $handledStamp->getResult();
+
+        switch ($input->getOption('format')) {
+            case 'parquet':
+                $export = $this->csv2parquet($export);
+                break;
+            case 'csv': break;
```
alexsegura, line 61: It's actually easy to also implement CSV, as it's what you get from the export code.

src/Command/ExportOrdersCommand.php
```diff
@@ -0,0 +1,160 @@
...
+            FlatColumn::float('platform_fee'),
+            FlatColumn::float('refunds'),
+            FlatColumn::float('net_revenue')
+        );
+
+        $writer = new Writer(Compressions::GZIP);
+
+        // Temporary file, no choice the library API is poorly designed.
+        // and i'm too lazy to implement it a DestinationStream class.
+        $tmp = sys_get_temp_dir() . '/' . uniqid() . '.parquet';
```
alexsegura, line 148: You can use the [tempnam](https://www.php.net/tempnam) function for this

src/Command/ExportOrdersCommand.php
```diff
@@ -0,0 +1,160 @@
...
+            FlatColumn::float('tip'),
+            FlatColumn::float('promotions'),
+            FlatColumn::float('total_products_excl_vat'),
+            FlatColumn::float('total_products_incl_vat'),
+            FlatColumn::float('total_vat'),
+            FlatColumn::float('stripe_fee'),
+            FlatColumn::float('platform_fee'),
+            FlatColumn::float('refunds'),
+            FlatColumn::float('net_revenue')
+        );
```
alexsegura, line 142: It would be preferable to have the exact same column names between CSV and Parquet. This way Parquet and CSV files are interchangeable seamlessly. 
I agree the current column names are messy, but they are known by lots of people, and we can re-map them in Cube with clean names. 
Then, if we want to rename the columns, we rename them in the handler that actually does the export, and the renaming is propagated everywhere.

src/Command/ExportOrdersCommand.php
```diff
@@ -0,0 +1,160 @@
...
+            ->setName('coopcycle:export:orders')
+            ->setDescription('Export orders');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        if (!$this->lock()) {
+            $output->writeln('The command is already running in another process.');
+            return Command::FAILURE;
+        }
```
alexsegura, line 40: Nice!

src/MessageHandler/ExportOrdersHandler.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function __invoke(ExportOrders $message): ?string
+    {
+        $stats = new RestaurantStats(
+            $this->entityManager,
+            $message->getFrom()->setTime(0, 0, 1),
+            $message->getTo()->setTime(23, 59, 59),
+            null,
+            $this->paginator,
+            $this->container->getParameter('kernel.default_locale'),
```
alexsegura, line 34: We should be able to override the locale via a property in the `ExportOrders` message. 
Because for the exports we are interested in, it is easier to have a stable language, i.e English.

src/Command/ExportTasksCommand.php
alexsegura: Also, with CSV, this command can also be used to send exports by email.

src/Command/ExportTasksCommand.php
```diff
@@ -0,0 +1,165 @@
...
+        $envelope = $this->messageBus->dispatch(new ExportTasks(
+            $this->parseDate($input->getOption('date-start')),
+            $this->parseDate($input->getOption('date-end'))
+        ));
+
+        /** @var HandledStamp $handledStamp */
+        $handledStamp = $envelope->last(HandledStamp::class);
+        $export = $handledStamp->getResult();
+
+        switch ($input->getOption('format')) {
```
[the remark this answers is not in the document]
r0xsh: Yep i agree, i already mixed up the file name and the format option in my tests :melting_face:

src/Command/ExportOrdersCommand.php
r0xsh: No i tried, but the library api is really annoying; [tempnam](https://www.php.net/tempnam) is actually creating the file, so the library complains the file already exists :face_exhaling:

src/Command/ExportOrdersCommand.php
r0xsh: I tried this morning but i got a issue, there is actually duplicate column name, which seems to not being supported by parquet format file
image

src/Command/ExportTasksCommand.php
r0xsh: > It's actually easy to also implement CSV, as it's what you get from the export code.
Yeah it's already the case, the command support CSV natively, so since there is no need to convert it into another format, the switch case branch for CSV do nothing


> Also, with CSV, this command can also be used to send exports by email.
Yep i designed the functions to be easily extensible to do so

alexsegura (approved):
Still can be improved to try to have a 1-1 interop between CSV & Parquet, but let's merge and try it for real 📈

Issue #4341, opened 2024-05-28
Double refund requests


vladimir-8:
**Describe the bug**

Sometimes, when creating a refund, double refund requests are created (both on the platform and on Stripe)

**Additional context**

https://coopcycle.slack.com/archives/C4UT4DGQ3/p1716060037373999

Look at the timestamps (1s interval between reimbursement requests)

Problematic orders:

112048 on 10/07/2023 (lcr)
140049 on 2024-03-04 (lcr)
147745 on 2024-04-13 (lcr)

vladimir-8:
Could somehow caused by this loop? https://github.com/coopcycle/coopcycle-web/blob/master/src/Controller/AdminController.php#L326

Atala:
Duplicate of #4608

Issue #4608, opened 2024-09-08
double customer reimbursement


HugoLCR:
<img width="959" alt="Capture d’écran 2024-09-08 212853" src="https://github.com/user-attachments/assets/6c549715-16e8-48ef-a6bb-6b434feadaed">
**Describe the bug**
When I refund amounts from a customer order via coopcycle (Stripe API connection), I often get a 2X refund (with just one click)

**To Reproduce**
1. go to an order
2. repay an amount by coopcycle

**Expected behavior**
I want it to reimburse the chosen amount x1

**Screenshots**
![Uploading image.png…]()

**Device (please complete the following information):**
 - Windows 11
 - Chrome

Atala:
order : https://lcr.coopcycle.org/admin/orders/178800

there is only one POST sent here at 21:18 https://app.datadoghq.com/apm/traces?query=env%3Aprod%20service%3Asymfony%20resource_name%3Aadmin_order%20%40http.host%3Alcr.coopcycle.org%20%40http.method%3APOST&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=true&graphType=flamegraph&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=all&storage=hot&view=spans&start=1725746400000&end=1725832740000&paused=true

trace of the POST in datadog https://app.datadoghq.com/apm/trace/66ddf868000000009efb5b8071d7cd86?colorBy=service&env=prod&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=11455850684370898310&spanViewType=metadata&timeHint=1725823080572.4211

no idea what is happening yet

Atala:
@HugoLCR 

is it a new thing? if yes since when?
how many times did it happen?

tcmjlnt:
Hey @Atala 

Here are other occurrences we reported in the past : 

- Order [2EGG (#112048)](https://lcr.coopcycle.org/admin/orders/112048) 2023-07-10
- Order [4G91 (#147745)](https://lcr.coopcycle.org/admin/orders/147745) 2024-04-13
- Order [48RH](https://lcr.coopcycle.org/admin/orders/140049) (#140049) 2024-03-04

Here is the thread in Slack
[...]

PR #4614, opened 2024-09-11
Implement multiple payments for Edenred

### LINKED ISSUE #4537: Properly implement multiple payments per order
Sylius, the e-commerce framework we use, natively implements having [multiple payments for each order](https://docs.sylius.com/en/1.12/book/orders/orders.html#completing-a-payment-with-a-state-machine-transition). 
However, in our in-house implementation of Sylius, we only allow one payment per order. 

Some things that are wrong:
* lots of pieces of code rely on the [`getLastPayment()`](https://github.com/search?q=repo%3Acoopcycle%2Fcoopcycle-web%20getLastPayment&type=code) method. 
* we have a `PaymentMethod` named `EDENRED+CARD` to circumvent this limitation, when there should be 2 payments, on with `CARD`, and another with `EDENRED`

This is a problem to implement Paygreen (#3042), because a user may want to pay with Edenred, and complete payment with credit card, using Paygreen.  

---

In Sylius Core, this is implemented by having a [`paymentState` property](https://github.com/Sylius/Sylius/blob/6f732828e803f3462cbddf42d932063499686b34/src/Sylius/Component/Core/Model/Order.php#L64) and its associated [state machine](https://github.com/Sylius/Sylius/blob/6f732828e803f3462cbddf42d932063499686b34/src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_order_payment.yml)

alexsegura:
Fixes #4537

This pull request changes the behavior when selecting the Edenred payment method, by creating 2 payments, one for Edenred with the max amount possible, and the rest to be completed by credit card. 

Also, in most of the rest of the code, it makes a loop on payments, instead of relying on the `getLastPayment` method.


src/Controller/OrderController.php
```diff
@@ -390,24 +388,7 @@ public function paymentAction(Request $request,
...
-                // TODO Make sure we are using Stripe, not MercadoPago
-                $stripeManager->configurePayment($payment);
-
-                return $this->render('order/payment.html.twig', array_merge($parameters, [
-                    'form' => $form->createView(),
-                    'error' => $error,
-                    'payment' => $payment,
-                ]));
-            }
+            // TODO Check if all $payments have succeeded
```
Atala, line 391: needs to be done before merging no?
at least port the old stripe code here if it is a stripe payment ?

src/Domain/Order/Handler/CheckoutHandler.php
```diff
@@ -23,56 +23,59 @@ public function __construct(
...
+            $this->eventRecorder->record(new Event\CheckoutSucceeded($order));
             return;
         }
 
-        // TODO Check if $payment !== null
+        $payments = $order->getPayments()->filter(
+            fn (PaymentInterface $payment): bool => $payment->getState() === PaymentInterface::STATE_CART);
 
-        $data = $command->getData();
+        // TODO Make sure card payment is always *BEFORE* Edenred payment
```
Atala, line 43: it is just sorting the payments array right?

Atala:
crash 1 :
- make a payment with edenred + card
- refund edenred (not sure if it is mandatory for the crash)
- refund (partially) the card payment

-> crash

> `AppBundle\Entity\Sylius\Payment::addRefund(): Argument #1 ($amount) must be of type int, null given, called in /var/www/html/src/Payment/Gateway/Edenred.php on line 29
`

crash-refund-1

Atala (commented):
hello,

I think the code is fine. I fixed the migration.

Please see my comment about the crash on refund https://github.com/coopcycle/coopcycle-web/pull/4614#issuecomment-2345830825

Also I think you should take the time to fix the two FIXME i left comment on before merging (show stripe card error and have edenred fix)

the 3rd fixme "Do not hardcode this here" i dont really understand the problem so maybe write more precisely what the problem is.

alexsegura:
The refund problem is fixed in f74674c9c7706cac381f29d8fc471d7420a45d4f, and will also fix #4341 & #4608

src/Domain/Order/Handler/CheckoutHandler.php
alexsegura: I added this comment because it was [hardcoded like this](https://github.com/coopcycle/coopcycle-web/blob/f74674c9c7706cac381f29d8fc471d7420a45d4f/src/Payment/Gateway.php#L52-L74) before. 

I don't remember exactly why, probably because it's easier to re-make a card authorization than an Edenred authorization (Edenred authorizations **MUST** be [cancelled when the order is refused/cancelled](https://github.com/coopcycle/coopcycle-web/blob/f74674c9c7706cac381f29d8fc471d7420a45d4f/src/Domain/Order/Reactor/CancelEdenredTransaction.php), to make sure we release the money).

PR #4622, opened 2024-09-13
Support cart from multiple restaurants in Platform catering

vladimir-8:


vladimir-8:
Not needed at the moment, as we are not going to support multiple restaurants in the same cart for now

PR #4631, opened 2024-09-16
fix: ordering from more than 1 restaurant

vladimir-8:
* fixed: don't allow ordering from more than one restaurant from the same cart (business context)
* fixed: e2e test for change restaurant modal
* added: e2e test for checkout flow in the business context


src/Sylius/Cart/RestaurantResolver.php
```diff
@@ -83,17 +82,20 @@ public function accept(OrderInterface $cart): bool
             return true;
         }
 
-        if (null !== $cart->getBusinessAccount()) {
-            return $cart->getBusinessAccount()->getBusinessRestaurantGroup()->getRestaurants()->contains($restaurant);
-        }
-
```
vladimir-8, line 89: That's the actual "fix"

Atala (approved):
Tested OK

PR #4641, opened 2024-09-18
Store & Restaurant config options

r0xsh:
Add basic crud to configure billing method


src/Form/RestaurantType.php
```diff
@@ -109,7 +110,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                         '%entity%' => 'restaurant',
                     ],
                     'help_html' => true,
-                ));
+                ))->add('billingMethod', ChoiceType::class, [
+                    'label' => 'form.billing_method.label',
+                    'help' => 'form.billing_method.help',
```
alexsegura, line 115: Missing translations

src/Form/RestaurantType.php
```diff
@@ -109,7 +110,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                    'choices' => [
+                        'form.billing_method.per_task' => 'per_task',
+                        'form.billing_method.total_percentage' => 'total_percentage',
```
alexsegura, line 118: Missing translations + maybe we should name the choices: 
- `unit`
- `percentage`

alexsegura (commented):
The billing method settings should be behind a feature flag

PR #4646, opened 2024-09-19
Feature/4618 business account more fields

vladimir-8:


src/Form/BusinessAccountType.php
```diff
@@ -21,21 +21,26 @@
...
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('name', TextType::class, ['label' => 'registration.step.company.name'])
+            ->add('name', TextType::class, [
+                'label' => 'registration.step.company.name'
+            ])
+            ->add('legalName', TextType::class, [
+                'label' => 'registration.step.company.legal_name',
+                'required' => !$this->authorizationChecker->isGranted('ROLE_ADMIN'),
```
r0xsh, line 38: Maybe they should be hidden for non-admin, no ?

src/Form/BusinessAccountType.php
```diff
@@ -21,21 +21,26 @@
...
+                'empty_data' => ''])
+            ->add('vatNumber', TextType::class, [
+                'label' => 'registration.step.company.vat_number',
+                'required' => !$this->authorizationChecker->isGranted('ROLE_ADMIN'),
```
r0xsh, line 42: same

r0xsh (commented):
Seems good :+1: 
Nice cleanup on the entity :)

src/Form/BusinessAccountType.php
vladimir-8: No, they are for `business manager` to setup. So an admin can skip them while creating a business account, but a `business manager` must enter them

PR #4649, opened 2024-09-20
Fix/4648 cash payment

vladimir-8:
Fixed:

1. Use `recomputeSingleEntityChangeSet` instead of `computeChangeSets`. I tested locally and was able to reproduce `ERROR:  bind message supplies 2 parameters, but prepared statement "" requires 20` error ([logs](https://app.datadoghq.com/logs?query=%40extra.request_id%3A881ac78b1c819f63e63600e14a07a51e%20&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Chost%2C%40extra.request_id%2Csource%2C%40logger.name&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1726248505166&to_ts=1726853305166&live=true) and [traces](https://app.datadoghq.com/apm/trace/66eb0ee000000000d5b3966f5da032cc?colorBy=service&env=prod&graphType=waterfall&panel_tab=flamegraph&shouldShowLegend=true&sort=time&spanID=18095358400174948562&spanViewType=metadata&timeHint=1726680800843.134&traceQuery=)). I don't fully understand why using `computeChangeSets` is an issue, but as we only update existing entities, `recomputeSingleEntityChangeSet` is enough, see comment in https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/events.html#onflush
2. Send notifications in `postFlush` event, `TaskUpdate` event sent down the line relies on the task ID field which is missing until the task is persisted.

It is not an issue if the order is paid with Stripe, because of this: https://github.com/coopcycle/coopcycle-web/issues/4650


Atala:
Tested OK, thanks for the fix

PR #4655, opened 2024-09-24
Add the possibility to auto-assign orders from a store

Atala:
auto-assign-to-rider

### DIFF
--- a/src/Action/TaskList/SetItems.php
+++ b/src/Action/TaskList/SetItems.php
@@ -17,45 +17,27 @@ use Symfony\Component\HttpFoundation\Request;
 final class SetItems
 {
     public function __construct(
-        private EntityManagerInterface $objectManager,
+        private EntityManagerInterface $entityManager,
         private UserManager $userManager,
         private TaskListManager $taskListManager,
         private TaskListNormalizer $taskListNormalizer
     )
-    {
-    }
-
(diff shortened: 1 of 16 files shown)


src/Action/TaskList/SetItems.php
```diff
@@ -17,45 +17,27 @@
...
-    }
+    {}
 
     public function __invoke(Request $request)
     {
         $date = new \DateTime($request->get('date'));
         $user = $this->userManager->findUserByUsername($request->get('username'));
 
-        $taskList = $this->getTaskList($date, $user);
+        $taskList = $this->taskListManager->getTaskListForUser($date, $user);
```
r0xsh, line 32: ~Maybe the `$date` should be the date of the delivery ?~

r0xsh: ~Like what happen if i create a delivery for the next week ?~

r0xsh (approved):
🚀

src/Action/TaskList/SetItems.php
Atala: mmm it is an unrelated change sorry
here we are in the endpoint that assigns task when on the dispatch panel so the $date is a parameter we passed (the date we are showing in the dispatch)

PR #4659, opened 2024-09-25
Add `applied_billing` in the exports

r0xsh:
Since the Foodtech and Lastmile shops are billed at different rates (yeah, I just figured that out thanks to @Atala), we need to include the shop types in the exports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,4 @@
 .phpactor.json
 
 .vscode
+.tool-versions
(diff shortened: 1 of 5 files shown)


alexsegura (commented):
I have a problem with the naming. 
IMHO "store type" doesn't look like the thing we want to describe. Maybe something like "billing channel"?

r0xsh:
What about `applied_billing` ?

Issue #4401, opened 2024-06-06
Improve performance : compute packages and weight ahead of time (in the database ?) to reduce the numbers of queries


Atala:
### What is the issue?

The Task object is linked to `packages` (many to many from package to task) and has `weight`. However now we want to display the following for delivery/orders :
 - `Pickup` : sum of weight of all the dropoffs, "sum of packages" of all the dropoffs
 - `Dropoff` : weight of the task (easy ^^), nicely serialized packages (like a JSON `{type: XL, quantity: 2}`)

However for now this is not precomputed, so we compute it each time we want to serialize a task in the normalizer ([ref](https://github.com/coopcycle/coopcycle-web/blob/master/src/Serializer/TaskNormalizer.php#L78-L109))

so for a dispatch panel with 300 tasks it adds ~600 queries. I created a custom endpoint to get all the tasks, so it will prefetch and "annotate the tasks" with the packages and weight data (https://github.com/coopcycle/coopcycle-web/blob/master/src/Action/Task/Collection.php#L19-L73) and it worked (loading time from 15s to 1s).

However the issue remains in part of the code. For example when you fire the `task:assigned` event  you may serialize ~30 tasks (when assigning a tour for example) and then you get ~60 queries more.

So I do think these fields (JSON packages and weight) should be precomputed when we load the tasks objects. Now they are two solutions : 
 - doctrine listener on task collections to get all of the thing (dont know how it would work in practice ??)
 - computed field in database and/or database view and/or stored procedures -> i think this is the way to go, like having an extra field on task in the database that is computed by postgres

What do you think? what would be the good solution?

alexsegura:
I think what impacts performance is not really the number of SQL queries, but the hydration of PHP objects. Serialization takes time, because not only it needs to load all the related objects (which happens at serialization time, because Doctrine load lazily those objects), but it also needs to hydrate them with all their internal dependencies, filling up the memory and slowing down the whole thing. 

I think we should keep the serializer simple, it should just serialize blindly what has been computed at previous stages. Maybe we could use a [custom Doctrine hydrator](https://medium.com/@thovandeth/creating-a-custom-hydrator-in-doctrine-and-symfony-for-enhanced-performance-with-dtos-f9f4fce8f035), that loads the packages and stuff as simple objects without any dependencies.

Atala:
[...]

Issue #4337, opened 2024-05-27
cash deliveries should have amount of money in the task info for the messenger


Paul-Eraman-CoopCycle:
when doing cash delivery the messenger must request money from the customer. These deliveries should have the amount of money in the pickup and dropoff task to facilitate this.

![image](https://github.com/coopcycle/coopcycle-web/assets/77277854/68700a88-2678-4200-8708-4d7bab46c73a)

vladimir-8:
It works only for food tech orders at the moment. It will not work for orders coming from a Delivery form (B2C customers) until https://github.com/coopcycle/coopcycle-web/issues/4746 is fixed.

PR #4672, opened 2024-10-01
Feature/4337 task extra info

vladimir-8:


src/Entity/TaskListRepository.php
```diff
@@ -0,0 +1,107 @@
...
+        $taskList = $this->findOneBy([
+            'courier' => $user,
+            'date' => $date,
+        ]);
+
+        if (null === $taskList) {
+            return null;
+        }
+
+        $taskIds = $taskList->getItems()->map(function (TaskList\Item $item) {
```
Atala, line 32: ! this will not work with tours. you should use $taskList->getTours()

src/Entity/TaskListRepository.php
```diff
@@ -0,0 +1,107 @@
...
+                't.doneAfter',
+                't.doneBefore',
+//                't.previous', //TODO
+//                't.next', //TODO
+                't.comments',
+                't.metadata',
+                'o.total AS orderTotal',
+                'org.name AS orgName',
+            ])
+            ->from(Task::class, 't')
```
Atala, line 53: do not forget to filter out cnacelled tasks

src/Serializer/MyTaskListNormalizer.php
```diff
@@ -0,0 +1,123 @@
...
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+class MyTaskListNormalizer implements NormalizerInterface
+{
+
+    public function __construct(
+        protected ObjectNormalizer $normalizer,
+        protected IriConverterInterface $iriConverterInterface,
+        private LoggerInterface $checkoutLogger
```
Atala, line 21: not `checkoutLogger` but `logger` ?

alexsegura (commented):
The idea is OK, but it shouldn't be linked to #4337, which can be fixed by a few lines of code in `TaskNormalizer`. Instead, it should be linked to another issue about fixing performance issues on the dispatch endpoints.

src/Entity/TaskListRepository.php
```diff
@@ -0,0 +1,115 @@
...
+                $task->getStatus(),
+                $task->getAddress(),
+                $task->getDoneAfter(),
+                $task->getDoneBefore(),
+                $task->getPrevious()?->getId(),
+                $task->getNext()?->getId(),
+                $task->getTags(),
+                $task->isDoorstep(),
+                $task->getComments(),
+                $task->getPackages()->map(function (Task\Package $package) {
```
Atala, line 82: it is wrong here `task.packages` depends on whether the task is a pickup or a dropoff
if pickup then the packages are the "sum" of the drop packages
see here https://github.com/coopcycle/coopcycle-web/blob/master/src/Action/Task/Collection.php#L11

Atala:
Related issues
coopcycle/coopcycle-app#1803
#4401

src/Entity/TaskListRepository.php
```diff
@@ -0,0 +1,184 @@
...
+                $task->getUpdatedAt(),
+                $task->getType(),
+                $task->getStatus(),
+                $task->getAddress(),
+                $task->getDoneAfter(),
+                $task->getDoneBefore(),
+                $task->getPrevious()?->getId(),
+                $task->getNext()?->getId(),
+                $task->getTags(),
+                $task->isDoorstep(),
```
Atala, line 132: for your information isDoorstep is soon to be deprecated (i removed it in an other branch)

src/Entity/TaskListRepository.php
```diff
@@ -0,0 +1,184 @@
...
+        $tasks = array_map(function ($row) use ($tasksByDeliveryId) {
+            $task = $row[0];
+            $deliveryId = $row['deliveryId'] ?? null;
+
+            $taskPackages = [];
+            $weight = null;
+
+            $tasksInTheSameDelivery = $deliveryId ? $tasksByDeliveryId[$deliveryId] : [];
+
+            if ($task->isPickup()) {
```
Atala, line 103: cool!
I think i had the reflex to write it in SQL as to go faster but I think it is cool if we move it in a utility function we can test it better and it will act as a spec :+1:

Atala:
i just add a note here not to forget to use the DTO on TaskListUpdated event (live update for riders)

features/tasks.feature
```diff
@@ -295,20 +219,14 @@ Feature: Tasks
             "doneBefore":"@string@.isDateTime().startsWith('2018-03-02T12:00:00')",
             "comments":"#bob",
             "updatedAt":"@string@.isDateTime()",
-            "isAssigned":true,
-            "assignedTo":"bob",
```
Atala, line 299: we don't need these in-app ?

src/Api/Dto/TaskPackageDto.php
```diff
@@ -0,0 +1,37 @@
...
+
+    public function __construct(
+        string $shortCode,
+        string $name,
+        int $averageVolumeUnits,
+        int $quantity)
+    {
+        $this->short_code = $shortCode;
+        $this->name = $name;
+        //FIXME; why do we have name and type with the same value?
```
Atala, line 32: it is legacy, i don't think it is used in the rider app anymore?

i spotted one usage here coopcycle-app/src/navigation/store/NewDeliveryForm.js in the source

src/Serializer/MyTaskMetadataDtoNormalizer.php
```diff
@@ -0,0 +1,60 @@
...
+use AppBundle\Api\Dto\MyTaskMetadataDto;
+use Symfony\Component\Serializer\Normalizer\ContextAwareNormalizerInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerAwareTrait;
+
+class MyTaskMetadataDtoNormalizer implements ContextAwareNormalizerInterface, NormalizerAwareInterface
+{
+    use NormalizerAwareTrait;
+
+    private const ALREADY_CALLED = 'MyTaskMetadataDtoNormalizer_ALREADY_CALLED';
```
Atala, line 14: what is the reason for these ALREADY_CALLED?

Atala (approved):

Atala:
all deliveries appeared as payed by credit card, but i guess it is an app-side problem, like a fallback that is not good

image

features/tasks.feature
vladimir-8: I did not find any usages of them in the Courier part, only the Dispatch

src/Serializer/MyTaskMetadataDtoNormalizer.php
vladimir-8: See the explanation here: https://symfonycasts.com/screencast/api-platform2-security/normalizer-aware#avoiding-recursion-with-a-context-flag

vladimir-8:
I checked in the database, and on my local instance all non-foodtech orders have payment method type `CARD`. Maybe we should introduce a payment method type `INVOICE` when we work on https://github.com/coopcycle/coopcycle/issues/41 ?

For now, it will be fixed by displaying only cash payments in the app: https://github.com/coopcycle/coopcycle-app/pull/1904

alexsegura (commented):
Can we remove the `My` prefix please?

Atala:
in this case what should be the name?

vladimir-8:
> Can we remove the My prefix please?

I used it to match `MyTasks` action https://github.com/coopcycle/coopcycle-web/blob/3644a3e14e4910c4120758398f2157b7ac7b5a67/src/Action/MyTasks.php#L13

I'll merge it as it is to start beta-testing, but let's discuss the name. I can rename all of them. 

How about `CourierTaskDto`? It's a Task representation for the courier role, and for the dispatch it could be `DispatchTaskDto`

PR #4690, opened 2024-10-08
full rework of `ExportTasksHandler`

r0xsh:
The old export was using CubeJS to perform the exports. I ported it to raw SQL.


r0xsh:
I guess it might be cool to also add a column for incidents, like `incident_count`, no ?

alexsegura (approved):

PR #4699, opened 2024-10-10
Allow reporting zero waste incidents

alexsegura:
This pull request introduces a way to add dynamic failure reasons (via classes implementing `FailureReasonsResolverInterface`). 

Also, it uses the `metadata` property of a `FailureReason` to describe additional form fields that should be attached to an incident. In this case, the form fields will allow describing how many zero waste containers have been returned by the customer. 

Finally, it adds a `metadata` property to the `Incident` class, to allow storing the form data.

### DIFF
--- a/app/DoctrineMigrations/Version20241009072602.php
+++ b/app/DoctrineMigrations/Version20241009072602.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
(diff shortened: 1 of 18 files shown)


r0xsh (commented):
Overall fine, i just don't get why you add metadata on Tasks entity and why you add the incident object in the Event :thinking:

r0xsh (approved):

r0xsh:
I'm not a huuge fan on the events subscriber pattern but well if it works it works :shrug:

PR #4700, opened 2024-10-10
ColisActiv export add option to export all

r0xsh:
image

### DIFF
--- a/src/Form/DataExportType.php
+++ b/src/Form/DataExportType.php
@@ -58,6 +58,7 @@ class DataExportType extends AbstractType
         }
         if ($this->colisactivEnabled) {
             $formats['form.task_export.colisactiv.label'] = 'colisactiv';
+            $formats['form.task_export.colisactiv_all.label'] = 'colisactiv_all';
         }
 
         if (count($formats) > 0) {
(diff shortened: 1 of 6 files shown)


src/Spreadsheet/ColisActivDataExporter.php
```diff
@@ -27,7 +27,11 @@ public function __construct(EntityManagerInterface $entityManager)
         $this->entityManager = $entityManager;
     }
 
-    public function export(\DateTime $start, \DateTime $end): string
+    public function export(
+        \DateTime $start,
+        \DateTime $end,
+        bool $includeAllTasks = false
```
alexsegura, line 33: This does not belong to the `DataExporterInterface` interface. 

Instead, you should move this to the constructor, and create another service with the `ColisActivDataExporter` class.

src/Form/DataExportType.php
```diff
@@ -87,7 +88,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
 
-                $content = $exporter->export($start, $end);
+                switch ($data['format']) {
+                    case 'colisactiv_all':
+                        $content = $exporter->export($start, $end, true);
+                        break;
+                    default:
+                        $content = $exporter->export($start, $end);
+                        break;
+                }
```
alexsegura, line 98: If you move `$includeAllTasks` to the constructor, you don't need to do this anymore, no specific use case 😉

alexsegura (approved):

PR #4710, opened 2024-10-22
Reduce the number of logs sent from timing functions

Atala:


app/config/config_prod.yml
```diff
@@ -83,7 +83,7 @@ monolog:
         timing:
             type: rotating_file
             path: "%env(APP_LOG_DIR)%/timing.%kernel.environment%.log"
-            level: debug
+            level: info
```
vladimir-8, line 86: This setting does not affect Datadog. It's only for local files. So we might keep it on `debug` however these logs might take a lot of disk space at some point 😁

Atala: yes it was i understood i saw only INFO logs in datado :+1:

Atala:
@vladimir-8 where are these changes ? in a branch or in master ?

vladimir-8:
They were in `master`, but I guess some log levels were reverted during rebase/merging

PR #4717, opened 2024-10-25
Frontend monitoring

vladimir-8:


r0xsh (approved):
LGTM, still concerned with the token in public but well, no choice :shrug:

PR #4718, opened 2024-10-28
Add a page to display all Recurrence rules

vladimir-8:
Screenshot 2024-10-31 at 16 04 00

includes:

- Recurrence rules ("recurring orders") created via the Delivery form
- Recurrence rules created via the Dispatch dashboard (possible options: Order+Delivery; Delivery without an Order/price; one/many Tasks without a Delivery)


src/Form/AddressBookType.php
```diff
@@ -133,24 +133,29 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
             if (null !== $address) {
 
-                $addresses = $options['with_addresses'];
-                if (!is_array($addresses)) {
-                    if (is_callable([ $addresses, 'toArray' ])) {
-                        $addresses = $addresses->toArray();
+                $addressBook = $options['with_addresses'];
```
Atala, line 136: should be `addressesBook`? (plural)

src/Service/DeliveryManager.php
```diff
@@ -146,4 +135,79 @@ public function setDefaults(Delivery $delivery)
...
+
+            if ($persist) {
+                $this->entityManager->persist($delivery);
+            }
+        }
+
+        return $delivery;
+    }
+
+    public function calculateRoute(TaskCollectionInterface $taskCollection): void
```
Atala, line 196: wouldnt that be calculated automatically by `TaskListSubscriber` ?

templates/store/recurrence_rules.html.twig
```diff
@@ -0,0 +1,24 @@
...
+
+{% block store_content %}
+
+  <h3>{{ 'stores.recurrence_rules'|trans }}</h3>
+
+  {% include '_partials/recurrence_rule/list.html.twig' with { items: recurrence_rules, empty_message: 'stores.recurrence_rules.empty_message' } %}
+
+{% endblock %}
+
+{#{% block styles %}#}
```
Atala, line 18: remove commented section

Atala:
I made it crash :
- go to the dispatch and create a delivery rule with just 1 dropoff
- go to recurrence rules list page for the store

image

Atala:
it does seem to me that the dropdown list to select a store when creating a recurrent rule in the dispatch does not display all the stores. is that expected? i can look deeper

vladimir-8:
> in french breadcrumb it is still abonnements ("subscription")

Are you sure it was the right branch? Looks good on my machine:

Screenshot 2024-11-07 at 11 47 56

vladimir-8:
> actually i thought the recurrence rules list page would be in the sidebar on "http://localhost/admin/deliveries" not on the store level. i guess admin users would like to see all the subscriptions for all his store. what do you think?

Yeah, I discussed it with Paul and we decided to put it on a store level for now. Maybe, at some point, we will make a global one as well

src/Form/AddressBookType.php
vladimir-8: No, it looks correct: https://en.wikipedia.org/wiki/Address_book

src/Service/DeliveryManager.php
vladimir-8: `TaskListSubscriber` is triggered only when an Entity is persisted. On the recurrence rules page, I only create  placeholders for Orders/Deliveries/Tasks, but don't persist them

vladimir-8:
> it does seem to me that the dropdown list to select a store when creating a recurrent rule in the dispatch does not display all the stores. is that expected? i can look deeper

For me, it displays everything. I'd say don't spend time on it, as we want to migrate away from that flow

vladimir-8:
> I made it crash :
> 
>     go to the dispatch and create a delivery rule with just 1 dropoff
>     go to recurrence rules list page for the store

It seems you were not running the latest version locally. The code in your stack trace is different from the last commit in this branch.

Now it handles both one pickup or one dropoff:

Screenshot 2024-11-07 at 13 10 40

Atala (approved):
thanks for your answer, i may have misstest the thing

just a short comment, in the translation string "recurrence_rules.table.generate_orders" the text should be "Generate tasks/orders" or even "Generate orders (as are getting rid of tasks) (or something like this)

i updated a french string for more clarity

PR #4722, opened 2024-10-30
Add an endpoint to asynchronously import deliveries CSV

Atala:


src/Spreadsheet/DeliverySpreadsheetParser.php
```diff
@@ -33,6 +38,7 @@ public function __construct(Geocoder $geocoder, PhoneNumberUtil $phoneNumberUtil
         $this->entityManager = $entityManager;
         $this->slugify = $slugify;
         $this->translator = $translator;
+        $this->defaultCoordinates = new GeoCoordinates(0,0);
```
alexsegura, line 41: Maybe we should use the instance default coordinates instead? In this case, the `$defaultCoordinates` should not be initialized in the constructor, to avoid unnecessary database calls when the class is initialized by dependency injection.

src/Spreadsheet/DeliverySpreadsheetParser.php
```diff
@@ -168,10 +207,22 @@ private function applyTags(TaggableInterface $task, $tagsAsString)
...
         }
     }
 
+    private function applyMetadata(Task $task, string $prefix, array $record)
+    {
+        $metadataColumn = $this->getColumn($prefix, 'metadata');
+
+        if (isset($record[$metadataColumn]) && !empty($record[$metadataColumn])) {
+            /* @throws JsonException */
+            $metadata = json_decode($record[$metadataColumn], JSON_THROW_ON_ERROR);
```
alexsegura, line 220: I wouldn't use JSON format to send metadata, but another "simpler" format like for packages (`foo=bar baz=bat`), or even URL encoded (`foo=bar&baz=bat`).

src/Spreadsheet/DeliverySpreadsheetParser.php
Atala: it makes more sense to me as well, so the dispatcher can see there is something wrong on the map, i will ask Antoine about it

src/Spreadsheet/DeliverySpreadsheetParser.php
```diff
@@ -2,71 +2,79 @@
...
+            if ($pickupAddress && $pickupAddress->getGeo()->isEqualTo($this->defaultCoordinates)) {
+                $delivery->getPickup()->addTags(Tag::ADDRESS_NEED_REVIEW_TAG);
+                //TODO: Trigger a incident.
+            }
 
             $delivery->getPickup()->setAddress($pickupAddress);
 
+            if (!$record['dropoff.address']) {
+                $dropoffAddress = $this->handleFaultyAddress($parseResult, $rowNumber, 'dropoff.address', 'dropoff', $options);
+            } else if (!$dropoffAddress = $this->geocoder->geocode($record['dropoff.address'])) {
```
r0xsh, line 69: :warning: geocode() can throw exception

src/Transporter/ImportFromPoint.php
```diff
@@ -63,7 +64,7 @@ public function import(
         }
 
         if ($address->getGeo()->isEqualTo($this->defaultCoordinates)) {
-            $task->setTags('review-needed');
+            $task->setTags(Tag::ADDRESS_NEED_REVIEW_TAG);
```
r0xsh, line 67: This tag don't exists by default, should be added via a migration

src/Spreadsheet/DeliverySpreadsheetParser.php
r0xsh: Done the changes

PR #4727, opened 2024-11-01
customise page titles

vladimir-8:
Screenshot 2024-10-29 at 13 38 41


vladimir-8:
> add the following comment

Done

> also i would add the title for the "Foodtech orders dispatch page".

Done

> is it interesting to have the date in the title like "Dispatch 2024-11-14"? maybe it is bad practice

Let's start with basic titles for now. If you see that somebody might benefit from it, we can add it later

PR #4729, opened 2024-11-05
Calculate CO2 consumption on TaskList update

Atala:


tests/AppBundle/Service/RouteOptimizerTest.php
```diff
@@ -42,15 +42,15 @@ public function setUp(): void
 
         $this->settingsManager->get('latlng')->willReturn('48.856613,2.352222');
 
-        $this->client = self::$container->get('vroom.client');
+        $this->client = self::getContainer()->get('vroom.client');
```
alexsegura, line 45: Are you sure this is the same?

app/DoctrineMigrations/Version20241031130556.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE task ADD co2_emissions INT NOT NULL DEFAULT 0');
```
alexsegura, line 23: Sorry to ask this question because the whole thing is built around this concept but... do we really need to store this in database, can't we calculate this on the fly?

Atala: I d rather save it in the DB so we have the data right there to export with the SQL query + be able to check things if we spot mistakes in the calculus

do you have an idea on how it will look if we calculate things on the fly?

PR #4731, opened 2024-11-07
Recurrent rules are sometimes (randomly) created double orders

vladimir-8:
Moved `useRecurrenceRulesGenerateOrdersMutation` from `UnassignedTasks` to `RightPanel` component to avoid it being called multiple times on `react-split` component resizing


vladimir-8:
@r0xsh I'll merge it; please let me know if you notice any issues

r0xsh:
I haven't tested the code, I've only reviewed it. While I don't see any particular issues with the implementation, I'm not convinced that giving the frontend the responsibility to create subscription rule tasks is the right approach.

I understand you want this page to be fast, which is why you're creating the tasks asynchronously. However, this can be achieved using [async features offered by Symfony](https://symfony.com/doc/5.4/the-fast-track/en/18-async.html).

The bug in this PR occurred because duplicate tasks were created due to multiple React re-renders. To me, this is a signal that we might not be using the right approach.

Nevertheless, if this implementation is working for now, let's proceed with what you've built. However, if we encounter similar bugs with this feature in the future, it might be worth considering a refactor.

r0xsh:
@vladimir-8 
I haven't tested the code, I've only reviewed it. While I don't see any particular issues with the implementation, I'm not convinced that giving the frontend the responsibility to create subscription rule tasks is the right approach.

I understand you want this page to be fast, which is why you're creating the tasks asynchronously. However, this can be achieved using [async features offered by Symfony](https://symfony.com/doc/5.4/the-fast-track/en/18-async.html).

The bug in this PR occurred because duplicate tasks were created due to multiple React re-renders. To me, this is a signal that we might not be using the right approach.

Nevertheless, if this implementation is working for now, let's proceed with what you've built. However, if we encounter similar bugs with this feature in the future, it might be worth considering a refactor.

vladimir-8:
Doing it asynchronously is a good option, but I guess it will make it harder to show a progress indicator to a user. For a moment, a user might think there is a bug and that the orders have not been created.

Issue #4725, opened 2024-10-31
When using "automatically assign to this rider" option in a store, the pickup and the dropoff may be assigned in the wrong order


Atala:
![Image](https://github.com/user-attachments/assets/4b1b3c09-3f16-442f-a6c2-cf273af0bc68)

PR #4732, opened 2024-11-08
Feat/barcode v2

r0xsh:
## This PR Includes

- **New Methods**: Added `assignTo(User)` and `unassign()` methods to the `Delivery` entity.
- **Updated Behavior for `/api/tasks/:id/{un}assign` Endpoints**: 
  - :warning: The `{un}assign` endpoints now apply to the entire delivery if the task is linked to one. Initially, I considered adding separate methods on the `Delivery` entity endpoint, but decided against it to maintain data integrity directly via a single endpoint.
- **Barcodes Added to the Following Screens**:
  - Dispatch Task Modal
  - Dispatch & Local Commerce Deliveries Lists
  - Dispatch & Local Commerce Delivery Details Screen
- **Improvements to `/api/barcode` Endpoint**:
  - To enhance UI responsiveness, I adopted a more RPC-oriented approach in designing this endpoint.

## To Be Completed

- [x] Add authentication to `BarcodeController` methods (consider using a [Voter](https://symfony.com/doc/5.x/security/voters.html))
- [ ] UI/UX adjustments
- [x] [Documentation](https://antoinebagnaud.notion.site/Barcode-5142eec676d84f75941aab68a585c97c?pvs=4) updates for users
- [x] Testing


features/tasks.feature
```diff
@@ -1360,6 +1367,7 @@ Feature: Tasks
             "hasIncidents": false,
             "incidents": [],
             "packages": [],
```
Atala, line 1369: barcodes should also be returned on packages right? or we will add it later

features/tasks.feature
```diff
@@ -1360,6 +1367,7 @@ Feature: Tasks
...
+            "barcode": "@array@",
```
Atala, line 1370: barcodes should also be returned on packages right? or we will add it later

src/Controller/BarcodeController.php
```diff
@@ -76,25 +79,45 @@ public function barcodeAction(
      * Possible actions: ask_to_assign, ask_to_unassign
      * ask_to_assign: Will prompt the user to self-assign
```
Atala, line 80: shall we have `ask_to_start` in the workflow? or maybe assign the task + start the task?

Atala (approved):
hello it seems good to me

can you check this bug and if it happens when you assign a delivery with a barcode? (drop is before pickup in the list) because I use the same functions as you
https://github.com/coopcycle/coopcycle-web/issues/4725

also there may be a UI/UX problem from:
- assign delivery1 with barcode
- assign delivery2 with barcode

then I think the rider will have the following assignments
1. delivery1.pickup
2. delivery1.dropoff
3. delivery2.pickup
4. delivery2.dropoff

but you may want to have all the pickups assigned first then the drops

features/tasks.feature
r0xsh: It already the case, the packages barcodes are returned. But you need to enable the group `barcode` when serializing the `Task` entity. I did it this way to avoid rendering all the barcodes when you call `GET /api/tasks` since this endpoint is critical and already a bit slow

src/Controller/BarcodeController.php
r0xsh: Yeah good idea ! But i'm expecting changes here. I'm pretty sure the auto-assign on scan can fire back on multi points delivery

PR #4745, opened 2024-11-15
Improve logging for Transporter sync

r0xsh:
title.


vladimir-8 (approved):
Looks good to me! The only thing I'd change is add `transporterName` to all log entries to have more information where are they coming from.

PR #4753, opened 2024-11-18
Update Mercado Pago usage

alexsegura:
I updated the Mercado Pago SDK (3.x) and successfully made a Mastercard payment in Colombia. 

I'd like to merge this and deploy it in Colombia to try it for real.


src/Form/Checkout/CheckoutPaymentType.php
```diff
@@ -42,6 +42,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ])
                 ->add('installments', HiddenType::class, [
                     'mapped' => false,
+                ])
+                ->add('issuer', HiddenType::class, [
```
Atala, line 46: this is specific to mercadopago right? maybe add a comment here `specific to mercadopago`

src/Form/Checkout/CheckoutPaymentType.php
```diff
@@ -42,6 +42,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ])
                 ->add('installments', HiddenType::class, [
```
Atala, line 43: maybe add a comment here explaining the `installments` concept for the payment form? it is also specific to mercadopago ?

alexsegura:
> i didnt get where you were checking for the country code to enable mercadopago? do you check for MX as well ?

@Atala it's done in GatewayResolver

https://github.com/coopcycle/coopcycle-web/blob/606748606e4dc8f10d62423927468da8356ab414/src/Payment/GatewayResolver.php#L23-L34

alexsegura:
> is this field needed in the mercadopago parameters in admin parameters?

@Atala from what I understand the client secret is needed when you want to "connect" (like Stripe) the shop. It's implemented, but I can't grab any client secret yet, so not able to test it.

alexsegura:
> this email field on payment form could be prefilled with the user email?

@Atala no, because it needs to be the email the customer is using for Mercado Pago (may be different)

alexsegura:
> hide edenred (and paygreen) settings on restaurant page?

@Atala they appear because you have enabled them with `EDENRED_ENABLED` env var 🤷‍♂️
Same for Paygreen.

src/Form/Checkout/CheckoutPaymentType.php
alexsegura: Installments is a general payment concept, it means "pay in 3 times" for example.

Atala:
shall we delete the "connect to mercadopago" on restaurant page then? what is the use of it?

alexsegura:
> shall we delete the "connect to mercadopago" on restaurant page then as it causes crashes and confusion? what is the use of it?

No, we keep it until I manage to grab a test client secret.

Atala:
what is the use of it when there is no client secret then?

at least hide or disable it when there is no client secret?

PR #4760, opened 2024-11-22
Fix tasks' after/before time display for recurrence rules

vladimir-8:


src/Form/DeliveryType.php
```diff
@@ -61,7 +61,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
             // When this is a new delivery,
             // set defaults for pickup/dropoff date
-            if (null === $delivery->getId()) {
+            if (null === $delivery->getId() && true === $options['asap_timing']) {
```
vladimir-8, line 64: That was the reason why the tasks' after/before time from the recurrence rule were not properly displayed

src/Form/DeliveryType.php
```diff
@@ -293,6 +299,7 @@ public function configureOptions(OptionsResolver $resolver)
             'with_weight' => true,
             'with_tags' => $this->authorizationChecker->isGranted('ROLE_ADMIN'),
             'with_dropoff_doorstep' => false,
+            'use_time_slots' => true,
```
vladimir-8, line 302: Also, there was a problem when a recurrence rule was set up using a pre-defined time slot. The time was saved correctly but was not properly displayed to a user

PR #4766, opened 2024-11-26
Add security token that allow a rider to use assign/unassign

r0xsh:
The action_token enables temporary elevated permissions for couriers.
It allows specific actions (like self-assignment) on the scanned task only.

Generation:
- Uses BarcodeUtils::getToken/1 to get the label's token
- Hashes it with xxh3 algorithm for additional security
- Uses runtime secret key to prevent token guessing

Security:
- Scoped to single task (token invalid for other tasks)
- Cannot be forged without access to runtime secret

### DIFF
--- a/src/Controller/BarcodeController.php
+++ b/src/Controller/BarcodeController.php
@@ -69,6 +69,20 @@ class BarcodeController extends AbstractController
         return $this->json([
             "ressource" => $iri,
             "client_action" => $clientAction,
+
+            /**
+             * The action_token enables temporary elevated permissions for couriers.
+             * It allows specific actions (like self-assignment) on the scanned task only.
+             *
+             * Generation:
+             * - Uses BarcodeUtils::getToken/1 to get the label's token
+             * - Hashes it with xxh3 algorithm for additional security
+             * - Uses runtime secret key to prevent token guessing
+             *
+             * Security:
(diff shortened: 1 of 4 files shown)


src/Controller/BarcodeController.php
```diff
@@ -65,9 +65,11 @@ public function barcodeAction(
         $this->taskManager->scan($task);
         $this->doctrine->getManager()->flush();
 
+        $iri = $iriConverter->getIriFromItem($task);
         return $this->json([
-            "ressource" => $iriConverter->getIriFromItem($task),
+            "ressource" => $iri,
             "client_action" => $clientAction,
+            "token_action" => hash('xxh3', BarcodeUtils::getToken($iri)),
```
Atala, line 72: maybe add a comment to explain what this token is about?

Atala (approved):

src/Controller/BarcodeController.php
r0xsh: https://github.com/coopcycle/coopcycle-web/pull/4766/commits/f31cfdd8339cda315ed11ae7a6926eb240361b11#diff-ce8ab65090e1c1907d56d08bbc3d1d313960dcc803973d848c2a9b9d125ca231R72-R85

Issue #4068, opened 2024-03-16
Esc button to deselect


Paul-Eraman-CoopCycle:
the only way to not have anything selected after having things selected is to draw a square on a part of the map with no points, which is not very UI/UX friendly.

[needescbutton.mp4.zip](https://github.com/coopcycle/coopcycle-web/files/14624955/needescbutton.mp4.zip)

request for escape button to deselect everything

Atala:
@Paul-Eraman-CoopCycle "esc" keycode is already binded to "close search"

want to handle both with "esc" ?

Paul-Eraman-CoopCycle:
does "bind" mean cant also be used somewhere else? because it would make sense to have it for both since you are never deselecting tasks on the map and searching at the same time

Atala:
it can be used for both

but i think it is useful for people with search :
- go on search mode to get your tasks
- search -> select task
- search -> select another task
ad lib
and then "close search"
and the drag n drop with the previously selected tasks

what do you think ?

Paul-Eraman-CoopCycle:
i think that makes the search function kind of clumsy, try to save things between search seems messy. 

the solution i see for that use case you are describing would be a selecting button (check box) that persists over time instead of just highlighting. or something like that, something more persisty than highlighting

Atala:
not sure if we 100% understand each other here, we should talk about it

Paul-Eraman-CoopCycle:
Taking out of Routes V.2 and backlogging

PR #4772, opened 2024-11-28
In the dispatch panel, add selection with SHIFT/MAJ key in unassigned tasks + ESC press shortcut to unselect all

Atala:


js/app/dashboard/components/RightPanel.js
```diff
@@ -80,8 +81,12 @@ const DashboardApp = ({ loadingAnim }) => {
     generateOrders(date)
   }, [date]);
 
+  const unselectAll = () => {
+    dispatch(selectTasksByIds([]))
+  }
+
   return (
-    <div className="dashboard__aside-container">
+    <div className="dashboard__aside-container" onKeyDown={ (e) => e.keyCode === 27 && unselectAll() }>
```
vladimir-8, line 89: `=== 27` create a const instead for a magic number

js/app/i18n/locales/en.json
```diff
@@ -93,6 +93,7 @@
         "ADMIN_DASHBOARD_FILTERS_SHOW_ALL": "Show all",
         "ADMIN_DASHBOARD_FILTERS_HIDE": "Hide",
         "ADMIN_DASHBOARD_HIDE_SHOW_ON_MAP": "Hide/show on map",
+        "ADMIN_DASHBOARD_ONLY_UNASSIGNED_WITH_SHIFT": "Only unassigned tasks column handle shift selection",
```
vladimir-8, line 96: we can mention that a user can use `command` key on macos / `Ctrl`? on windows to select individual tasks. I tested it works for both unassigned and assigned tasks

vladimir-8 (approved):
Looks good to me. I guess you fixed https://github.com/coopcycle/coopcycle-web/issues/4068 too

js/app/i18n/locales/en.json
vladimir-8: Yes! With screenshots. Shift, for example, did not even know that the right panel was resizable

PR #4777, opened 2024-11-29
Fix new delivery rangepicker

r0xsh:


js/app/widgets/DateTimeRangePicker.js
```diff
@@ -0,0 +1,191 @@
...
+    const updatedSecondOptions = secondSelectOptions.map(option => {
+      const isBefore = after.isBefore(option.time)
+      return {
+        ...option,
+        disabled: !isBefore,
+      }
+    })
+
+    setSecondSelectOptions(updatedSecondOptions)
+  }, [timeValues.after])
```
r0xsh, line 76: Okay, this definitely works, BUT (it's a bit nitpicking) why not perform this computation in `handleChange()`?
I see two different patterns/control flows to update the state here. This is okay in a small component like this, but it could lead to hidden behaviors in larger codebases.

PR #4780, opened 2024-12-04
Rework timeslot picker delivery

CamilleNerriere:
Switch to a datepicker and select


r0xsh:
New expected behavior:
- Allow admin to override the timeslots
- Store owners keep the regular timeslots picker

# TODO
- [ ] Fix the timeslots picker on the store delivery page when the admin has overriden the timeslots. (Only show text)
- [ ] Also update/integrate the new datepicker for the store delivery page

Atala:
TimeSlotPicker :
- [ ] if only one date is available, dont show date picker but the date
- [ ] when loading having the first date + first time as default
- [ ] same when changing the selected timeslot
- [ ] same when changing the date
- [ ]  localize date correctly in the datepicker it should be something like "5 décembre 2024", not 2024/12/05
- [ ] add a little bit of margin between the two select
- [ ] crash on time selection on store page after having changed the selected timeslot with radio buttons
- [ ] format more nicely the picking range for an existing delivery for a store (using little date at the moment)
- [ ]

PR #4784, opened 2024-12-06
Previous validation prompt

r0xsh:
https://github.com/user-attachments/assets/166fb542-b065-4d7b-98e1-cc7de3490b50

### DIFF
--- a/js/app/dashboard/redux/actions.js
+++ b/js/app/dashboard/redux/actions.js
@@ -115,6 +115,9 @@ export const INSERT_IN_UNASSIGNED_TOURS = 'INSERT_IN_UNASSIGNED_TOURS'
 
 export const SET_TOURS_ENABLED = 'SET_TOURS_ENABLED'
 
+export const TASK_CONFIRMATION_REQUIRED = 'TASK_CONFIRMATION_REQUIRED';
+export const TASK_CONFIRMATION_RESOLVED = 'TASK_CONFIRMATION_RESOLVED';
+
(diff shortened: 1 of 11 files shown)


js/app/dashboard/redux/actions.js
```diff
@@ -680,7 +683,135 @@ export function startTask(task) {
   }
 }
 
-export function completeTask(task, notes = '', success = true) {
+/**
+ * Creates a task queue processor with the given context
+ * @returns {Function} - Configured task queue processor
+ */
+function createTaskProcessor(context) {
```
vladimir-8, line 690: To be honest, this looks a bit like "spaghetti code," especially a callback `onRequireConfirmation` that returns a promise that passes `onResolve` inside a redux action... 
Are you doing all this because `TaskModalContent` is a class component?

Maybe we can: 

- move a call to `completeTask` inside `TaskCompleteForm`
- re-implement `completeTask` using an RTK query mutation hook: https://github.com/coopcycle/coopcycle-web/blob/51c13a0c04eb90d19b5aa9be3c243bdcbaa73673/js/app/api/slice.js#L4 https://redux-toolkit.js.org/rtk-query/usage/mutations#mutation-hook-behavior
- show `TaskConfirmationDialog` conditionally based on `completeTask` response
- mark pending pickups as done directly from `TaskConfirmationDialog` (by dispatching an action that will trigger the same queue-like behaviour)

r0xsh:
Yeah @vladimir-8, you're right on all counts - I'm not really satisfied with how the code turned out either. I'll tackle this feature again when I have more time to do it properly.

PR #4789, opened 2024-12-11
Fix delivery price calculation when two packages have the same name

Atala:


src/Entity/PackageSet.php
```diff
@@ -40,6 +41,13 @@ class PackageSet
 
     protected $id;
     protected $name;
+
+    /**
+     * @Assert\Unique(
+     *  normalizer="AppBundle\Entity\Package::getPackageName",
```
vladimir-8, line 47: Do you do it because there is no way to call `getName` directly?

features/retail_prices.feature
vladimir-8, line 1: 👍

src/Entity/Task.php
```diff
@@ -1401,4 +1401,13 @@ public function setDistanceFromPrevious($distanceFromPrevious)
...
     }
+
+    public function getStore(): Store|null
+    {
+        if(!is_null($this->getDelivery())) {
+            return $this->getDelivery()->getStore();
+        }
+
+        return null;
+    }
```
vladimir-8, line 1412: is not it the same as `$this->getDelivery()?->getStore()`

vladimir-8 (approved):

src/Entity/Task.php
Atala: ah yes indeed i didnt know this PHP syntax

src/Entity/PackageSet.php
Atala: I dont know exactly. "normalizer" arg has to be an anonymous function so I had difficulties to find the correct format so I ended up with a static function.

src/Entity/Task.php
Atala: I will change

PR #4804, opened 2024-12-16
Feature/invoicing

vladimir-8:


src/Entity/Sylius/Order.php
```diff
@@ -1961,20 +2022,20 @@ public function getDeliveryPrice(): PriceInterface
...
             throw new \LogicException('Order has no delivery price');
         }
 
         $productVariant = $deliveryItem->getVariant();
 
-        if (str_starts_with($productVariant->getCode(), 'CPCCL-ODDLVR')) {
-            // price based on the PricingRuleSet
-            return new PricingRulesBasedPrice($deliveryItem->getUnitPrice());
-        } else {
+        if (str_starts_with($productVariant->getCode(), 'RBTRR-PRC-')) {
```
r0xsh, line 2033: Isn't this change can cause issue for the already existing products variant in the database who start with `CPCCL-ODDLVR` ?

r0xsh: And tbh, i don't get why you need to change this in the first place :thinking:

src/Entity/Sylius/Order.php
```diff
@@ -1961,20 +2022,20 @@ public function getDeliveryPrice(): PriceInterface
...
             // custom price
             return new ArbitraryPrice($productVariant->getName(), $deliveryItem->getUnitPrice());
+        } else {
+            // price based on the PricingRuleSet
+            return new PricingRulesBasedPrice($deliveryItem->getUnitPrice());
```
r0xsh, line 2038: Huh and you swapped the if branches, i'm confused lol

r0xsh (commented):
All good to me !
I investigate a bit more about the change related to `CPCCL-ODDLVR` i think i get it now; but i prefer to double check with you before merging this PR

src/Entity/Sylius/Order.php
vladimir-8: `CPCCL-ODDLVR` is a `Product` code, while I'm adding a prefix to a `Product Variant` code that is uuid currently.

vladimir-8: I use to later to figure out if a product variant has an arbitrary price or the price was calculated by the pricing rules

vladimir-8: Maybe it should instead be a separate column in ProductVariant table to make it more explicit

vladimir-8: I now understood what you meant initially 😄 While, it should not be an issue to change it, I think it was a good solution previously, I'll revert it and try to put a flag about the price in a separate column

docker-compose.yml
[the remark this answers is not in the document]
vladimir-8: Done in https://github.com/coopcycle/coopcycle-web/pull/4804/commits/e10d1af8d0b53039fe689476db4465e29791da70

PR #4805, opened 2024-12-17
Calculate the distance for a task into a tasklist even when no vehicle is set

Atala:


src/Entity/Task.php
vladimir-8, line 1: Where do you plan to use distance and CO2 emissions?

I'm not sure that putting them into a Task entity is a good idea. It seems to me that `distanceFromPrevious`, at least, belongs more to a `TaskCollectionItem`, as it depends on the position of a Task in a list and should be managed together.

src/Doctrine/EventSubscriber/TaskSubscriber/TaskListProvider.php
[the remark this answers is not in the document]
Atala: agree

src/Entity/Task.php
Atala: I was wondering about that as well.

At the time I put it on Task directly, as we are for real interested only on the last value:  a task may be assigned several days in a row, but actually only the last assignment if several make sense (when we accomplish the task for real). so in reality there is only one value we are interested into.

src/Doctrine/EventSubscriber/TaskSubscriber/TaskListProvider.php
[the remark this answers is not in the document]
Atala: It has been like this for a long time, I ll do it in another PR

src/Entity/Task.php
Atala: (also https://github.com/coopcycle/coopcycle/issues/166 )

Atala:
Issue here https://github.com/coopcycle/coopcycle/issues/167

src/Entity/Task.php
vladimir-8: > we are for real interested only on the last value

Then it seems a waste of time to recalculate all these values all the time. Maybe subscribe to a task:done event and calculate them only then? (I'd also rename the fields in that case to something in the past like `traveled_distance`, `co2_consumed`)

vladimir-8: Yes, I think it’s better if they are set after task:done event, makes the logic behind these fields a bit more clear

app/DoctrineMigrations/Version20250114155816.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE task RENAME COLUMN co2_emissions TO emitted_co2');
+        $this->addSql('ALTER TABLE task RENAME COLUMN distance_from_previous TO traveled_distance_meter');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+
```
vladimir-8, line 30: Not super critical here, but it's good to keep `down` migration if we need to revert

src/Doctrine/EventSubscriber/TaskListSubscriber.php
```diff
@@ -49,11 +48,12 @@ public function getSubscribedEvents()
...
     }
 
-    private function calculate(TaskList $taskList)
+    private function calculate(TaskList $taskList, EntityManagerInterface $em)
     {
         $coordinates = [];
         $tasks = [];
-        $vehicle = $taskList->getVehicle(); 
+        $vehicle = $taskList->getVehicle();
+        $uow = $em->getUnitOfWork();
```
vladimir-8, line 56: it seems you are not using neither `$uow`, nor `EntityManagerInterface $em`

src/Domain/Task/Handler/MarkAsDoneHandler.php
```diff
@@ -53,5 +57,58 @@ public function __invoke(MarkAsDone $command)
...
+        foreach ($taskList->getTasks() as $item) {
+            $coordinates[] = $item->getAddress()->getGeo();
+        }
+
+        // going back to the warehouse
+        if (!is_null($vehicle)) {
+            $coordinates[] = $taskList->getVehicle()->getWarehouse()->getAddress()->getGeo();
+        }
+
+        $route = $this->routing->route(...$coordinates)['routes'][0];
```
vladimir-8, line 89: We can live with it, but it's still a bit strange that we recalculate a route for each single task while we already calculated a route for the TaskList. I guess the issue is that we do not store route legs in the TaskList table?

vladimir-8 (approved):

src/Domain/Task/Handler/MarkAsDoneHandler.php
Atala: yes we just have the total distance stored in TaskList, we could maybe serialize the whole travel as JSON?

Atala: I will add a comment for now

src/Doctrine/EventSubscriber/TaskListSubscriber.php
Atala: cleaned

app/DoctrineMigrations/Version20250114155816.php
Atala: added

PR #4826, opened 2025-01-15
Feat/new edit form

CamilleNerriere:
Adapt the new delivery form in order to use it in edit mode


Atala (changes requested):
Hi,

here is the problem i found :
- if you have a date range that is on several days, then it doesnt show the complex date picker has it should
- it does not show correctly phone number that was previously set on delivery
(- it does not display manually set price -> to be seen later)
- it does 3 calls to calculate when you load the page
image

small UI improvments :
- datepicker field should be before package field on drops
- the initial loader should be bigger and in the middle of the page

PR #4858, opened 2025-01-23
Invoicing: mark exported orders

vladimir-8:
- Added: save a file id in which each order appears
- Display status next to each order if it was previously exported or not
- Added a filter to only show orders which were not exported before

Screenshot 2025-01-27 at 16 09 01


alexsegura (approved):

Atala:
@vladimir-8 I think you should display human readable name instead of `requestId` to the user (as you did for the filename)

PR #4874, opened 2025-01-29
Implement properly parcels & labels in Woopit delivery request.

### LINKED ISSUE #4863: Add package barcodes in Woop integration
Woop can send packages and we need to provides them barcode for this packages

These packages may or may not be already created on our side

We need to store their packages ID so we dont create them several times. The problem is that we don't get a package ID from them.

alexsegura:
Fixes #4863

### DIFF
--- a/features/woopit.feature
+++ b/features/woopit.feature
@@ -277,18 +277,35 @@ Feature: Woopit
       """
       {
         "deliveryId":"yXJPEP30xgja",
+        "parcels": [
+          {
+            "id": @string@
+          },
+          {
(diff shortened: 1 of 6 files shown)


Atala (approved):
Looks good to me, I would suggest the following changes :
 - have a more "hard" testing that check that we have same `parcel.id` and `labels.parcelId` in the response and that the packageId generated is always the same
 - maybe lock the Woopit package in the admin interface (but I guess it should be harcoded, so it is ugly...)

Issue #4875, opened 2025-01-30
Crash when refusing an order from the admin interface


Atala:
**Describe the bug**

https://coopcycle.slack.com/archives/G01JR0YF5V3/p1737717167991969

**To Reproduce**
1. Go to the admin page of an order that is new
2. Click on Refuse button
3. The page crash with "order can not be cancelled" message

alexsegura:
I can't reproduce.

PR #4876, opened 2025-01-30
Fix crash when refusing order from the Order page

Atala:
#4875

Refused order status is also valid when all tasks of a delivery are cancelled


r0xsh:
Should allow transition from refused to cancelled into `state_machine.yml`

alexsegura (commented):
You could also just allow the transition from `refused` to `cancelled` in `app/config/state_machine.yml`

```yml
cancel:
  from: [new, accepted, started, ready, refused]
  to: cancelled
```

PR #4889, opened 2025-02-06
Fix : wrong value in VolumeUnits steps when creating a new rule

Atala:


js/app/components/PriceRangeEditor.js
```diff
@@ -116,12 +118,16 @@ export default ({ defaultValue, onChange }) => {
...
-            onChange({
-              attribute: unitToAttribute(e.target.value),
-              price,
-              step: multiplyIfNeeded(stepEl.current.value, e.target.value),
-              threshold: multiplyIfNeeded(thresholdEl.current.value, e.target.value),
-            })
+
+            const newUnit = e.target.value
+
+            if (newUnit === 'vu') {
```
vladimir-8, line 124: There is a bug if the switch is between units that are both multiplied (kg, km); it needs an extra check that the `current unit` is 1000x (kg, km) and the `newUnit` is not (vu), and vice versa:

Screenshot 2025-02-10 at 14 35 17

vladimir-8 (changes requested):

js/app/components/PriceRangeEditor.js
vladimir-8: May be add a helper to check if a unit is multiplied or not instead of hard-coding 'vu' to allow adding more units

Atala: I think it should be refactored at a "higher level" to remove all these multiplication/division from the component, and just multiply by *1000 in the change handler if it kg/km
let's do that later when we have more time, knowing that the price screen is soon to be refactored

vladimir-8 (approved):

PR #4904, opened 2025-02-18
Support both google places and google places (new)

Atala:


js/app/components/AddressAutosuggest/google.js
```diff
@@ -50,35 +55,95 @@ const autocompleteOptions = {
...
+          }
+        })
+      }
       this._autocompleteCallback(predictionsAsSuggestions, value)
 
-    }
+    }).catch(error => {
+      // our API key is not valid for the places API (new), fallback to legacy API
+      if (error.response.status === 403) {
+        console.error("[Google adapter] Using legacy Places API")
```
vladimir-8, line 111: Maybe log to datadog instead so we can see errors and warn cooperatives?

See https://github.com/coopcycle/coopcycle-web/blob/master/js/app/hooks/useDatadog.js or `window.DatadogLogger` can be accessed without a hook if needed

vladimir-8 (approved):

PR #4916, opened 2025-03-03
Add "tour.name" field in CSV file delivery import to automatically add a delivery to a tour on import

Atala:


src/Spreadsheet/DeliverySpreadsheetParser.php
```diff
@@ -241,6 +248,24 @@ private function applyTags(TaggableInterface $task, $tagsAsString)
         }
     }
 
+    private function setTour(Delivery $delivery, string $routeName) {
```
vladimir-8, line 251: > string $routeName

`tourName`?

tests/AppBundle/Resources/spreadsheet/deliveries_with_tour.csv
```diff
@@ -0,0 +1,3 @@
+pickup.address,pickup.address.name,pickup.address.description,pickup.address.telephone,pickup.comments,pickup.timeslot,dropoff.address,dropoff.address.name,dropoff.address.description,dropoff.address.telephone,dropoff.comments,dropoff.timeslot,dropoff.packages,weight,pickup.metadata,dropoff.metadata,tourName
```
vladimir-8, line 1: maybe `tour.name` to keep it consistent?

tests/AppBundle/Spreadsheet/DeliverySpreadsheetParserTest.php
```diff
@@ -56,18 +62,17 @@ protected function createParser(): AbstractSpreadsheetParser
 
         $this->packageRepository = $this->prophesize(ObjectRepository::class);
 
-        $this->entityManager
-            ->getRepository(Package::class)
-            ->willReturn($this->packageRepository->reveal());
+        $this->tourRepository = self::getContainer()->get(TourRepository::class);
```
vladimir-8, line 65: maybe a test fails, because it should be a mock?

src/Spreadsheet/DeliverySpreadsheetParser.php
```diff
@@ -241,6 +248,24 @@ private function applyTags(TaggableInterface $task, $tagsAsString)
...
+        foreach ($delivery->getTasks() as $task) {
+
+            $date = $task->getAfter();
+            $tour = $this->tourRepository->findByNameAndDate($routeName, $date);
+
+            if (is_null($tour)) {
+                $tour = new Tour();
+                $tour->setName($routeName);
+                $tour->setDate($date);
+                $this->entityManager->persist($tour);
```
vladimir-8, line 261: I'm not sure if it's a good place to do that. In my understanding, the responsibility of this class is parsing only, and persistence should be handled elsewhere (ImportDeliveriesHandler)

Atala: yes that the main problem I have with this code, it looks like an unwanted side-effect...

I had difficulties to "find the tour" without persisting it, i ll try to do better

Atala: see https://github.com/coopcycle/coopcycle-web/pull/4916/commits/c8d54d5053bb096e8d46333361788173b68f6b4d

tests/AppBundle/Spreadsheet/DeliverySpreadsheetParserTest.php
Atala: mmm I dont think so
I ve had weird doctrine issues when working on such parts of the code
I think it is mostly due that there are so much listeners crossed that sometimes doctrine lose track of what is where

src/Spreadsheet/DeliverySpreadsheetParser.php
vladimir-8: that looks better; the only thing: I would not 'pollute' the Delivery entity with `tourName` as that value is only needed during parsing. May a parser could return an object/array like this instead?

```
if (!$parseResult->rowHasErrors($rowNumber)) {
                $parseResult->addData($rowNumber, [
                    'delivery' => $delivery,
                    'tourName' => $record['tour.name'] ?? null
                ]);
            }
```

Atala: I am not 100% convinced by this approach.

Maybe the good solution would have to be a DTO for "Delivery" that could be used for every input type or several DTO as DeliveryInput and childs CSVDeliveryInput, APIDeliveryInput ?

PR #4920, opened 2025-03-05
Feat : in beta delivery form synchronize dropoff dates when pickup date changes (timeslot and non-timeslot cases)

Atala:
Expected behavior :
 - when pickup date changes, if drop(s) date is after change it to pickup date
 - when pickup timeslot change, change drop timeslot
 - when pickup timeslot choice change, if drop(s) date is after change it to pickup date

### DIFF
--- a/assets/react/controllers/Delivery/DeliveryForm.jsx
+++ b/assets/react/controllers/Delivery/DeliveryForm.jsx
@@ -413,7 +413,7 @@ export default function({ storeId, deliveryId, order }) {
               }
 
               // Case 2: Time slot changed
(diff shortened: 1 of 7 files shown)


r0xsh (changes requested):
i got hurt in my own confusion, the code is barely readable and i noticed few bugs.

- hourOptions.map() crash
- Weird behavior selecting hours that change other values is a unexpected way
- Some missing time-slots that are visible is the legacy form

I feel like this need a whole refactoring :/

r0xsh (approved):
Still a bit hard to follow whats going on, but you know. The business logic is complex, so as the code.
But from my testing i cannot see anymore bugs, well done :)

PR #4922, opened 2025-03-10
feat: add refresh behavior on foodtech tracking screen

r0xsh:


vladimir-8 (changes requested):
Tested on a mobile phone. I found a few issues:

1. Guest checkout is not supported
2. Poll continues to fetch an order state even when the page is in the background

Take a look at https://redux-toolkit.js.org/rtk-query/usage/polling It seems RTK query has a polling mechanism built-in and an option to disable it when a page is not focused.

To support guest checkout check how it is done on the payment page:

A token is passed to `guestSlice`:

https://github.com/coopcycle/coopcycle-web/blob/master/js/app/order/payment/index.js#L31

And then `apiSlice` has a configuration with a list of endpoints enabled for guest checkout: 

https://github.com/coopcycle/coopcycle-web/blob/master/js/app/api/baseQuery.js#L8

Tip: If you have issues with accessing an instance from a mobile browser see https://github.com/coopcycle/coopcycle-web/wiki/Local-development#testing-frontend-on-mobile-devices

Atala:
i gave a try to RTK query here 23522be6e1720eac00e485851f0f5c37a28e663b

PR #4925, opened 2025-03-11
In tests, fallback on geocode earth in backend when addok provider fails

Atala:


src/Service/Geocoder.php
```diff
@@ -77,6 +81,18 @@ private function getGeocoder()
         return $this->geocoder;
     }
 
+    private function createGeocodeEarthProvider() {
+        $rateLimiter = RateLimiterMiddleware::perSecond($this->rateLimitPerSecond, $this->rateLimiterStore);
```
alexsegura, line 85: The `$limit` should be 10 requests per second for Geocode Earth

https://geocode.earth/#pricing

Atala: I think there was an other problem is that I used the same rate limit store as the open cage one. see here : ab8dbda70a0d0ef1ae5edf398b53255c9d7d38eb

alexsegura (commented):
Why are tests still failing when this is supposed to fix the problem? 🤔

Issue #4510, opened 2024-07-12
Return debugging info in price calculation endpoint


alexsegura:
Currently, to debug pricing rules calculation, we send API requests to `/pricing_rules/{id}/evaluate`. 

Instead, the debugging information should be returned in the response of the `/retail_prices/calculate` endpoint. 

For example:

```json
{
    "@type": "RetailPrice",
    "amount": 1250,
    "currency": "EUR",
    "tax": {
        "amount": 0,
        "included": true
    },
    "calculation": {
        "rules": [
            {
                "expression": "weight > 1000",
                "matched": false
            },
            {
                "expression": "distance > 5000",
                "matched": true,
                "price": 500
            }
        ]
    }
}
```

PR #4940, opened 2025-03-19
Add a pricing rule type that works as a percentage

vladimir-8:
- Added: price choice to update current sub-total by percentage:

Screenshot 2025-04-02 at 17 38 08

Docs: https://docs.coopcycle.org/en/admin/pricing_rule/#price

- Updated: price debug tool in both old and new Delivery Forms:

Debug information is returned in two formats:

1. Rules (`calculation` field), as described in https://github.com/coopcycle/coopcycle-web/issues/4510
2. Order items (`items` field); based on a simplified Sylius OrderItem/ProductVariant/ProductOptions structure as the first step towards https://github.com/coopcycle/coopcycle/issues/261

Screenshot 2025-04-02 at 17 36 07

Docs: https://docs.coopcycle.org/en/admin/pricing_testing/


src/Action/Delivery/CalculateRetailPrice.php
```diff
@@ -68,10 +116,28 @@ public function __invoke(Delivery $data, Request $request)
...
             $this->currencyContext->getCurrencyCode(),
             $taxAmount,
             'included' === $request->query->get('tax', 'included')
         );
     }
 }
+
+class CalculationItem {
+    /**
+     * @param RuleResult[] $rules
```
Atala, line 131: doesnt work directly as function annotation?

Atala (commented):
hello, I am surprised by the following : percentage is not applied if the pricing strategy is the "first rule that match". which kind of make sense but I think should not be the case (percentage rules should be always applied or be applied in a separate "rules run")

- debug mode should be made accessible to admin users thanks to a link or button so it is used

src/Action/Delivery/CalculateRetailPrice.php
vladimir-8: As far as I know, php does not support typed arrays. Or am I missing something?

vladimir-8:
> hello, I am surprised by the following : percentage is not applied if the pricing strategy is the "first rule that match". which kind of make sense but I think should not be the case (percentage rules should be always applied or be applied in a separate "rules run")

I have not considered this case before, but actually, they could be applied even with the "first rule that match" strategy. For example, if there are some rules per task, and percentage is the only matching rule per order. I removed the warning from UI.

Screenshot 2025-04-16 at 11 29 06

assets/react/controllers/Delivery/DeliveryForm.jsx
```diff
@@ -359,11 +357,13 @@ export default function({ storeId, deliveryId, order, isDispatcher }) {
         const { response, error } = await httpClient.post(url, infos)
 
         if (error) {
+          setCalculateResponseData(error.response.data)
           setPriceErrorMessage(error.response.data['hydra:description'])
           setCalculatePrice(0)
         }
 
         if (response) {
+          setCalculateResponseData(response)
```
Atala, line 366: is it needed to have two states here ? (calculateResponseData and calculatePrice)

Atala:
I was thinking about this use case. In my understanding percentages come "on top of existing price" so it should be applied to the previously calculated price (even if the setting is "the first rule that matches")  

image

assets/react/controllers/Delivery/DeliveryForm.jsx
vladimir-8: Yes, the difference is that `calculateResponseData` contains not only success but also an error response, which is needed to display debug info when the price is not calculated

vladimir-8:
> I was thinking about this use case. In my understanding percentages come "on top of existing price" so it should be applied to the previously calculated price (even if the setting is "the first rule that matches"). here below an example with order based rules.
> 
> it could be used like this :
> 
>     have rules to calculate pricings that are applied in any way
>     have a percentage IF urgent then add +20% to price
> 
> hence the two separate "runs" and percentage applied in a 2nd part
> 
>     do the price calculus as usual (for orders, for stops, ..)
>     apply percentage rules on top of the calculated price
> 
> but i don't know what was the intended use case for percentage rule in the first place?

That's one of the intended use cases. 

I don't treat percentage as a special kind of rule. I'm treating them as a standard rule where `percentage` is just a different way for a rule to calculate a price. So, in your example, it does not make sense to use a percentage-based rule with the "The first rule that matches" setting, because there will never be a price to apply a percentage to. That's why initially I had a comment suggesting to use percentages only with "All the matching rules" setting. However, if you mix rules per point and rules per order, there is a way to make percentage rule work even with the "The first rule that matches" setting https://github.com/coopcycle/coopcycle-web/pull/4940#issuecomment-2810402359

assets/react/controllers/Delivery/DeliveryForm.jsx
Atala: I mean you can refactor a little bit `setPriceErrorMessage` and the corresponding component so we have only one state in the parent DeliveryForm component ?
but is is a detail no worry

Atala (approved):
OK

I've seen that you added that it works only with "all the rule that match" in the docs as well. Maybe a small help text in the UI below the "percentage" select could help?

Anyway the PR looks good to me. I cant test it today so you can go ahead and merge if you want or wait for me to test again on Monday but I think it s good

PR #4948, opened 2025-03-26
In the store setting, when there is no default timeslot selected force select the first timeslot as default

Atala:
Avoid a crash in the delivery form


r0xsh (approved):

r0xsh:
Thanks for the fix :)

PR #4950, opened 2025-03-26
Update to symfony flex

Atala:
Problems for now

- [x] when installing symfony/validator on my machine, the symfony/validator/Resources folder is not created (wtf I see it in the repo), thus we don't have the translations
- [x] the sylius services are not loaded in the services
- [x] "bearer_token" config key used for security is not recognized as valid ("invalid configuration key")
- [x]  craue_config_cache_provider is not correctly defined
- [x] port back the https://github.com/coopcycle/coopcycle-web/commit/144f9df55f160f964b2ada4d3d6fdba9b1cf76b0 commit to new bin/console and index.php (if necessary, as it seems to be loaded by symfony by default now)
- [x] fix "scripts" in composer.json, it crashes "symfony-cmd" not found
```
    "scripts": {
        "auto-scripts": {
            "cache:clear": "symfony-cmd",
            "assets:install %PUBLIC_DIR%": "symfony-cmd"
        },
        "post-install-cmd": [
            "@auto-scripts"
        ],
        "post-update-cmd": [
            "@auto-scripts"
        ]
    },
```

- [x] remove Behat/demo.php, bootstrap.php
- [x] port the code in [`app/AppKernel.php`](https://github.com/coopcycle/coopcycle-web/blob/3d2e4a03b0a1df096fb3cab1a48988a08d6f2ca3/app/AppKernel.php) to `src/Kernel.php` (needed for production)
- [ ] reintegrate drupal hardener


alexsegura:
`symfony/validator` Resources folder is removed by `drupal/core-vendor-hardening` Composer plugin

https://github.com/drupal/core-vendor-hardening/blob/11.x/Config.php#L78

alexsegura:
Not sure why/how, but Behat is still using the `app/autoload.php` file. If we remove it, Behat is broken (env vars are not loaded)

PR #4951, opened 2025-03-26
Use PHP attributes for symfony and API platform

Atala:
Still need to migrate manually the custom validators we added (not a big deal)


alexsegura:
Also I noticed we are still using `Vich\UploaderBundle\Mapping\Annotation`

PR #4952, opened 2025-03-27
Migrate to RTK 2.0 and Redux 5.0

vladimir-8:
Enables `skipPollingIfUnfocused` behaviour added in https://github.com/coopcycle/coopcycle-web/pull/4922


package.json
```diff
@@ -153,19 +153,17 @@
...
-    "react-redux": "^7.2.9",
+    "react-redux": "^8.1.0",
     "react-select": "^5.8.0",
     "react-split": "^2.0.14",
     "react-spreadsheet": "^0.9.2",
     "react-stickynode": "^4.1.0",
     "react-timeago": "^7.2.0",
     "react-toastify": "^7.0.4",
     "reduce-reducers": "^1.0.4",
-    "redux": "^4.2.0",
```
alexsegura, line 164: So now `redux` & `redux-thunk` are required via `react-redux`?

vladimir-8: As I understand, we don’t need to have a direct dependency on redux and redux-thunk. Redux toolkit will bring a supported version of both.

Ideally, we should not use them directly either.

https://coopcycle.slack.com/archives/G01JR0YF5V3/p1743187275000519?thread_ts=1743115574.590749&cid=G01JR0YF5V3

Issue #306, opened 2018-04-12
When finishing a task on the dashboard, the icon marker on the map is not updated


Atala:

couleurmate:
Actually the whole iframe isn't synced in real time

Issue #305, opened 2018-04-12
When creating a foodtech delivery (on order ready) delivery is not pushed to the dashboard


Atala:

alexsegura:
More generally speaking the dashboard is not updated when new tasks are created.

alexsegura:
I can't reproduce. 

Tasks are correctly pushed to the dashboard when the order is accepted.

PR #4960, opened 2025-04-14
[PERTE-306] Send tasks' related messages to users with ROLE_DISPATCHER

JuanIgAcosta:
### Issue: [#306](https://github.com/coopcycle/coopcycle/issues/306)

Related issues:
- [coopcycle/coopcycle#235](https://github.com/coopcycle/coopcycle/issues/235)
- [coopcycle/coopcycle#305](https://github.com/coopcycle/coopcycle/issues/305)
- [coopcycle/coopcycle#313](https://github.com/coopcycle/coopcycle/issues/313)

---
We need to send missing websocket messages to `ROLE_DISPATCHER` users so they can be in-sync from mobile app UI with web's dispatch UI.
The system currently does that for `ROLE_ADMIN` users only.

**Tasks:**
- [x] Create lower-level function that receives an event, a list of roles and an _optional_ array of "additional" user to be included as recipient.
Make sure no duplicated users are selected as a message receivers.
- [x] Add a few testcases for `src/Domain/Task/Reactor/PublishLiveUpdate.php` .
- [x] Add a few testcases for `src/Service/LiveUpdates.php`.
- [x] Make sure that the message `task:rescheduled` is also sent (and possibly others, check and add accordingly).
We'll add support for those in the mobile app in the near future.
- [x] Add verification at [Courier/taskEntityReducer.js](https://github.com/coopcycle/coopcycle-app/blob/ae60074209aee2d85f3f982db3e7f94e2eca7372/src/redux/Courier/taskEntityReducer.js#L342) to process courier messages only `(if task_list.user === currentuser)` (this will be done as part of [#305](https://github.com/coopcycle/coopcycle/issues/305) at PR [#1975](https://github.com/coopcycle/coopcycle-app/pull/1975))
  * Done at commit: https://github.com/coopcycle/coopcycle-app/pull/1975/commits/b216c681512e8b810c64f2af092bfdfecef00b49


src/Domain/Task/Reactor/PublishLiveUpdate.php
```diff
@@ -18,19 +29,35 @@ public function __construct(LiveUpdates $liveUpdates)
...
         // legacy event and new version of event
         // see https://github.com/coopcycle/coopcycle-app/issues/1803
         if ($event instanceof TaskListUpdated) {
             $user = $event->getCourier();
-            $this->liveUpdates->toUsers([ $user ], $event);
-        } else if ($event instanceof TaskListUpdatedv2) { // can be safely broadcasted both to riders and admins
-            $this->liveUpdates->toAdmins($event);
-            $user = $event->getTaskList()->getCourier(); // not used in the rider part of the app yet
-            $this->liveUpdates->toUsers([ $user ], $event); // not used in the rider part of the app yet
+            $this->liveUpdates->toUsers([$user], $event);
```
Atala, line 40: If I am not mistaken you are overly complicating things here

the main problem we are solving here is that there is no filtering on tasklist:updated event handler in the rider part so if an user is a rider + an admin/dispatcher the event will mess with his assignment ([this line:](https://github.com/coopcycle/coopcycle-app/blob/ae60074209aee2d85f3f982db3e7f94e2eca7372/src/redux/Courier/taskEntityReducer.js#L342) it needs to be fixed by adding a check that the tasklist to update is the one of the rider)

so then you just need to be careful with this event for now; so the code could be

>         if ($event instanceof TaskListUpdated) {
>             $user = $event->getCourier();
>             $this->liveUpdates->toUsers([ $user ], $event);
>         } else if ($event instanceof TaskListUpdatedv2) { // can be safely broadcasted both to riders and admins
>             $this->liveUpdates->toRoles(['ROLE_ADMIN', 'ROLE_DISPATCHER'], $event);
>             $user = $event->getTaskList()->getCourier(); // not used in the rider part of the app yet
>             $this->liveUpdates->toUsers([ $user ], $event); // not used in the rider part of the app yet
>         } else {
>             $this->liveUpdates->toRoles(['ROLE_ADMIN', 'ROLE_DISPATCHER'], $event);
>         }

Atala (commented):
I dont see why you need such a complicated check, just broadcast events to all admins and dispatchers for now it is good

the only problem is for task_list:updated because we use it in the rider app of the app so if the rider is both rider+admin/dispatcher because there is no filtering it will display any updated tasklist to this rider (huge mess)

corresponding line : https://github.com/coopcycle/coopcycle-app/blob/ae60074209aee2d85f3f982db3e7f94e2eca7372/src/redux/Courier/taskEntityReducer.js#L342

we should actually add the check in the app code ( `if task_list.user === currentuser` ) so we dont break it by mistake. can you do it or do I do it?

src/Domain/Task/Reactor/PublishLiveUpdate.php
diegomanuel: > We are going to do that change (if task_list.user === currentuser) at [Courier/taskEntityReducer.js](https://github.com/coopcycle/coopcycle-app/blob/ae60074209aee2d85f3f982db3e7f94e2eca7372/src/redux/Courier/taskEntityReducer.js#L342) as part of https://github.com/coopcycle/coopcycle/issues/305 at PR https://github.com/coopcycle/coopcycle-app/pull/1975.

Done at commit: https://github.com/coopcycle/coopcycle-app/pull/1975/commits/b216c681512e8b810c64f2af092bfdfecef00b49

Atala: hello,

I am not very fan of filtering, I would send all for the stake of simplicity but maybe it can spare resource not to send everything. Please send the tour events in the same if... else... please.

Also please move the list of accepted events on top to increase readability, like 
```
$broadcastedEvents = [...]

if $event in $brodacstedEvents:
```

diegomanuel: Finally did the changes you said @Atala at this commit:
https://github.com/coopcycle/coopcycle-web/pull/4960/commits/3480bb8bd5558e83e4ec26457f6f7eb7043eb862

alexsegura (approved):

PR #4961, opened 2025-04-15
Fix: "products" column (products outside a section) is empty in menu sections form

Atala:
The bug : the "products" column in the menu editor form is empty

Why? I am not PHP wizard but when we added the annotation with rector, the value of "products" in the twig template is now NULL (here https://github.com/coopcycle/coopcycle-web/commit/534bc0a395bbc31f60ae857446f6cc828f6756c5#diff-cb098e0335fc596eab9af092879d3298620efb6a6ae84441078dd5976738ca8eR12 )

### DIFF
--- a/src/Utils/MenuEditor.php
+++ b/src/Utils/MenuEditor.php
@@ -11,7 +11,7 @@ class MenuEditor
     /**
      * @var \Doctrine\Common\Collections\Collection
      */
-    public $products;
+    private $products;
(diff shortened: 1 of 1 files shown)


alexsegura:
> when we added the annotation with rector, the value of "products" in the twig template is now NULL

It's because in Twig, public properties have priority over getters. Before, it was `getProducts()` that was used.

PR #4963, opened 2025-04-22
Feat: improve cart loading time

Atala:
Uses 5 query less to load the cart from session

### DIFF
--- a/src/Entity/Sylius/OrderRepository.php
+++ b/src/Entity/Sylius/OrderRepository.php
@@ -32,7 +32,7 @@ class OrderRepository extends BaseOrderRepository
      * @param $indexBy
      * @return QueryBuilder
      */
-    public function createOpmizedQueryBuilder($alias, $indexBy = null)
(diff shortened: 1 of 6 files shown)


src/Entity/Sylius/OrderRepository.php
```diff
@@ -55,6 +55,17 @@ public function createOpmizedQueryBuilder($alias, $indexBy = null)
         return $qb;
     }
 
+    public function findCartById($id) : ?Order {
+        $qb = $this->createOpmizedQueryBuilder('o');
```
alexsegura, line 59: Can we fix this typo?

alexsegura:
Does it trigger less db queries?

Atala:
5 queries less

src/Entity/Sylius/OrderRepository.php
Atala: yes sure

PR #4970, opened 2025-04-24
Little tech debt

diegomanuel:
These are some changes and additions we did to fix some small things and make our lives easier.


cypress/support/commands.js
```diff
@@ -30,21 +30,20 @@ Cypress.Commands.add('terminal', command => {
   if (prefix) {
     cmd = `${prefix} ${cmd}`
   }
-  cy.exec(cmd)
+  cy.exec(cmd, {timeout: 60000})
 })
 
 Cypress.Commands.add('symfonyConsole', command => {
-  const prefix = Cypress.env('COMMAND_PREFIX')
```
r0xsh, line 37: Why did you remove the PREFIX behavior ?

r0xsh (approved):
Nice work! 👍
Thanks a lot for helping clear out the technical debt! 😄
I might be a bit nitpicky, but I'm not a huge fan of the "testdataX" naming convention in the Makefile; the names aren't very explicit.

Also, I don't think this is used anywhere, so it's not a big deal. But just out of curiosity: why did you remove the command prefix mechanism for Symfony commands?

cypress.config.js
```diff
@@ -1,42 +1,43 @@
...
 
   nodeVersion: "system",
   viewportWidth: 1600,
 
+  defaultCommandTimeout: 10000,
+
   retries: {
     // Configure retry attempts for `cypress run`
-    runMode: 2,
+    runMode: 4,
```
vladimir-8, line 23: Does it really help? I notice that if it consistently fails 2-3 times, then the issue is probably somewhere else 😄 (more retries just make a test run longer)

cypress/e2e/local-commerce/@admin/recurrence_rules/create_delivery_with_recurrence_rule_as_admin_store_without_timeslots.cy.js
vladimir-8, line 1: Did these strange timeouts (10001,10002,10003) help in the end? If not, maybe we can remove them?

Makefile
```diff
@@ -16,29 +23,96 @@ osrm:
...
+	@while clear && cypress run --browser chrome --headless --no-runner-ui --spec cypress/e2e/local-commerce/@admin/update_price.cy.js; do :; done
+cypress-open:
+	@cypress open
+cypress-install:
+	@docker compose exec -e APP_ENV=test -e SYMFONY_ENV=test -e NODE_ENV=test webpack npm install -g cypress@14.3.2 @cypress/webpack-preprocessor@6.0.4
+	@npm install -g cypress@14.3.2 @cypress/webpack-preprocessor@6.0.4
 
 jest:
-	@docker compose exec -e SYMFONY_ENV=test -e NODE_ENV=test webpack npm run jest
+	@docker compose exec -e APP_ENV=test -e SYMFONY_ENV=test -e NODE_ENV=test webpack npm run jest
```
alexsegura, line 54: `SYMFONY_ENV` can be removed

Makefile
```diff
@@ -16,29 +23,96 @@ osrm:
...
+test: phpunit jest behat cypress
+
+testdata1:
+	@docker compose exec php bin/console coopcycle:fixtures:load -f cypress/fixtures/high_volume_instance.yml --env test
+testdata2:
+	@docker compose exec php bin/console coopcycle:fixtures:load -f cypress/fixtures/dispatch.yml --env test
+testdata3:
+	@docker compose exec php bin/console coopcycle:fixtures:load -f cypress/fixtures/package_delivery_orders.yml --env test
+testdata4:
+	@docker compose exec php bin/console coopcycle:fixtures:load -f cypress/fixtures/restaurant.yml --env test
```
alexsegura, line 105: Can those have more explicit names?

Makefile
```diff
@@ -16,29 +23,96 @@ osrm:
 phpunit:
 	@docker compose exec php php bin/console doctrine:schema:update --env=test --force --no-interaction --quiet
 	@docker compose exec php php vendor/bin/phpunit
+# Add as annotation in any testcase:
+# /**
+# * @group only
+# */
+# public function testSomething()..
+phpunit-only:
+	@clear && docker compose exec php php vendor/bin/phpunit --group only
```
alexsegura, line 32: Not sure this belongs here... or maybe rename it `phpunit-debug` and use `@group debug` annotation?
There are other ways to execute only some tests, like 

```
vendor/bin/phpunit <filename>
vendor/bin/phpunit <filename> --filter=<testSomeMethodName>
```

Makefile
```diff
@@ -16,29 +23,96 @@ osrm:
...
 
 behat:
 	@docker compose exec php php vendor/bin/behat
+# For now, just change here the `features/file.feature:xx` to run a specific test
+behat-only:
+	@clear && docker compose exec php php vendor/bin/behat features/stores.feature:96
```
alexsegura, line 38: It's also possible to add a `@debug` tag on top of a scenario, and then run `vendor/bin/behat --tags="@debug"`

Makefile
alexsegura: Also, wondering what is the usage of this? Each Cypress scenario will load the needed fixtures, no?

cypress/support/commands.js
diegomanuel: We are now calling `cy.terminal('bin/console ${command} --env="test"')` there!
And the `Cypress.env('COMMAND_PREFIX')` thing is already done inside `terminal` command:
```
Cypress.Commands.add('terminal', command => {
  const prefix = Cypress.env('COMMAND_PREFIX')
  cy.exec(prefix ? `${prefix} ${command}` : command, {timeout: 60000})
})
```

cypress.config.js
diegomanuel: Yepp u r right, made no difference at all. Reverting!

diegomanuel: Yepp u r right, made no difference at all. Reverting!

cypress/e2e/local-commerce/@admin/recurrence_rules/create_delivery_with_recurrence_rule_as_admin_store_without_timeslots.cy.js
diegomanuel: They didn'! We used'em only to quickly point out the timeouting one, but made no difference anyways, will remove.
NOTE: The default timeout was changed to `10000` so everything now should wait for 10s max.

Makefile
diegomanuel: Cool, done!

Makefile
diegomanuel: Done! Changed them to be more verbose:
```
testdata-dispatch:
testdata-restaurant:
testdata-package-delivery-orders:
testdata-high-volume-instance:
```
The usage is mainly because we often use the `test` local env (like at `http://localhost:9080/` from `web` and `192.168.0.xx:9080` for `app`) because the fixtures in there are handy to test our changes in the mobile app (for example, the @vladimir-8's `high_volume_instance.yml` one)
If we use our `dev` local env, then `make demodata` does the trick for that env.

Makefile
diegomanuel: Niceeee, done, thanks!

cypress.config.js
diegomanuel: Well, in the end, I tried a ridicously high retry number (9) and seemed to make a difference.

This one passed at the last retry:
image

This one in the 5th run:
image


:loudspeaker:  So I will let the `runMode: 9` and see how it behaves with some time being enabled.
We can always change back to whatever value at any moment directly at `master`.

docker-compose.yml
```diff
@@ -246,6 +246,7 @@ services:
       - ./cube.js:/cube/conf/cube.js
     logging:
       driver: none
+    attach: false
```
alexsegura, line 249: I never found how to avoid all those useless logs, nice 👍

cypress/support/commands.js
```diff
@@ -26,25 +26,40 @@
...
-  let cmd = `bin/console ${command} --env="test"`
-  if (prefix) {
-    cmd = `${prefix} ${cmd}`
-  }
-  cy.exec(cmd)
+  cy.terminal(`bin/console ${command} --env="test"`)
+})
+
+Cypress.Commands.add('loadFixtures', (fixture, setup=false) => {
```
alexsegura, line 36: The problem is that sometimes we want to load _multiple_ fixtures. This function should accept a variadic number of arguments.

cypress/e2e/local-commerce/@admin/beta_create_delivery_order.cy.js
alexsegura, line 1: Please refrain from changing indentation of the whole file, it makes it impossible to read the actual diff.

cypress/e2e/local-commerce/@admin/beta_create_delivery_store_with_invalid_price.cy.js
alexsegura, line 1: Please refrain from changing indentation of the whole file, it makes it impossible to read the actual diff.

cypress/support/commands.js
diegomanuel: Did the change! It supports now a list of fixtures.
Applied over the added tests that were doing the manual call to `symfonyConsole`.

PR #4983, opened 2025-05-07
API; Delivery entity; Use DeliveryCreateOrUpdateProcessor in both POST and PUT requests

vladimir-8:


src/Api/State/DeliveryCreateOrUpdateProcessor.php
```diff
@@ -0,0 +1,96 @@
...
+        // @see ApiPlatform\Symfony\EventListener\ValidateListener
+        $errors = $this->validator->validate($delivery);
+        if (count($errors) > 0) {
+            throw new ValidationException($errors);
+        }
+
+        /** @var ArbitraryPrice $arbitraryPrice */
+        $arbitraryPrice = null;
+        if ($data instanceof DeliveryInput) {
+            $arbitraryPrice = $data->arbitraryPrice;
```
r0xsh, line 49: How does it work now? since you removed arbitrary price on the entity ?

src/Api/State/DeliveryCreateOrUpdateProcessor.php
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        $useArbitraryPrice = $this->authorizationCheckerInterface->isGranted('ROLE_DISPATCHER') && $arbitraryPrice;
```
r0xsh, line 52:
```suggestion
        $useArbitraryPrice = $this->authorizationCheckerInterface->isGranted('ROLE_DISPATCHER') && !is_null($arbitraryPrice);
```

r0xsh: It's better to be type safe, PHP type casting can be weird sometimes. Also this code wont work it the user try to set an arbitrary price of `0` since 0 = false

r0xsh (approved):

src/Api/State/DeliveryCreateOrUpdateProcessor.php
vladimir-8: It reads a value directly from `DeliveryInput` (DTO)

src/Api/State/DeliveryCreateOrUpdateProcessor.php
vladimir-8: Done

PR #5016, opened 2025-06-03
New deliveries import command + proximus support

r0xsh:
### Usage examples:
Performing an import via the `proximus` importer and the `proximus` transformer
- `bin/console coopcycle:deliveries:import /api/stores/11 -i proximus -t proximus`

Performing an import **without** importer & transformer
- `bin/console coopcycle:deliveries:import /api/stores/420 https://foo.bar/deliveries.csv`

Performing an import **without** importer but **with** a transformer (it can be a non CSV file, since the transformer will translate it)
- `bin/console coopcycle:deliveries:import /api/stores/1337 -t foobar https://weirdo.file/deliveries.txt`

### Help page:
```
/var/www/html # bin/console coopcycle:deliveries:import --help
Description:
  Imports deliveries

Usage:
  coopcycle:deliveries:import [options] [--] <store> [<csv>]

Arguments:
  store                          Store name
  csv                            CSV content

Options:
  -t, --transformer=TRANSFORMER  transformer name
  -i, --importer=IMPORTER        importer name
  -d, --dry-run                  Dry run mode
```

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -116,6 +116,7 @@ CUBEJS_DUCKDB_S3_SECRET_ACCESS_KEY=coopcycle
 CUBEJS_DUCKDB_S3_PATH=/images/exports/%type%
 INSEE_PORTAIL_API_KEY=
 STANDTRACK_API_KEY=
+PROXIMUS_SYNC_URI=
 
 ###> acseo/typesense-bundle ###
 TYPESENSE_URL=http://localhost:8108


config/services.yaml
```diff
@@ -1634,3 +1634,11 @@ services:
     arguments:
       - '@AppBundle\LoopEat\Client'
       - '@monolog.logger.loopeat'
+
+  AppBundle\Transporter\Proximus\ProximusImporter:
+    lazy: true
+    arguments:
+      - "%env(PROXIMUS_SYNC_URI)%"
```
diegomanuel, line 1641: We may want to add `PROXIMUS_SYNC_URI` to `.env.dist`?
With an empty value, but just to have it defined there so we all know that it exists!

PR #5018, opened 2025-06-05
Port the edit recurring orders page to the new form

vladimir-8:
Port the edit recurring orders page to the new form:

Screenshot 2025-06-09 at 12 00 40


r0xsh (approved):
As we where talking about yesterday ; i'm okay keeping the whole "mode if" behavior

PR #5034, opened 2025-06-19
Implement multiple pickups

### LINKED ISSUE #5033: Enable multiple pickups
if we have multiple pickups then we need to find a way to allocate the dropoff packages between those pickups
which is complicated. should also only be allowed for specific people who we trust and are superusers

alexsegura:
Fixes #5033

**TODO**
- [x] Make `TaskMapper::getWeight` compatible with multi pickups
- [x] ~Change `TasksProvider::postProcessResult` implementation to support multi pickups~ superseded by #5035

### DIFF
--- a/js/app/components/delivery-form/DeliveryForm.js
+++ b/js/app/components/delivery-form/DeliveryForm.js
@@ -75,6 +75,17 @@ function getFormattedValue(value) {
   return value
 }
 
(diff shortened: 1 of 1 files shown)


r0xsh (commented):
I don't see any code changes or tests in the pricing calculator department; will it work as it is ?

PR #5069, opened 2025-08-14
Add an option to re-calculate price

vladimir-8:
### added: display order details on existing orders

Screenshot 2025-08-21 at 16 48 00

### added: re-calculate price for existing orders - users now can recalculate delivery price when modifying orders and choose to keep an old price or switch to the new price

Screenshot 2025-08-21 at 16 48 37

## Under-the-hood improvements 

- Delivery /POST / PUT: return Delivery entity in response (instead of a DeliveryInputDto, use DeliveryInputDto only as input type)
- Add Order field into a Delivery response
- rename: order_minimal -> foodtech_order_minimal
- define redux store RootState type (Delivery form)

## Mics

- API tests: assert entire 'order' object in responses
- e2e tests: wait for 'delivery-itinerary' component to load before proceeding - avoiding flack e2e tests
- copy sort-import config from -app repo


src/Api/Dto/DeliveryMapper.php
```diff
@@ -102,12 +104,53 @@ public function map(
...
+
+            if (null === $variant) {
+                continue;
+            }
+
+            foreach ($variant->getOptionValues() as $productOptionValue) {
+                /** @var ProductOptionValue $productOptionValue */
+
+                // Find the PricingRule linked to this ProductOptionValue
+                $pricingRule = $this->findPricingRuleByProductOptionValue($productOptionValue);
```
r0xsh, line 132: I'm not very familiar with the DTO Mapping pattern, but this seems incorrect to me. 🤔
Making a database call in the DTO layer feels off; I believe it should belong to the repository or persistence layer. I realize I might be a bit nitpicky about this, and it's not a major issue (whether I'm right or wrong to raise it, haha).

js/app/components/delivery-form/hooks/useSubmit.ts
```diff
@@ -278,8 +282,10 @@ export default function useSubmit(
       setIsSubmitted(true);
 
       if (modeIn(mode, [Mode.DELIVERY_CREATE, Mode.DELIVERY_UPDATE])) {
-        const deliveryId = data.id;
-        const orderId = data.order?.id;
+        const delivery = data as Delivery;
+        const deliveryId = delivery.id;
+        const orderUri = delivery.order?.['@id'];
+        const orderId = orderUri?.split('/').pop();
```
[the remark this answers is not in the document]
vladimir-8: Yeah, alternatively, we could add an `id` to an order object, but it does not look much better to me

src/Api/Dto/DeliveryMapper.php
vladimir-8: I guess, ideally, there should be a bi-directional relationship between a pricing rule and a product option value (currently, there is only a one-way relationship from a pricing rule to a product option value) and then let Doctrine manage it. I'll add it as a comment

PR #5084, opened 2025-09-11
Range-based manual supplements

vladimir-8:
⚠️ Please review, but don't merge yet (needs testing with QA coops)

---

Added: range-based manual supplements

Screenshot 2025-09-25 at 17 47 59

that can be configured in the pricing rules as:

Screenshot 2025-09-25 at 17 49 15

Improved: display quantities in the order items and 'price per unit':

Screenshot 2025-09-25 at 17 50 54

### DIFF
--- a/app/DoctrineMigrations/Version20251002081202.php
+++ b/app/DoctrineMigrations/Version20251002081202.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20251002081202 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $hash = [];
+
+        $stmt = $this->connection->prepare('SELECT id, restaurant_id FROM mercadopago_account');
+        $result = $stmt->execute();
+        while ($mpAccount = $result->fetchAssociative()) {
+            $hash[$mpAccount['restaurant_id']] = $mpAccount['id'];
+        }
+
+        $this->addSql('ALTER TABLE mercadopago_account DROP CONSTRAINT fk_b056d88b1e7706e');
+        $this->addSql('DROP INDEX uniq_b056d88b1e7706e');
+
+        $this->addSql('ALTER TABLE mercadopago_account DROP restaurant_id');
+        $this->addSql('ALTER TABLE restaurant ADD mercadopago_account_id INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE restaurant ADD CONSTRAINT FK_EB95123F295E6E66 FOREIGN KEY (mercadopago_account_id) REFERENCES mercadopago_account (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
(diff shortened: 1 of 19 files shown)


app/DoctrineMigrations/Version20250917000001.php
r0xsh, line 1: This migration seems a bit scary to me.
I might be confused by the convoluted data model of stylus; but it seems you're treating a one-to-many relationship as a one-to-one ?

Maybe to be extra sure we wont lose data we can add something like that at the top of the migration

```php
public function up(Schema $schema): void
{
$result = $this->connection->fetchOne('
        SELECT COUNT(*) 
        FROM pricing_rule pr1
        WHERE EXISTS (
            SELECT 1 
            FROM pricing_rule pr2 
            WHERE pr2.product_option_value_id = pr1.product_option_value_id 
            AND pr2.id != pr1.id
        )
');

if ($result > 0) { 💥 } 

// The current migration code here
```

js/app/components/delivery-form/components/ManualSupplement.tsx
```diff
@@ -2,47 +2,28 @@ import { useMemo } from 'react';
...
   const { values, setFieldValue } = useDeliveryFormFormikContext();
 
   const price = useMemo(() => {
     return rule.priceAst ? parsePriceAST(rule.priceAst, rule.price) : null;
   }, [rule.priceAst, rule.price]);
 
-  const isChecked = useMemo(() => {
-    return values.order.manualSupplements.some(
+  const isRangeBased =
+    price instanceof PriceRange && price.attribute === 'quantity';
```
r0xsh, line 23: oh ! we got `instenceof` nice

src/Api/State/DeliveryCreateOrUpdateProcessor.php
```diff
@@ -215,4 +220,70 @@ public function process(
 
         return $delivery;
     }
+
+    private function hasManualSupplementsChanged(ManualSupplements $manualSupplements, OrderInterface $existingOrder): bool
```
r0xsh, line 224: I think this function code check is order dependent.
We may want to `usort` both of them before checking them via that double loop iteration might prevent us from unexpected behavior in the future

src/Pricing/OnDemandDeliveryProductProcessor.php
r0xsh, line 1: This one hurts my smooth brain

app/DoctrineMigrations/Version20250917000001.php
vladimir-8: > it seems you're treating a one-to-many relationship as a one-to-one ?
Yes, while we modelled it as a one-to-many relationship in the database, in reality, it was always a one-to-one relationship (one pricing rule linked to one product option value). If we get the same product option value linked to multiple pricing rules, it means there was a bug somewhere. However, it should not be a big issue if we drop redundant values, as until now, product option values have only stored custom pricing rule names. In the worst-case scenario, we will only lose the recent name changes.

(Once this PR is merged, we will actually have a one-to-many relationship, but in reverse: one pricing rule can be linked to many product option values)

src/Api/State/DeliveryCreateOrUpdateProcessor.php
vladimir-8: Updated in https://github.com/coopcycle/coopcycle-web/pull/5084/commits/4122402a2e2ecdc06ff396850fd5a4ca93b6c3ce

src/Pricing/OnDemandDeliveryProductProcessor.php
vladimir-8: I hope it's better now 😄 https://github.com/coopcycle/coopcycle-web/pull/5084/commits/9ba556b2fadbeadbf1f7fa6e3e09e30357c1e4b1

PR #5102, opened 2025-10-09
Implement pawaPay

alexsegura:
This pull request also contains a migration to extend the length of `craue_config_setting.value` column, to be able to hold the pawaPay API key

pawapay


app/DoctrineMigrations/Version20251008141706.php
r0xsh: FYI, `CHAR`, `VARCHAR`, `TEXT`, and others char types are stored the same way and have no performance impact
https://www.postgresql.org/docs/current/datatype-character.html

But i think the goal here is getting ride of the length limitation

r0xsh (approved):
I'll add some logging and add order id to the payment description. But other than that it's all good

PR #5110, opened 2025-10-16
Feature: API to accept inaccurate order details incidents & UI to manage them

vladimir-8:
- Defined API request structure to report inaccurate order details: see https://github.com/coopcycle/coopcycle-web/pull/5110/files#diff-ff262451f34f10722e7a59369e6e0ca0eab529065fa67b3df81d353a69f4fb83
- UI to accept/decline suggestions

Screenshot 2025-10-30 at 12 37 58

- Localise hard-coded strings
- FIXED: user display on the incident page
- FIXED: remove packages that are no longer present in the input (order editing)
- Internal: introduce `AppRootWithDefaults` for antd-based pages

### DIFF
--- a/js/app/admin/incidents/[id]/components/IncidentImages.tsx
+++ b/js/app/admin/incidents/[id]/components/IncidentImages.tsx
@@ -6,9 +6,9 @@ import { useTranslation } from 'react-i18next';
 import store from '../redux/incidentStore';
 import { selectImages, selectIncident } from '../redux/incidentSlice';
 
-async function _handleUpload(id, file, t) {
+async function _handleUpload(id, file) {
   if (file.size > 5 * 1024 * 1024) {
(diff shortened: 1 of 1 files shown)


src/Action/Incident/CreateIncident.php
```diff
@@ -73,4 +88,76 @@ public function __invoke(Incident $data, ?UserInterface $user, Request $request)
         return $data;
     }
 
+    private function preFillMissingDataInSuggestion(Incident $incident): void
+    {
+        $metadata = $incident->getMetadata();
+
+        if (empty($metadata)) {
+            return;
+        }
```
r0xsh: I'm a bit confused, it seems the overall feature work/is built around suggestion in metadata. But where is it injected in the first place ?
I initially thought it was this `preFillMissingDataInSuggestion` function. But seems she skip it when metadata are empty 😕

r0xsh (approved):

PR #5128, opened 2025-11-05
Recalculate price on task cancellation

vladimir-8:
- [x] Recalculate order price on task cancellation
- [x] Introduce 'order:price_updated' event
- [x] Don't modify manually applied prices
- [x] Don't apply to foodtech and B2C delivery orders
- [x] Handle per distance recalculation: Exclude cancelled tasks from a route (distance, duration, polyline)
- [x] Handle per packages recalculation: Exclude cancelled tasks from the sum of packages and weight 
- [x] Don't assign cancelled tasks to a messenger 
- [x] Display 'order:price_updated' in the order history
Screenshot 2025-11-20 at 15 46 23

### DIFF
--- a/tests/Behat/FeatureContext.php
+++ b/tests/Behat/FeatureContext.php
@@ -150,6 +150,7 @@ class FeatureContext implements Context, SnippetAcceptingContext
     public function clearData()
     {
         $this->databasePurger->purge();
+        $this->redis->flushDB();
     }
 
     /**


js/app/components/order/id/OrderHistory.tsx
```diff
@@ -53,6 +54,23 @@ const OrderEventDetails = ({ event }: { event: OrderEvent }) => {
...
+            total={event.data.old_total}
+            taxTotal={event.data.old_tax_total}
+          />
+          <TotalPrice
+            overridePrice={false}
+            total={event.data.new_total}
+            taxTotal={event.data.new_tax_total}
+          />
+        </div>
+      ) : null}
```
r0xsh: Maybe the type check should be performed on the `TotalPrice` component, it'll make `OrderHistory.tsx` cleaner and `TotalPrice` more error resilient

src/MessageHandler/Task/Command/CancelHandler.php
```diff
@@ -13,20 +23,140 @@
...
+    private function processOrderIfNeeded(OrderInterface $order, Delivery $delivery): void
+    {
+        if ($order->isFoodtech()) {
+            // It's not possible to cancel tasks for foodtech orders only an entire order can be cancelled
+            // so the logic that follows does not apply
+            $this->logger->info('Skipping processing for foodtech order', ['order' => $order->getId()]);
+            return;
+        }
+
+        $this->deliveryManager->calculateRoute($delivery);
```
r0xsh: This is prone to throwing :/

- Minimalist approach: wrap this in a try/catch and notify the user that the price could not be recalculated.
- The "Yikes" way: I’ll move this into an async worker job for two reasons:
  1. The combined cancel-recalculation + route calculation can be quite long (especially the route calculation).
  2. If calculateRoute throws, I can easily retry the failed job(s).

vladimir-8: Done in https://github.com/coopcycle/coopcycle-web/pull/5128/commits/238cb77ecab204f4a0ff661f9fb43e0ca9c838b2

PR #5138, opened 2025-11-14
Task<->Order navigation

vladimir-8:
Added a way to navigate from an Order to Tasks:

Screenshot 2025-11-14 at 12 19 35

Added a way to navigate from a Task to an Order:

Screenshot 2025-11-14 at 12 20 12

### DIFF
--- a/js/app/dashboard/components/TaskModalContent.js
+++ b/js/app/dashboard/components/TaskModalContent.js
@@ -29,7 +29,7 @@ import {
 import { selectCurrentTask, selectCurrentTaskEvents } from '../redux/selectors'
 import { selectSelectedDate } from '../../coopcycle-frontend-js/logistics/redux'
 import { phoneNumberExample } from '../utils'
-import { OrderAlert } from './OrderAlert';
+import { TaskOrderInfo } from './TaskOrderInfo';
 
 const itemColor = event => {
   switch (event.name) {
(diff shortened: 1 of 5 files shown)


r0xsh:
haha can't resist doin it :upside_down_face: 
But thanks a lot, that's will truly enhance the UX, nice change to see :raised_hands:

js/app/dashboard/components/Navbar.js
```diff
@@ -355,6 +361,19 @@ const NavbarAntd = () => {
...
+      setTimeout(() => {
+        // highlight task in the list
+        dispatch(selectTask(initialTask))
+        dispatch(openTaskTaskList(initialTask))
+        // open task modal
+        dispatch(setCurrentTask(initialTask))
+      }, 500)
+    }
+  }, []) // no deps to run once
+
```
r0xsh: I guess it's too complex to have the task modal living it's own life and be displayed in the order path rather than into the dashboard with this hook ?

js/app/dashboard/redux/actions.js
```diff
@@ -508,14 +514,51 @@ export function openNewTaskModal() {
...
+    dispatch(_setCurrentTask(task))
+
+    window.history.replaceState(
+      {},
+      document.title,
+      window.Routing.generate('admin_dashboard_fullscreen', routeParams)
+    )
+  }
+}
+
```
r0xsh: Perfect 👌

r0xsh (commented):
Hope nothing wrong will happen with the removal of that event subscriber. But in any case that the direction we want to move to: cleaner data model 👍

js/app/dashboard/components/Navbar.js
vladimir-8: Yeah, it's definitely not what I was planning to do :) But on top of it, there are some benefits in displaying a task in the familiar dashboard interface where dispatchers can have more context on what has happened around this task, like what are the other tasks assigned to the same messenger, what are their route etc.

PR #5140, opened 2025-11-18
Order history page

vladimir-8:
Added: display order, tasks and incidents history into a single timeline

Order history

### DIFF
--- a/features/settings.feature
+++ b/features/settings.feature
@@ -35,6 +35,7 @@ Feature: Settings
         "motto": "@string@||@null@",
         "order_confirm_message": @string@,
         "edenred_client_id": "@string@",
-        "edenred_authorization_endpoint": "@string@"
+        "edenred_authorization_endpoint": "@string@",
+        "paygreen_public_key": "@null@"
       }
       """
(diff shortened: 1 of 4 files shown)


templates/order/item.html.twig
```diff
@@ -77,13 +77,13 @@
...
-    {% if delivery is not empty %}
-      <div id="delivery-info"
-           data-delivery="{{ delivery|serialize('jsonld', {groups: ['delivery', 'address', 'set-temporary-id-when-missing']}) }}"
-           data-is-dispatcher="{{ is_granted('ROLE_DISPATCHER') ? 'true' : 'false' }}">
-        {# rendered by React #}
-      </div>
-    {% endif %}
+    <div id="react-root"
+         data-order="{{ order|serialize('jsonld', {groups: ['order', 'pricing_deliveries']}) }}"
+         data-delivery="{{ delivery is not empty ? delivery|serialize('jsonld', {groups: ['delivery', 'address', 'set-temporary-id-when-missing']}) : '' }}"
```
r0xsh: I think otherwise react will crash
```suggestion
         data-delivery="{{ delivery is not empty ? delivery|serialize('jsonld', {groups: ['delivery', 'address', 'set-temporary-id-when-missing']}) : {} }}"
```

PR #5173, opened 2026-01-27
Fix/5148 require packages no package set

### LINKED ISSUE #5148: Require input of packages should not be checkable if there is no package set defined
Image

hackmanser:
Fixes #5148.

Context
On the store “Settings” tab, “Require input of packages” can currently be enabled even when no package set is selected/available, which leads to an inconsistent configuration.

What changed
- When no package set is selected, the “Require input of packages” checkbox is disabled and is automatically unchecked.
- On submit, we normalize the store configuration so `packagesRequired` cannot stay enabled when `packageSet` is null.

Notes
- The second commit (tooltip + `aria-disabled`) is optional UX/a11y and can be dropped easily if maintainers prefer keeping the PR minimal.

- I haven't been able to run tests locally—could a maintainer verify that submitting with packagesRequired: true + packageSet: null now correctly persists packagesRequired: false?

### DIFF
--- a/translations/messages.an.yml
+++ b/translations/messages.an.yml
@@ -406,7 +406,6 @@ form.store_type.package_set.label: 'Paquet conchunto'
 form.store_type.package_set.placeholder: 'Tría un paquet conchunto'
 form.store_type.weight_required.label: 'Pedir lo peso'
 form.store_type.packages_required.label: 'Pedir entrada de paquetz'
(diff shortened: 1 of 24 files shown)


alexsegura (approved):
Awesome! I just kind of reverted the translations (0a38f5ce063139a7a622e0784401f5b5b48ba8a5), because only EN translations are mandatory, if any translation is missing it will fallback to EN. 

Thank you for this @hackmanser

Issue #5180, opened 2026-01-29
Implement shops search by address using Tile38


alexsegura:
Currently search by address is implemented by looping through each enabled shop, and check if the searched address matches its `deliveryPerimeterExpression` (see [RestaurantFilter.php](https://github.com/coopcycle/coopcycle-web/blob/55eb933658211a64eb62e45ef3e13af9008a5f21/src/Utils/RestaurantFilter.php)). 

This is implemented like this, because restaurants can be either configured with a zone (a polygon) or a distance (a radius around the restaurant). 

Of course, this is highly inefficient. Even if all restaurants are configured with the same zone, it would still evaluate each restaurant individually. 

Instead, it _should_ be possible use Tile38 to store all restaurants as objects representing their zone or circle, and check if the point (the address searched) is inside.

PR #5181, opened 2026-01-29
Improve shop search by address performance

alexsegura:
Until #5180 is fixed let's try to improve what we have 🙂 

Currently, when all restaurants have the same zone, the implementation will retrieve the (same) zone from database multiple times (weird because Doctrine should have result cache) and evaluate it against the same address, when it could be done just once. 

This PR avoids doing things multiple times.


alexsegura:
Should also belong to this PR

03fd19cea1f6860ed5edd28f1a6276235d27db50...15f2d2f80dc07820bf61e2889b1f761a350ebccf

PR #5304, opened 2026-08-22
Upgrade Picqer barcode generator to v3

casperbakker:
## Summary

- upgrade `picqer/php-barcode-generator` from `^2.4` to `^3.2`
- update the lockfile from v2.4.2 to v3.3.0

CoopCycle requires PHP 8.2 or newer, which satisfies the v3 PHP requirement. The existing `BarcodeGeneratorSVG::getBarcode()` API remains supported.

## Validation

- `composer validate --strict` (passes with existing manifest warnings)
- exercised the SVG generation and barcode-width parsing path against v3.3.0 successfully
- `git diff --check`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -160,7 +160,7 @@
         "paygreen/paygreen-php": "*",
         "flow-php/parquet": "^0.16",
         "flow-php/filesystem": "^0.16",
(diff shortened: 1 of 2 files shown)


alexsegura:
Thank you @casperbakker

PR #5311, opened 2026-09-01
Delivery: External reference.

r0xsh:


alexsegura:
If I understand well, there is already the [`ref` property in the Task class](https://github.com/coopcycle/coopcycle-web/blob/0b10f82ff2a0a0a0e1313bcf497eee9653669ef9/src/Entity/Task.php#L439) offering the same functionality

r0xsh:
> If I understand well, there is already the [`ref` property in the Task class](https://github.com/coopcycle/coopcycle-web/blob/0b10f82ff2a0a0a0e1313bcf497eee9653669ef9/src/Entity/Task.php#L439) offering the same functionality

I had literally never seen that property before, haha. But no, perhaps we could say that `$ref` and `$barcode` serve the same purpose. However, the external reference should be stored on the delivery itself ; or perhaps on the pickup task, reusing the `$ref` field.

features/deliveries.feature
```diff
@@ -1147,6 +1147,78 @@ Feature: Deliveries
...
+      | payment_methods.yml |
+      | stores.yml          |
+    And the store with name "Acme" has an OAuth client named "Acme"
+    And the OAuth client with name "Acme" has an access token
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I add "Accept" header equal to "application/ld+json"
+    And the OAuth client "Acme" sends a "POST" request to "/api/deliveries" with body:
+      """
+      {
+        "externalReference": "FRMSY515302766",
```
alexsegura: In which existing use case is it used?

r0xsh: It's legacy code now.
But the idea of the `externalReference` is to keep track of a delivery from a foreign information system, so it make sense to expose it via API.

PR #5314, opened 2026-09-03
Fix stale Cypress CI description in README

marekl11:
Two details in the README's Cypress section no longer match
`.github/workflows/test_e2e.yml`: the workflow runs on every push rather
than only on the master branch, and the artifact step is called "Upload
images for failed tests" (plural). README only.


alexsegura:
Thank you @marekl11

PR #5335, opened 2026-09-20
Remove two unused IRI helpers

marekl11:
`iriToId` and `getIriPrefix` in `js/app/utils/iri.ts` have no callers anywhere in the repo, so this removes them.

A search for both names, as whole words and as plain substrings, finds only their own definitions. The one module that imports from `utils/iri` pulls in `restaurantToIri` and `userToIri`, which stay.

`getIriPrefix` is also subtly wrong if anyone did reach for it: it guards on `parts.length < 2` but then reads `parts[2]`, so a two-segment IRI returns `/foo/undefined`. That seemed like a reason to delete it rather than keep it around.

Deletion only, nothing else touched.

For transparency: found by a local model looking for unused exports, then checked by hand before opening.


alexsegura:
Thank you @marekl11

